PR #79, opened 2026-03-05
feat(embed): add MUNINN_OPENAI_URL env var for custom OpenAI base URL

### LINKED ISSUE #78: Add configurable OpenAI URL for embeddings
I'd like to point the OpenAI endpoint at somewhere custom for embeddings and also to set an embedding model.

ojamin:
### Summary

Adds `MUNINN_OPENAI_URL` to override the OpenAI embeddings base URL, so users can point embedding requests at OpenAI-compatible endpoints like LocalAI or an internal gateway.

### What changed

- New env var `MUNINN_OPENAI_URL` accepted in both `resolveEmbedInfo()` and `buildEmbedder()`, with support for plain HTTP URLs (`http://localhost:8080/v1`) and provider-style URLs (`openai://text-embedding-3-small?base_url=http://localhost:8080`)
- Saved config (`plugin_config.json`) also respects the URL override
- If the override is invalid (e.g. `ftp://...`), OpenAI is skipped rather than silently falling back to `api.openai.com` — avoids accidentally leaking traffic to the public API when the user explicitly configured a private endpoint
- Port range validation added to `parseHTTPBaseURL()` (rejects port 0 / 65536)
- Provider URLs are sanitized before logging to avoid leaking internal hostnames
- `muninn help` and `muninn init` mention the new env var
- Docs updated in README, architecture, plugins, quickstart, and self-hosting

### Tests

- Valid URL override resolves correct provider/model
- Invalid override skips OpenAI, falls through to next provider
- Invalid env override takes precedence over valid saved config
- Port boundary validation

### Notes

This does not add any additional functionality to the web ui.

Aligns with the enrich option such that a plugin_config.json config file can be configured as follows:
{
    "embed_provider": "openai",
    "embed_url": "openai://qwen3-embedding-4b?base_url=http://127.0.0.1:8080/v1",
    "embed_api_key": "api-key",
    "enrich_provider": "openai",
    "enrich_url": "openai://GLM-4.7-Flash-GGUF?base_url=http://127.0.0.1:8080/v1",
    "enrich_api_key": "api-key"
}

Closes #78

### DIFF
--- a/README.md
+++ b/README.md
@@ -353,7 +353,7 @@ When you're ready to customize:
 |------|-----|
 | Embedder: bundled (default) | On automatically — set `MUNINN_LOCAL_EMBED=0` to disable |
 | Embedder: Ollama | `MUNINN_OLLAMA_URL=ollama://localhost:11434/nomic-embed-text` |
-| Embedder: OpenAI | `MUNINN_OPENAI_KEY=sk-...` |
+| Embedder: OpenAI | `MUNINN_OPENAI_KEY=sk-...` (+ optional `MUNINN_OPENAI_URL=http://localhost:8080/v1`; invalid override disables OpenAI init) |
 | Embedder: Voyage | `MUNINN_VOYAGE_KEY=pa-...` |
 | Embedder: Cohere | `MUNINN_COHERE_KEY=...` |
 | Embedder: Google (Gemini) | `MUNINN_GOOGLE_KEY=...` |
--- a/cmd/muninn/help.go
+++ b/cmd/muninn/help.go
@@ -328,6 +328,7 @@ func printHelp() {
 	fmt.Println()
 	fmt.Printf("  %-28s %s\n", "MUNINN_OLLAMA_URL", "Local Ollama embed model (e.g. ollama://localhost:11434/nomic-embed-text)")
 	fmt.Printf("  %-28s %s\n", "MUNINN_OPENAI_KEY", "OpenAI embeddings API key (text-embedding-3-small, 1536d)")
+	fmt.Printf("  %-28s %s\n", "MUNINN_OPENAI_URL", "Optional OpenAI base URL or provider URL override")
 	fmt.Printf("  %-28s %s\n", "MUNINN_VOYAGE_KEY", "Voyage AI embeddings API key (voyage-3, 1024d)")
 	fmt.Printf("  %-28s %s\n", "MUNINN_COHERE_KEY", "Cohere embeddings API key (embed-v4, 1024d)")
 	fmt.Printf("  %-28s %s\n", "MUNINN_GOOGLE_KEY", "Google Gemini embeddings API key (text-embedding-004, 768d)")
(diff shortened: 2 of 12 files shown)


scrypster (changes requested):
Thanks for this — the use case is real and the overall approach is sound. `MUNINN_OPENAI_URL` directly addresses #78 and #77, and the fail-closed security posture (invalid override disables OpenAI rather than silently falling back to `api.openai.com`) is exactly the right call.

That said, there are a few concrete issues to fix before merge. None require a redesign — they're all in-place fixes.

## Required changes

### 1. Log leak in `openAIEmbedLogAttrs`

`openAIEmbedLogAttrs` logs `provCfg.BaseURL` verbatim, which defeats the `sanitizeProviderURLForLog` function added elsewhere. Internal hostnames shouldn't appear in logs.

Fix: log a boolean or redacted form instead, e.g.:
```go
return []any{"model", provCfg.Model, "custom_base_url", provCfg.BaseURL != "https://api.openai.com"}
```

### 2. Duplicated URL resolution logic

Both `resolveEmbedInfo` and `buildEmbedder` independently read `MUNINN_OPENAI_URL`, call `resolveOpenAIEmbedProviderURL`, and implement the same validity/precedence logic. This means the two paths can drift — a bug fix in one may not land in the other (this already happened: the `openaiSource` fallback-to-`cfg.EmbedURL` logic is slightly different between the two functions).

Extract the resolution into a single call site so both functions share one code path.

### 3. Case-insensitive scheme detection

```go
if strings.HasPrefix(raw, "openai://") {
```

URI schemes are case-insensitive per RFC 3986. `OPENAI://text-embedding-3-small?base_url=...` would be misclassified as a bare HTTP URL and fail validation. Use `strings.HasPrefix(strings.ToLower(raw), "openai://")` or parse with `net/url` first.

### 4. Silent query/fragment drop in `parseHTTPBaseURL`

If someone passes `http://gateway.com/api?key=foo`, the query string is silently dropped. That's the worst outcome — the user thinks their config is applied but it isn't. Either reject it with a clear error ("base_url must not contain query parameters") or preserve it. Don't drop silently.

## Nice-to-have (low effort, include while revising)

### 5. Add a doc comment to `parseHTTPBaseURL`

The `/v1` stripping is correct (the OpenAI embed client appends `/v1/embeddings` itself) but non-obvious. A one-line comment prevents future confusion:
```go
// Strip trailing /v1 — the OpenAI embed client appends /v1/embeddings itself.
cleanPath = strings.TrimSuffix(cleanPath, "/v1")
```

### 6. Add missing test cases

- Bare HTTP URL with a path prefix: `MUNINN_OPENAI_URL=https://gateway.example.com/openai/v1`
- `MUNINN_OPENAI_URL` set without `MUNINN_OPENAI_KEY` (should skip OpenAI entirely — correct behavior, worth documenting as a test)

---

The env var naming, the `EmbedURL` field reuse (backward-compatible since it's disambiguated by `embed_provider`), and the test coverage of the main paths are all good. Fix the above and this is ready to merge.

scrypster (approved):
Tests look solid, use case is real and the implementation is sound. Merging — will follow up with a cleanup PR addressing the minor issues noted in the review.

PR #146, opened 2026-03-08
Harden body-based vault auth and always serialize fact memory_type

### LINKED ISSUE #145: Body-based vault enforcement is bypassable on some REST writes and fact memory_type is omitted from REST reads
## Summary

  REST vault enforcement is inconsistent on some body-based routes.

  Two request shapes can bypass or weaken vault validation:

  - JSON bodies sent with a non-JSON `Content-Type` can skip body vault extraction even though the
  handler still decodes the JSON body
  - batch writes with a top-level `vault` can fail to validate conflicting `engrams[].vault` values

  There is also a REST serialization inconsistency where fact memories (`memory_type = 0`) can be omitted
  from responses because the field is tagged with `omitempty`.

  ## Observed behavior

  - `POST /api/engrams` can accept a JSON body with `Content-Type: text/plain`, and auth may scope the
  request to the API key vault while the handler later uses a different body `vault`
  - `POST /api/engrams/batch` can accept a top-level `vault` while preserving conflicting non-empty item
  vaults
  - fact-classified memories can come back without `memory_type` even though the API schema documents it
  as an integer field

  ## Expected behavior

  - vault extraction and vault/key enforcement should apply consistently whenever the body is valid JSON,
  regardless of `Content-Type`
  - batch requests should reject any conflict between top-level `vault` and `engrams[].vault`
  - fact memories should serialize `memory_type: 0` instead of omitting the field

  ## Impact

  These auth gaps can allow cross-vault writes under specific request shapes, and the serialization issue
  makes REST responses inconsistent for fact memories.

ojamin:
## Summary

  This follow-up hardens REST vault enforcement on body-based routes and fixes a REST serialization
  inconsistency for fact memories.

  The main problem was that vault resolution and vault enforcement were not applied consistently across
  all valid JSON request shapes. In particular, auth could skip body vault extraction when a client sent
  JSON with a non-JSON `Content-Type`, and batch validation could stop at the top-level `vault` without
  checking conflicting item-level vaults.

  This PR closes those gaps, keeps the earlier body-vault routing changes intact, and preserves the REST
  fix for `memory_type: 0`.

  ## Changes

  - inspect request bodies for vault routing when they are valid JSON objects, even if `Content-Type` is
  non-JSON
  - keep rejecting malformed JSON bodies on JSON-style write routes
  - validate top-level batch `vault` against every non-empty `engrams[].vault`
  - continue preferring query `vault` when present, but require body/query consistency
  - keep `handleSubscribe` on the authenticated context vault instead of re-reading raw query input
  - update changelog/docs for the body-vault behavior

  ## Why

  - `text/plain` JSON bodies were still decoded by handlers, so skipping vault extraction in auth created
  a real cross-vault write risk
  - batch requests need a single authoritative vault; partial validation leaves room for mixed-vault
  payloads
  - `memory_type = 0` is a real value, not “missing data”, so `omitempty` was the wrong contract

  ## Tests

  Added or updated tests for:

  - body-vault auth with matching API key
  - public body-vault routing
  - query/body vault mismatch
  - batch body-vault inference
  - mixed batch vault rejection
  - `text/plain` JSON cross-vault rejection
  - top-level batch vault vs item vault rejection
  - REST routing coverage for the same write and batch cases
  - REST read serialization of `memory_type: 0`

  ## Validation

  Validated with focused Go test suites and full repo verification:

  - `go test ./internal/auth`
  - `go test ./internal/transport/rest`
  - `go build ./...`
  - `go vet ./...`
  - `go test ./...`

  Also validated live against an AIC-managed MuninnDB instance:

  - body-vault create succeeded with `201`
  - `text/plain` cross-vault write returned `401 VAULT_KEY_MISMATCH`
  - batch top-level/item mismatch returned `400 INVALID_VAULT_REQUEST`

  ## Release Checklist

  - [x] `CHANGELOG.md` updated with changes
  - [x] OpenAPI spec updated if API routes changed (not applicable: no route/schema changes)
  - [x] SDK types updated if request/response schemas changed (not applicable: no schema shape changes)
  - [x] `docs/` updated if user-facing behavior changed
  - [x] `go build ./...` clean
  - [x] `go vet ./...` clean
  - [x] `go test ./...` passes

Closes #145

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Retry and retroactive enrichment now only mark entity and relationship stages complete after successful persistence, avoiding partial-state retries, nil-result crashes, and silent graph-write failures.
 - Entity and relationship response parsing now rejects nested wrapper keys like `meta.entities` / `meta.relationships` instead of treating them as valid empty results.
 - Vault-scoped REST routes now resolve non-default vaults consistently from authenticated request bodies as well as `?vault=`, and reject mismatched query/body vaults.
+- Vault-scoped REST routes are setup to deprecate vault passed in the body in a later release.
 - REST read responses now include `memory_type: 0` for fact-classified memories instead of omitting the field.
 
 ---
--- a/docs/auth.md
+++ b/docs/auth.md
@@ -88,7 +88,7 @@ curl http://127.0.0.1:8475/api/engrams?vault=default \
 ```
 
 The key implicitly identifies the vault. If the key belongs to `default` and the request specifies a different vault, the request is rejected.
-For body-based REST routes, the vault may be supplied either as `?vault=` or in the JSON body. If both are present, they must match.
+For body-based REST routes, the vault must be supplied as `?vault=`. Some routes allow vault to be in the JSON body for compatibility but this will be deprecated soon. If both are present, they must match.
 
 ---
 
--- a/internal/auth/middleware.go
+++ b/internal/auth/middleware.go
@@ -12,12 +12,18 @@ import (
 
 // VaultAuthMiddleware enforces vault-level API key auth.
 // Vault is resolved from ?vault= query param first, then from JSON request bodies
-// on body-based routes, and finally defaults to the key's vault (or "default"
-// for unauthenticated/public requests) when no explicit vault is provided.
+// on body-based routes, and finally defaults to "default" when no explicit
+// vault is provided.
 // Public vaults allow unauthenticated access in observe mode.
 // If a Bearer token is present, it is always validated regardless of vault visibility.
 func (s *Store) VaultAuthMiddleware(next http.HandlerFunc) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
+		vault, resolveErr := resolveRequestVault(r, "default")
+		if resolveErr != nil {
+			writeVaultRequestError(w, http.StatusBadRequest, resolveErr)
+			return
+		}
+
 		authHeader := r.Header.Get("Authorization")
 
 		if authHeader != "" {
@@ -25,11 +31,6 @@ func (s *Store) VaultAuthMiddleware(next http.HandlerFunc) http.HandlerFunc {
 			key, err := s.ValidateAPIKey(token)
 			if err != nil {
 				http.Error(w, `{"error":"invalid api key"}`, http.StatusUnauthorized)
-				return
-			}
-			vault, resolveErr := resolveRequestVault(r, key.Vault)
-			if resolveErr != nil {
-				writeVaultRequestError(w, http.StatusBadRequest, resolveErr)
 				return
 			}
 			// Enforce vault scoping: the key must be issued for the requested vault.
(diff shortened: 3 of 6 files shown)


ojamin:
I might need to go relook at this one in a bit

scrypster:
No rush — take your time. Just let us know when you're happy with it and we'll merge.

scrypster:
Hey @ojamin — we ran both approaches past Claude Opus (as an architecture review) and wanted to share the findings in case it's useful as you revisit.

**Opus recommendation: Option A — query param as the sole vault source (our approach in #148).**

The core reasoning: vault is an *authorization* input, not a *data* input. It belongs in the request envelope (URL/query params), not the payload body — same reason OAuth tokens go in the `Authorization` header rather than the JSON body. Treating body vault as a legitimate auth mechanism conflates the control plane with the data plane, which is exactly what created #145 in the first place.

Specific concerns Opus raised about the middleware-body-parsing approach:
- **Body read/restore is fragile** — `io.NopCloser` replay breaks with large bodies, chunked transfer encoding, or other middleware that also reads the body
- **Content-type sniffing** (inspecting `text/plain` bodies for JSON vault fields) is a vulnerability class, not a fix
- **Batch vault consistency** is a business rule that belongs in the handler, not auth middleware
- **Two auth input sources** makes the system harder to audit (query params appear in access logs by default; body fields don't)

**Our #148** took the simpler path: handlers unconditionally overwrite body vault with `ctxVault(r)`. The body `vault` field stays in the struct for schema compat but is silently ignored. One source of truth, no middleware changes, no parsing.

That said — totally a collaboration here. You caught things we missed (the `handleSubscribe` bug especially). If you see it differently or think the body vault support is worth the tradeoff, we're all ears. Just sharing the Opus analysis as input.

ojamin:
Your call tbh, I might be concerned about breaking changes for no reason with this and maybe we should just force query only right now.

If we do that, imo we should hard error on vault being passed in the body to ensure no one is getting unexpected behaviour though.

ojamin:
@scrypster I'm good with this now, so if you want to add to it etc. or merge feel free.

Testing passes here

  - go test ./internal/auth ./internal/transport/rest
  - go test ./...
  - Live REST checks all passed:
      - query-only authenticated write: 201
      - matching query/body vault write: 201
      - mismatched body vault: 400 INVALID_VAULT_REQUEST
      - omitted vault with non-default key: 401 VAULT_KEY_MISMATCH
      - batch write with implicit/matching vaults: 201
      - batch item vault mismatch: 400 INVALID_VAULT_REQUEST
      - GET /api/session?vault=...: 200
      - read-back of a created engram: 200, including memory_type: 0

ojamin:
@scrypster as I said before, the only reason I've not done it like that in this patch is that it doesnt seem like a good idea to push breaking changes in minor version. If anyone is making requests with vault in the body this would be a hard breaking change.

If you want to do minor version breaking changes then I can totally change this to work that way but swapping a structured bit of this patch out at a good breaking change point will be super easy and wouldnt really add any perm complexity as long as we track actually doing it at an appropriate point.

As part of this patch we already have the reusable helpers in place that give us single points to swap out to the error on vault in body so its a super easy switch out.

If you wanted, we could merge this and I could throw up another PR that transitions us to the final state that can just get merged when you hit the right point.

Either way, at this point its more of a "whats the project policy for breaking changes" call than a coding one, just let me know!

PR #148, opened 2026-03-08
fix(rest): close auth bypass on body vault override and fix memory_type omission (#145)

### LINKED ISSUE #145: Body-based vault enforcement is bypassable on some REST writes and fact memory_type is omitted from REST reads

scrypster:
## Summary

- **Auth bypass (write/read)**: `POST /api/engrams`, `POST /api/engrams/batch`, `POST /api/activate`, `POST /api/link`, and `POST /api/engrams/links/batch` all had `if req.Vault == ""` guards that let a client override the auth-middleware-resolved vault by supplying a different vault in the JSON body. A key scoped to vault A could operate on vault B. Fixed by unconditionally assigning `req.Vault = ctxVault(r)`.
- **`memory_type` serialization**: `MemoryType uint8` was tagged `json:"memory_type,omitempty"`, silently dropping zero-value (fact type = 0) from JSON responses. Fixed by removing `omitempty` from the json tag (msgpack tag unchanged).

## Test Plan
- [ ] `TestVaultRouting_Write_BodyVaultIgnored` — body vault cannot override auth vault on write
- [ ] `TestVaultRouting_BatchCreate_BodyVaultIgnored` — per-item body vault ignored in batch create
- [ ] `TestVaultRouting_Activate_BodyVaultIgnored` — body vault ignored on activate
- [ ] `TestVaultRouting_Link_BodyVaultIgnored` — body vault ignored on link
- [ ] `TestVaultRouting_BatchGetEngramLinks_BodyVaultIgnored` — body vault ignored on batch link fetch
- [ ] `TestReadResponse_MemoryTypeZeroIncludedInJSON` — `memory_type: 0` present in JSON response
- [ ] Full REST and mbp test suites pass
- [ ] CI green

Closes #145

### DIFF
--- a/internal/transport/mbp/types.go
+++ b/internal/transport/mbp/types.go
@@ -86,7 +86,7 @@ type WriteRequest struct {
 	Embedding    []float32     `msgpack:"embedding,omitempty" json:"embedding,omitempty"`
 	Vault        string        `msgpack:"vault,omitempty" json:"vault,omitempty"`
 	IdempotentID string        `msgpack:"idempotent_id,omitempty" json:"idempotent_id,omitempty"`
-	MemoryType   uint8         `msgpack:"memory_type,omitempty" json:"memory_type,omitempty"`
+	MemoryType   uint8         `msgpack:"memory_type,omitempty" json:"memory_type"`
 	TypeLabel    string        `msgpack:"type_label,omitempty" json:"type_label,omitempty"`
 
 	// Inline enrichment: caller-provided data that bypasses background enrichment.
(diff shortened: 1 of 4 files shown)


scrypster:
Closing in favor of #146 from @ojamin which covers the same bugs plus additional hardening (handleSubscribe fix, text/plain JSON body validation, mixed batch vault rejection). Better and more comprehensive solution.

PR #151, opened 2026-03-09
fix(ui): guard missing id in live feed, restore vault modal load, improve SSE error logging

To3Knee:
## Summary

Three Alpine.js bugs traced to a single root cause: a malformed SSE `memory_added` event with a missing `id` field corrupts Alpine's `x-for` key tracking, cascading to break the Entity Graph and Vault Selector.

- **Fix 1** — Guard in `_handleLiveMessage()` to skip `memory_added` events missing `id`. Prevents Alpine duplicate-key corruption and the cascade failure.
- **Fix 2** — Add `loadVaults()` to the `vaultModalOpen` watcher. Vault list was only fetched at login; new vaults created after page load never appeared without a full reload.
- **Fix 3** — Replace silent `catch (_) {}` in the SSE `onmessage` handler with `console.warn`. Makes SSE processing errors visible in DevTools instead of failing silently.

## Root Cause Detail

`undefined === undefined` is `true` in JavaScript. Once one `memory_added` event with `id: undefined` enters `liveFeed`, the existing dedup check (`liveFeed.some(item => item.id === msg.data.id)`) blocks future duplicates — but the poisoned item stays in the array. Alpine's `x-for` uses `item.id` as the DOM anchor key, so every re-render hits a duplicate `undefined` key, corrupting internal state and breaking all reactive components on the page.

## Files Changed

- `web/static/js/app.js` — all 3 fixes
- `FIXES.md` — full writeup: problem, root cause, diff, verification steps, and deployment options (Docker rebuild, --dev hot patch, build from source)
- `muninndb.service` — systemd unit template for self-hosters

## Notes

- Also documents that the `ghcr.io/scrypster/muninndb:latest` Docker image (as of March 2026) is missing `POST /api/engrams/links/batch` and `GET /api/admin/entity-graph` — both required by the UI. Building from source resolves this.
- Fixes are independent and can be applied separately.
- Verified on a live Rocky Linux 8 instance with real vault data.


scrypster:
Hey @To3Knee — this is a great catch. The Alpine `undefined` key corruption is subtle and nasty — the dedup check becoming a permanent block once a poisoned item enters `liveFeed` is exactly the kind of bug that's hard to trace. Really appreciate the detailed root cause writeup too.

All three JS fixes are solid and we'd like to merge them. Two things we'd ask before merging:

1. **FIXES.md** — the content is great context but it belongs in the issue/PR description and git history rather than the repo root (it'll go stale and references your fork). Could you drop that file from the PR?
2. **muninndb.service** — a systemd unit is genuinely useful for self-hosters and we'd love to have it, but it has placeholder secrets and should live in a `deploy/` or `contrib/` directory. Would you be open to pulling it into a separate focused PR so it gets the attention it deserves?

Once those two files are out, we're good to merge. Thanks again — verified on a live Rocky Linux instance with real data is exactly what we want to see.

PR #152, opened 2026-03-09
fix(enrich): deadline-aware early exit for replay enrichment

maybebyte:
## Summary

- Add deadline-aware early exit to the replay enrichment loop — when the context expires, stop gracefully instead of letting remaining engrams fail with `context deadline exceeded`
- Surface `failed` and `remaining` counts in the MCP response so callers get accurate feedback instead of the misleading `skipped: 0`
- The invariant `processed + skipped + failed + remaining == total` always holds

## Problem

`muninn_replay_enrichment` shares a single 30-second MCP request context across all engrams. When enriching 50 items, the first ~8 complete before the deadline fires, and the remaining ~42 all fail simultaneously with `context deadline exceeded`. The response reported `processed: 8, skipped: 0` — silently dropping 42 failures.

## Changes

- **`internal/engine/engine_replay.go`**: Check `ctx.Err()` before each engram in the loop; on cancellation, return a result with accurate `Remaining` count. Track `Failed` count for enrichment errors.
- **`internal/mcp/handlers.go`**: Include `failed` and `remaining` fields in the JSON response.
- **`internal/engine/engine_replay_test.go`**: Tests for failed counting, mid-loop cancellation, and pre-expired context.
- **`internal/mcp/handlers_test.go`**: Tests verifying the new fields appear in the MCP response.

## Test plan

- [x] `TestReplayEnrichment_FailedCount` — mock plugin fails on 2nd call, verify `Failed: 1`
- [x] `TestReplayEnrichment_ContextCancellation` — cancel after 2 enrichments, verify `Remaining > 0`
- [x] `TestReplayEnrichment_ContextAlreadyExpired` — pre-cancelled context, verify all items reported as `Remaining`
- [x] `TestHandleReplayEnrichment_FailedAndRemainingInResponse` — verify `failed`/`remaining` in MCP JSON
- [x] Manual: run `muninn_replay_enrichment` with limit=50 and confirm response now shows remaining count instead of silent failures

### DIFF
--- a/internal/mcp/tools.go
+++ b/internal/mcp/tools.go
@@ -581,7 +581,7 @@ func allToolDefinitions() []ToolDefinition {
 		// Enrichment replay
 		{
 			Name:        "muninn_replay_enrichment",
-			Description: "Re-run the enrichment pipeline for memories in a vault that are missing specific digest stages (entities, relationships, classification, summary). Use this to retroactively enrich memories that were stored before an LLM provider was configured, or to fill in specific pipeline stages that were skipped. Supports dry_run=true to preview what would be processed without writing.",
+			Description: "Re-run the enrichment pipeline for memories in a vault that are missing specific digest stages (entities, relationships, classification, summary). Use this to retroactively enrich memories that were stored before an LLM provider was configured, or to fill in specific pipeline stages that were skipped. Supports dry_run=true to preview what would be processed without writing. The response includes processed (successfully enriched), skipped (already enriched or nothing to enrich), failed (enrichment or persistence errors), and remaining (not reached before context deadline/cancellation) counts.",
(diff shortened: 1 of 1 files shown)


scrypster:
Hey @maybebyte — really appreciate this, the silent failure on context deadline expiry is a real problem and the fix is well thought out. The accounting invariant (`processed + skipped + failed + remaining == total`) is exactly the right approach, and the tests are solid.

One issue blocking the merge: the PR references `UpdateDigest` with a `TypeLabel` parameter that isn't on develop yet, so it won't compile against the current branch. Could you rebase onto develop (or let us know which dependency this is built on top of) and we'll get it merged once it lines up?

Great work on both this and #149 — really valuable contributions.

Issue #124, opened 2026-03-07
Node Graph on Memory Graph page improvements


ojamin:
# On the Memory Graph page

## Visualisation size
Current: the visualisation is fixed max height and width
Expected: the visualisation expands to fill the available space

## Label Info Select
Current: all node labels/ids are shown in full always
Expected: a selector to switch between show in full, show short version, show none

## Info Popout
Current: when selecting a node a brief toast notification pops up at the bottom of the screen
Expected: any of
- a "more" link in the toast to show a popup with full node info
- a show node info toggle that shows node info when a node is clicked through either opening a side bar or opening a popup

## Node Arrangement
Current: after clicking Load Graph, the nodes load in all clustered together and frequently on top of each other
Expected: the nodes load in with reasonable spacing such that association groupings etc. can be easily seen in the visualisation

## Node Arrange Buttons
Current: other than zoom/fit there are no interaction points for the visualisation
Expected: a variety of different options to have the visualisation auto arrange the nodes based on various indicators etc.

## Node Filter
Current: the Load Graph button pulls the top 50 nodes
Expected: a filter that lets the user pick from either the top x, nodes with the most/least links, most recent, oldest, added/updated in x datetime range, nodes with x steps removed from node id or the top node for a search

scrypster:
Thanks @ojamin — this is a solid list of UX improvements for the graph page, and you've articulated each one clearly. We've looked at these and the top priorities land on: better viewport/sizing, a proper info sidebar instead of the toast, and smarter initial node spacing. The filter controls and additional layout options are on the roadmap too. We'll keep this ticket as the tracking point.

PR #155, opened 2026-03-09
feat(graph): memory graph UX — spacing, resize, label toggle

To3Knee:
Addresses #124 — memory graph UX improvements.

## What changed

**Node spacing:** fCoSE layout tuned — `nodeRepulsion: 6500`, `idealEdgeLength: 120`, `tile: true` with 30px padding. Nodes load spread out and grouped by association instead of clustering at center.

**Graph fills available space:** `ResizeObserver` on `#cy` calls `cy.resize()` on container size changes (debounced 150ms). Fit is handled by `layoutstop` event to avoid zooming on pre-layout positions.

**Label toggle:** Button in toolbar cycles **Full → Short (20 chars with ellipsis) → Off**. State persists while graph is loaded. Uses per-element `node.style('label', …)` bypass for instant response without re-rendering.

## Notes

- Rebased onto current `develop`
- Verified on Rocky Linux 8.10, muninndb vault (451 entities / 624 relationships)

### DIFF
--- a/web/static/js/app.js
+++ b/web/static/js/app.js
@@ -63,6 +63,7 @@ document.addEventListener('alpine:init', () => {
     // Graph
     graphLoaded: false,
     graphTab: 'memory',
+    graphLabelMode: 'full', // 'full' | 'short' | 'none'
     _cy: null,
     entityGraphLoaded: false,
     entityGraphStatus: '',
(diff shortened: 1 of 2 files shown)


scrypster:
Both passes look clean — ready to merge. One last thing before we do: the PR title and body still reference the entity graph fix (Fix 1) that landed separately in #151. Could you update the title to something like `feat(graph): memory graph UX — spacing, resize, label toggle` and trim the body to just the memory graph work? Keeps the git log clean.

PR #160, opened 2026-03-09
fix(ui): live feed detection, vault modal refresh, Alpine crash guards

vcheckk:
## Summary

- **Remove duplicate `const newVaultCount`** that crashed strict-mode runtimes (redeclared in same scope)
- **Add engram count-diff tracking** (`_prevEngramCount`) so `_fetchNewestEngram` fires when new memories are stored via MCP or API
- **Restore `_fetchNewestEngram` method** lost during prior refactors — was called but never defined
- **Add `loadVaults()` call** when vault modal opens so newly created vaults appear without a page reload
- **Guard `memory_added` events** with missing `id` to prevent Alpine x-for `"can't access property 'after', v is undefined"` crash cascade

## Test plan

- [x] Store a memory via MCP — confirm it appears in live feed without page refresh
- [x] Create a new vault — confirm it appears in vault picker modal without page refresh
- [x] Unit tests pass (`make test` — 4 pre-existing `TestLiveIntegration_*` failures are auth-config issues on `upstream/main`, unrelated to this change)
- [ ] Verify no console errors on dashboard load
- [ ] Verify live feed handles rapid successive writes without duplicates

### DIFF
--- a/web/static/js/app.js
+++ b/web/static/js/app.js
@@ -306,9 +306,15 @@ document.addEventListener('alpine:init', () => {
       // Load initial data (gated on auth check)
       await this.checkAuth();
 
-      // Fetch vault engram counts whenever the vault picker modal opens.
+      // Fetch vault list and engram counts whenever the vault picker modal opens.
+      // loadVaults() was previously missing here — the vault list would only
+      // refresh at login/auth-check, so newly created vaults wouldn't appear
+      // until page reload.
       this.$watch('vaultModalOpen', (open) => {
-        if (open) this.loadVaultStats();
+        if (open) {
+          this.loadVaults();
+          this.loadVaultStats();
+        }
       });
     },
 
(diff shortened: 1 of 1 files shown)


scrypster:
Hey @vcheckk — thanks for the contribution! After reviewing this against `develop`, here's what we found:

- The `memory_added` guard, `loadVaults()` in the vault modal watcher, and the comment improvements all landed in **PR #151** which merged to `develop` earlier today — so those pieces are already in.
- The new `_fetchNewestEngram()` + `_prevEngramCount` tracking is a different approach, but has a cross-vault race condition: `stats_update` broadcasts a global engram count, so a write to vault B while viewing vault A would trigger a fetch against the wrong vault. The existing `memory_added` SSE path already handles live feed updates — if that's unreliable for you, let us know the scenario and we can look at it properly.

We're going to close this one as superseded. **We're planning to merge `develop` into `main` shortly**, so all these fixes will be available soon — no action needed on your end.

Branch from `develop` for future PRs and you'll always be working against the latest. Thanks again!

PR #167, opened 2026-03-10
fix(ui): add ?vault= query param to all POST api calls

To3Knee:
## Summary

`VaultAuthWithAdminBypass` reads the vault exclusively from the URL query string when an admin session cookie is present — it does not inspect the request body. Seven POST calls in the web UI passed `vault` only in the JSON body, causing `ctxVault(r)` to resolve to `"default"` instead of the selected vault. `resolveHandlerVault` then rejected the body vault as a mismatch, returning a 400 with error code 4003 (`vault must match the authenticated request vault`).

**Affected callers:**

| Endpoint | Function | Symptom |
|---|---|---|
| `/api/engrams/links/batch` | `loadGraph()` | "Graph failed: 400" |
| `/api/activate` | `searchMemories()` | silent failure on non-default vaults |
| `/api/engrams` | write memory modal | write fails silently |
| `/api/link` | link modal | link fails silently |
| `/api/explain` | explain score | explain fails silently |
| `/api/consolidate` | multi-select consolidate | consolidate fails silently |
| `/api/decide` | decide modal | decide fails silently |

The `fix(ui): replace silent catch blocks` commit (already in main) surfaced these as visible errors; the `loadGraph` failure became the most noticeable symptom.

## Fix

Added `?vault=encodeURIComponent(this.vault)` to the URL for all seven POST calls so the middleware can correctly scope the request before the handler runs. The vault field in the body is unchanged (it is still validated by `resolveHandlerVault` against the context vault for defense-in-depth).

## Test plan

- [ ] Load the memory graph on a non-default vault — should render without "Graph failed" error
- [ ] Search memories on a non-default vault — results should return
- [ ] Write a new memory on a non-default vault — should succeed
- [ ] Create a link on a non-default vault — should succeed
- [ ] Explain score on a non-default vault — should return
- [ ] Consolidate memories on a non-default vault — should succeed
- [ ] Submit a decision on a non-default vault — should succeed

### DIFF
--- a/web/static/js/app.js
+++ b/web/static/js/app.js
@@ -759,7 +759,7 @@ document.addEventListener('alpine:init', () => {
         if (this.searchMode && this.searchMode !== 'balanced') {
             body.mode = this.searchMode;
         }
-        const data = await this.apiCall('/api/activate', {
+        const data = await this.apiCall('/api/activate?vault=' + encodeURIComponent(this.vault), {
           method: 'POST',
           body: JSON.stringify(body),
         });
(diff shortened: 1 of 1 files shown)


To3Knee:
Sorry about that — here is the update.

Added `?vault=` to the four missed call sites:

- `PUT /api/engrams/{id}/tags` — edit tags modal
- `POST /api/link` (×2) in `resolveContradiction`
- `PUT /api/engrams/{id}/state` (×2) in `resolveContradiction`

Also converted the raw `fetch()` calls in `resolveContradiction` and `updateLifecycleState` to `this.apiCall()` for consistent `credentials: 'same-origin'` handling. Total: 11 call sites fixed across the file.

PR #181, opened 2026-03-10
test(cognitive): add bearer auth to live integration tests

scrypster:
## Summary

- Adds `bearerTransport` RoundTripper that injects `Authorization: Bearer <token>` on every request
- Adds `liveClient(t)` helper that reads `MUNINN_TEST_TOKEN` (skip if unset) and `MUNINN_TEST_URL` (default `http://127.0.0.1:8475`), verifies connectivity, then returns an authenticated client
- Replaces the 4 unauthenticated skip guards in `mechanics_proof_test.go` with `client, base := liveClient(t)`
- Replaces all `http.Post`/`http.Get` calls in the 4 live integration tests with `client.Post`/`client.Get`

## Behavior change

| Scenario | Before | After |
|---|---|---|
| No daemon, no token | Skip | Skip |
| Daemon running, no token | **FAIL** (401 on write endpoints) | **Skip** |
| Daemon running, token set | **FAIL** (401 on write endpoints) | **PASS** |
| CI (no daemon, no token) | Skip | Skip |

## Test plan

- [x] `go build ./internal/cognitive/...` — compiles cleanly
- [x] `go test ./internal/cognitive/... -count=1` — all unit tests pass (live tests skip without token)
- [ ] `MUNINN_TEST_TOKEN=<key> go test ./internal/cognitive/... -run TestLiveIntegration -v` — validates against running server

### DIFF
--- a/internal/cognitive/mechanics_proof_test.go
+++ b/internal/cognitive/mechanics_proof_test.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"math"
 	"net/http"
+	"os"
 	"strings"
 	"sync"
 	"testing"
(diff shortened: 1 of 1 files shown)


scrypster:
Changes already landed in develop via #182 squash merge.

PR #188, opened 2026-03-10
fix(retroactive): idle backoff + log suppression for phantom counts

vcheckk:
## Summary

- **Idle backoff**: When `CountWithoutFlag` reports pending items but the scan produces no actual work (phantom count mismatch), the poll interval exponentially increases from 3s → 6s → 12s → ... → 3min cap. Resets to 3s immediately on `Notify()` (new engram written) or when real work is done.
- **Log suppression**: "starting"/"complete" demoted to `DEBUG` for idle passes. `INFO`-level "complete" now only fires when items were actually processed, with new `pass_processed` / `total_processed` fields for clearer observability.
- **New helper**: `processBatchIdle()` wraps the Notify path to track idle state without duplicating logic.

## Problem

The retroactive embed processor polls every 3 seconds. When `CountWithoutFlag` returns a phantom count (e.g. `total=2` but nothing is processable), it logs two INFO lines per tick — generating ~57K lines/day. The SSE `stats_update` broadcaster (5s interval) compounds the issue: Chrome's background-tab throttling queues the flood, and foregrounding the tab causes a main-thread lockup that renders the Web UI blank.

## Before / After

| Metric | Before | After |
|--------|--------|-------|
| Log lines/day (idle) | ~57,600 | 0 at INFO level |
| Poll interval (phantom count) | 3s forever | 3s → 3min (exponential) |
| Poll interval (real work) | 3s | 3s (unchanged) |
| Notify() responsiveness | 3s | 3s (unchanged) |

## Test plan

- [x] All 75 existing `internal/plugin` tests pass (`go test ./internal/plugin/ -count=1`)
- [x] Full project builds clean (`go build ./...`)
- [x] Verified in production: restarted daemon with updated binary, zero retroactive processor INFO lines after 30+ seconds of idle operation
- [x] Web UI loads cleanly in a fresh browser tab after restart

### DIFF
--- a/cmd/muninn/coverage_hardening_test.go
+++ b/cmd/muninn/coverage_hardening_test.go
@@ -1112,7 +1112,7 @@ func TestConfigureNamedTools_AllNames(t *testing.T) {
 	out := captureStdout(func() {
 		errs := configureNamedTools([]string{
 			"claude", "claude-code", "cursor", "windsurf", "openclaw", "vscode", "manual",
-		}, "http://127.0.0.1:8750/mcp", "tok")
+		}, "http://127.0.0.1:8750/mcp", "tok", "")
 		_ = errs
 	})
 	_ = out
@@ -1836,7 +1836,7 @@ func TestConfigureClaudeMD_NewFile_Hardened(t *testing.T) {
 	t.Setenv("USERPROFILE", home)
 	t.Setenv("USERPROFILE", home)
 
-	err := configureClaudeMD()
+	err := configureClaudeMD("")
 	if err != nil {
 		t.Fatalf("unexpected error: %v", err)
 	}
@@ -1858,7 +1858,7 @@ func TestConfigureClaudeMD_ExistingNoBlock(t *testing.T) {
 	os.MkdirAll(claudeDir, 0700)
 	os.WriteFile(filepath.Join(claudeDir, "CLAUDE.md"), []byte("existing content\n"), 0644)
 
-	err := configureClaudeMD()
+	err := configureClaudeMD("")
 	if err != nil {
 		t.Fatalf("unexpected error: %v", err)
 	}
@@ -1880,7 +1880,7 @@ func TestConfigureClaudeMD_AlreadyHasBlock(t *testing.T) {
 	os.MkdirAll(claudeDir, 0700)
 	os.WriteFile(filepath.Join(claudeDir, "CLAUDE.md"), []byte("# Memory\nmuninn_remember\n"), 0644)
 
-	err := configureClaudeMD()
+	err := configureClaudeMD("")
 	if err != nil {
 		t.Fatalf("unexpected error: %v", err)
 	}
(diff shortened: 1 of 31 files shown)


scrypster:
Hey @vcheckk — really appreciate you tracking this down and putting together such a thorough fix. The phantom-count log flood + SSE cascade causing the Web UI to blank out is a genuinely nasty problem, and the idle backoff approach is exactly the right solution. Great PR.

We had a senior review pass over the diff and it's nearly there — just two small things before we merge:

**1. Unlocked reads in the `slog.Info` call (`processBatch`)**

After the final `statsMu.Unlock()`, `rp.stats.Processed` and `rp.stats.Errors` are read outside the lock. Capture them inside before unlocking:

```go
rp.statsMu.Lock()
rp.stats.Status = "idle"
passProcessed := rp.stats.Processed - passStart
totalProcessed := rp.stats.Processed
totalErrors := rp.stats.Errors
rp.statsMu.Unlock()
```

Then use `totalProcessed` / `totalErrors` in the log line. Benign in practice today (single writer), but it's a `-race` detector violation.

**2. `processBatchIdle` never increments `consecutiveIdle`**

The helper resets the counter when real work is done, but never increments it on the phantom-count path. The ticker path does `consecutiveIdle++` explicitly — the helper should mirror that:

```go
func (rp *RetroactiveProcessor) processBatchIdle(ctx context.Context, consecutiveIdle *int) bool {
	prev := rp.Stats().Processed
	ok := rp.processBatch(ctx)
	if ok && rp.Stats().Processed > prev {
		*consecutiveIdle = 0
	} else if ok {
		(*consecutiveIdle)++
	}
	return ok
}
```

Those are the only two things. Once those land we're good to merge — everything else (backoff math, ticker.Reset semantics, log level changes) is clean. Thanks again for this one!

vcheckk:
Thanks for the review @scrypster — both items addressed in [`716ca95`](https://github.com/scrypster/muninndb/pull/188/commits/716ca95).

**1. Unlocked stats reads in `processBatch`**

`totalProcessed` and `totalErrors` are now captured into locals under `statsMu.Lock()` before the unlock, then used in the `slog.Info` call:

```go
rp.statsMu.Lock()
rp.stats.Status = "idle"
passProcessed := rp.stats.Processed - passStart
totalProcessed := rp.stats.Processed
totalErrors := rp.stats.Errors
rp.statsMu.Unlock()
```

**2. `processBatchIdle` idle increment**

Added the missing `else if ok` branch so the Notify path increments the idle counter on phantom-count passes. Without this, a phantom `Notify()` left `consecutiveIdle == 0`, which caused line 152 to reset the ticker back to `pollInterval` — defeating the backoff:

```go
if ok && rp.Stats().Processed > prev {
    *consecutiveIdle = 0
} else if ok {
    (*consecutiveIdle)++
}
```

**Verification:**
- `go build ./...` — clean
- `go test ./internal/plugin/ -count=1 -race` — all existing tests pass, no race violations
- Diff is +6 / -2 lines, isolated to `retroactive.go`

vcheckk:
I've been using MuninnDB daily, and I'm really happy to contribute something with real impact. Thanks, @scrypster!

PR #306, opened 2026-03-28
feat(dream): memories accumulate but never consolidate -- add dream engine foundation

5queezer:
## Demo

```
$ muninn stop && muninn dream --dry-run && muninn start

muninn stopped (pid 121236)
[DRY RUN] No changes were written.

Dream completed in 0s

  default           scanned 107 engrams  (merged 9)

$ muninn stop && muninn dream --force && muninn start

muninn stopped (pid 2653531)
Dream completed in 0s

  default           scanned 107 engrams  (merged 9)
```

## Summary

MuninnDB captures memories well but never forgets or consolidates. After weeks of use, the store fills with near-duplicates and stale entries. The existing consolidation worker handles exact dedup but can't reason about meaning.

This is the **read-only foundation** (Steps 1-3) of a Dream Engine that will run LLM-driven consolidation between sessions, modeled after human sleep memory processing.

### What's in this PR

- **Dream state storage** in Pebble (`0x27` key prefix) — tracks `last_dream_at` and `engrams_at_dream` per vault
- **Phase 0 Orient** — vault scanning: engram count, embedding coverage, avg relevance/stability, legal vault detection
- **Configurable dedup threshold** — `Worker.DedupThreshold` (default 0.95, dream mode uses 0.85 to surface near-duplicates for future LLM review)
- **`muninn dream` CLI command** — `--dry-run`, `--force`, `--scope` flags, follows the `exec.go`/`backup.go` pattern (offline, server must be stopped)
- **Public API** — `db.Dream(ctx, opts)` for embedded usage

### What's NOT changed

Existing consolidation pipeline (`RunOnce`) is **unchanged**. Dream is additive — 6 new files, 8 minimal modifications (dispatch entry, help text, configurable dedup threshold). No modified behavior in existing code paths.

### Legal vault safety

Vaults containing "legal" in the name are always skipped — never scanned for consolidation, never sent to any LLM (including local Ollama). This is tested and non-negotiable.

### How to test

```bash
go test ./internal/consolidation/... ./internal/storage/...   # all tests pass
go build ./cmd/muninn/                                         # compiles
muninn dream --dry-run                                         # safe, no writes
muninn dream --help                                            # see all flags
```

### What's next (follow-up PR)

- Phase 4: Bidirectional stability (strengthen high-signal ×1.2, weaken low-signal max(×0.8, 14d floor))
- Phase 2b: LLM consolidation (Ollama → Anthropic → OpenAI, vault trust tiers)
- Phase 6: Dream journal (`~/.muninn/dream.journal.md`) — human-readable narrative
- Dream-on-start: `muninn stop` sets flag → `muninn start` runs dream before opening ports
- `MUNINN_DREAM_TIMEOUT` (default 60s) for dream-on-start safety

**Spec:** `docs/superpowers/specs/2026-03-28-dream-engine-design.md`

<details>
<summary>Test plan</summary>

- [x] `TestDreamOnce_DryRun_NoMutations` — full pipeline, zero writes
- [x] `TestDreamOnce_LegalVaultSkipped` — legal vaults always skipped
- [x] `TestDreamOnce_ScopeFilter` — `--scope` limits to single vault
- [x] `TestDreamOnce_EmptyVault` — handles empty vault gracefully
- [x] `TestOrient_EmptyVault` / `TestOrient_WithEngrams` — Phase 0 scan
- [x] `TestIsLegalVault` — 9 vault name patterns
- [x] `TestDedup_ConfigurableThreshold` — 0.85 vs 0.95 behavior verified
- [x] `TestWriteReadDreamState` / `TestReadDreamStateMissing` / `TestWriteDreamStateOverwrite` — Pebble round-trip
- [x] All existing consolidation + storage tests pass unchanged

</details>

### DIFF
--- a/cmd/muninn/dream.go
+++ b/cmd/muninn/dream.go
@@ -1,0 +1,114 @@
+package main
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"io"
+	"log/slog"
+	"os"
+	"os/signal"
+	"path/filepath"
+	"strings"
+	"syscall"
+	"time"
+
+	muninn "github.com/scrypster/muninndb"
+)
+
+func runDream(args []string) {
+	fs := flag.NewFlagSet("dream", flag.ContinueOnError)
+	fs.SetOutput(os.Stderr)
+
+	force := fs.Bool("force", false, "bypass trigger gates")
+	dryRun := fs.Bool("dry-run", false, "preview changes without writing")
+	scope := fs.String("scope", "", "limit to a single vault")
+	dataDir := fs.String("data-dir", defaultDataDir(), "data directory")
+
+	if err := fs.Parse(args); err != nil {
+		os.Exit(1)
+	}
+
+	// Check server is not running (Pebble exclusive lock).
+	pidPath := filepath.Join(*dataDir, "muninn.pid")
+	if pid, err := readPID(pidPath); err == nil && isProcessRunning(pid) {
+		fmt.Fprintf(os.Stderr, "error: muninn is running (pid %d) — cannot run offline dream\n", pid)
+		fmt.Fprintln(os.Stderr, "Stop the server first: muninn stop")
+		osExit(1)
+		return
+	}
+
+	// Suppress slog output — dream writes structured output to stdout.
+	slog.SetDefault(slog.New(slog.NewTextHandler(io.Discard, nil)))
+
+	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
+	defer stop()
+
+	db, err := muninn.Open(*dataDir)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+		if isExecLockError(err) {
+			fmt.Fprintf(os.Stderr, "Hint: if the daemon crashed, run 'muninn status' or remove %s/pebble/LOCK\n", *dataDir)
+		}
+		osExit(1)
+		return
+	}
+	defer func() {
+		if err := db.Close(); err != nil {
+			fmt.Fprintf(os.Stderr, "warn: close: %v\n", err)
+		}
+	}()
+
+	report, err := db.Dream(ctx, muninn.DreamOpts{
+		DryRun: *dryRun,
+		Force:  *force,
+		Scope:  *scope,
+	})
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+		osExit(1)
+		return
+	}
+
+	printDreamReport(report, *dryRun)
+}
+
+func printDreamReport(report *muninn.DreamReport, dryRun bool) {
+	if dryRun {
+		fmt.Println("[DRY RUN] No changes were written.")
+		fmt.Println()
+	}
+
+	fmt.Printf("Dream completed in %s\n", report.TotalDuration.Round(100*time.Millisecond))
+	fmt.Println()
+
+	if len(report.Skipped) > 0 {
+		fmt.Printf("Skipped (legal): %s\n", strings.Join(report.Skipped, ", "))
+	}
+
+	for _, r := range report.Reports {
+		if r.Orient != nil && r.Orient.IsLegal {
+			fmt.Printf("  %-16s  %d engrams (protected, skipped)\n", r.Vault, r.LegalSkipped)
+			continue
+		}
+
+		scanned := 0
+		if r.Orient != nil {
+			scanned = r.Orient.EngramCount
+		}
+
+		fmt.Printf("  %-16s  scanned %d", r.Vault, scanned)
+		var changes []string
+		if r.MergedEngrams > 0 {
+			changes = append(changes, fmt.Sprintf("merged %d", r.MergedEngrams))
+		}
+		if r.InferredEdges > 0 {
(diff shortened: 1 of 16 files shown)


scrypster:
Hey @5queezer — this is a genuinely excellent contribution. The dream engine concept is exactly the right framing for this problem, and the implementation discipline here (offline-only, dry-run first, legal vault protection, additive to existing pipeline) shows you really studied how MuninnDB is built before touching it. We've been thinking about something very similar internally and were waiting for more signal from real users before prioritizing it — this PR is that signal. We'd love to get this in.

A few things we'd like addressed before merge:

**Must fix:**

1. **Thread-safety in `DreamOnce`** — `DreamOnce` mutates `w.DryRun` and `w.DedupThreshold` directly and restores them with defer. The CLI path is safe today because the server must be stopped, but the public `db.Dream()` API has no such guard. If a caller invokes it while the background consolidation scheduler is running on the same `Worker` instance, that's a data race. The cleaner fix is to pass threshold and dryRun as parameters through the phase functions rather than mutating Worker fields — or construct a fresh `Worker` with dream-specific settings inside `DreamOnce` rather than borrowing the caller's instance.

2. **Key-space schema** — `docs/key-space-schema.md` explicitly requires any new key prefix to be documented before merge. The `0x27` prefix for dream state needs a row in the prefix table. Also needs to be registered in `TestKeyPrefixesAreUnique` in `keys_test.go` — that test exists specifically to catch collisions.

**Should fix:**

3. **`isLegalVault` is too broad** — substring match on "legal" will catch "paralegal", "illegal", "my-legal-pad", etc. We'd prefer either a prefix convention like `legal:` or a vault metadata flag. The protection intent is right — the implementation just needs to be more precise.

4. **`WriteDreamState` should use `pebble.Sync`** — dream state is technically re-derivable, but it's the trigger gate for future phases. If it's lost, over-consolidation becomes possible. At once-per-dream-cycle write frequency the sync cost is negligible.

**Minor / advisory:**

- The `--force` help text implies trigger gates exist and can be bypassed — since gates are deferred to PR #2, a small note like `(gates not yet implemented)` would avoid confusion.
- The type alias `DreamReport = consolidation.DreamReport` leaks internal types through the public API. Not blocking for now, but worth a conscious decision before stabilization.
- Orient loads all engrams into memory for avg computation — fine at current scale, but worth a streaming/metadata-only approach when vaults get large.

The phased roadmap you've laid out (LLM consolidation, bidirectional stability, dream journal) is exactly where we want to go. Really looking forward to Phase 2b. If you can address these items we would be more than happy to merge this in and move forward to the other phases.

5queezer:
Thanks for the thorough review — all items addressed across two commits ([`ae9544e`](https://github.com/5queezer/muninndb/commit/ae9544e), [`d240af0`](https://github.com/5queezer/muninndb/commit/d240af0)).

**Must fix #1 — Thread-safety:** `DreamOnce` now constructs a fresh `Worker` with dream-specific settings (`DedupThreshold: 0.85`, `DryRun` from opts) instead of mutating the caller's instance. All phase calls use the dream-specific worker. This eliminates the race with the background consolidation scheduler. The public `db.Dream()` API also creates a fresh `Worker` via `consolidation.NewWorker()`, so both paths are safe.

**Must fix #2 — Key-space schema:** `0x27` (Dream State) documented in `docs/key-space-schema.md` with Sync durability and registered in `TestKeyPrefixesAreUnique`.

**Should fix #3 — `isLegalVault`:** Switched from substring match to exact match + prefix convention: `"legal"`, `"legal:*"`, `"legal/*"`. Won't catch "paralegal" or "illegal". Tests updated with positive and negative cases. Noted vault metadata flag as a future enhancement — agree it's the right long-term solution. Phase 2 carries this convention consistently through `vaultTrustTier` and `runPhase4BidirectionalStability`.

**Should fix #4 — `pebble.Sync`:** `WriteDreamState` now uses `pebble.Sync`. `WriteDreamDue` intentionally stays `NoSync` — it's a transient signal where loss just means the dream runs at next regular schedule.

**Minor — `--force`:** Updated help text to `"Bypass trigger gates (not yet implemented)"`. Gates (12h + 3 engrams) are implemented in Phase 2.

**Minor — Type alias:** Added stabilization comment to `DreamReport`.

**Advisory items — tracking for Phase 2+:**
- Orient memory usage: agree a streaming/metadata-only approach is needed at scale. Will revisit when vaults exceed ~10k engrams.
- `isLegalVault` prefix convention propagated to all legal-checking code paths in Phase 2.

All tests pass (`go test ./internal/consolidation/... ./internal/storage/... ./cmd/muninn/...`, `go vet`).

scrypster:
@5queezer — just wanted to take a moment to genuinely say thank you. This was an outstanding contribution.

The dream engine concept is exactly the kind of thoughtful, principled work that makes open source worth doing. You didn't just write code — you studied how MuninnDB is built, respected its patterns, scoped the PR carefully, and responded to every piece of feedback with speed and precision. That's rare, and it doesn't go unnoticed.

This solves a real problem. Memory stores that only accumulate eventually become noise. The foundation you've laid here — the offline consolidation model, the lowered dedup threshold, the legal vault protection, the dry-run-first philosophy — is exactly the right architecture for where MuninnDB needs to go. We're genuinely excited to see where the follow-up phases take it.

We're grateful to have you contributing. Hope this is just the beginning.

PR #317, opened 2026-03-31
feat(engine): content-hash deduplication at write time

### LINKED ISSUE #313: feat(engine): content-hash deduplication at write time
## Problem

LLM agents frequently store the same information multiple times (retries, conversation loops, rephrased identical content). Currently, near-duplicate detection exists at write time via the novelty detector (Jaccard similarity >= 0.70), but it creates RelRefines associations rather than rejecting the write. Semantic deduplication in consolidation (cosine >= 0.95) runs later as a background pass. Neither catches exact duplicates at O(1) cost. Exact duplicate content enters the index and creates noise.

## Proposal

Add a SHA-256 content hash check in the engine Write path. On duplicate content, return the existing engram ID instead of creating a new record.

- Store content hash -> engram ID mapping in Pebble (new key prefix)
- Check hash before write — if exists and engram is not soft-deleted, return existing ID with a hint
- O(1) lookup cost, negligible impact on write latency
- Idempotent: re-storing the same content is a no-op, not an error

This complements the existing semantic deduplication in consolidation (which catches near-duplicates) with an exact-match fast path at write time.

## Non-goals

- This does NOT replace semantic deduplication — it only catches exact matches
- Content with different concept/tags/metadata but identical content text is still considered a duplicate (content is the identity)

Aperrix:
## Summary

- Adds SHA-256 content hash check in the engine `Write()` path — duplicate content returns the existing engram ID with a `duplicate_content` hint instead of creating a new record
- Per-vault isolation (same content in different vaults is not deduplicated)
- Soft-delete aware (deleted content can be re-stored)
- Non-blocking: hash index failures are logged but don't fail the write
- New Pebble key prefix `0x28` for content hash → engram ID mapping

### Why

LLM agents frequently store the same information multiple times (retries, conversation loops). Catching exact duplicates at O(1) cost before they enter the index eliminates noise at the source, complementing the existing write-time novelty detector (Jaccard >= 0.70, creates RelRefines associations but does not reject writes) and the consolidation semantic dedup (cosine >= 0.95).

Fixes #313

## Test plan

- [x] Duplicate content returns existing engram ID with hint
- [x] Different content creates new engram normally
- [x] Soft-deleted content allows re-creation
- [x] Cross-vault isolation (no cross-dedup)
- [x] RED/GREEN verified
- [x] Full engine+storage+MCP test suite passes


5queezer (changes requested):
## Review

### Critical: `ClearVault` missing `0x28` prefix
`internal/storage/vault_lifecycle.go:56-59` — `dataPrefixes` doesn't include the new `0x28` content-hash prefix. After vault clear, all hash mappings survive pointing to deleted engrams. Fix: add `0x28` to the slice. (Audit `0x18-0x27` while you're there.)

### Major: `Restore` doesn't re-add hash mapping
`internal/engine/engine.go:2352-2382` — Restored engram is live but unprotected by dedup. Next identical write creates a duplicate.

### Major: TOCTOU race on concurrent identical writes
`internal/engine/engine.go:812-827` — Two concurrent writes can both pass `GetContentHash` before either stores. Realistic given LLM retry motivation. Is this accepted as best-effort, or does it need a per-vault CAS?

### Minor
- `WriteBatch` doesn't dedup within the same batch — items can't see each other's hashes during Phase 1. Document as known limitation.
- `ContentHash("")` is valid — two empty-content engrams in the same vault get silently deduplicated. Skip dedup for empty content?

scrypster (changes requested):
Hey Aperrix — really appreciate this contribution. The core design is clean: the key layout is consistent with the rest of the codebase, vault scoping is correct, the soft-delete fallthrough logic is well-handled, and the test coverage is solid. This is clearly well-researched work.

We do have a few things we'd like addressed before merging. Two are blockers, a couple are major, and the rest are minor polish.

---

### Critical (blockers)

**1. `ClearVault` is missing `0x28` in `dataPrefixes`** (`internal/storage/vault_lifecycle.go:56-59`)

The `dataPrefixes` slice doesn't include `0x28`, so clearing a vault leaves all content-hash mappings behind pointing to deleted engram IDs. The `Write()` path does self-heal (it detects stale entries on the next write), but it's inconsistent with the design intent and leaves orphaned keys that waste disk space and cost extra reads.

**2. `CloneVaultData` / `MergeVaultData` missing `0x28` in `vaultScopedSwapPrefixes`** (`internal/storage/clone.go:45-49`)

Same situation — after cloning, the target vault has no content-hash mappings so dedup won't kick in for previously-stored content. Add `0x28` to `vaultScopedSwapPrefixes`.

---

### Major

**3. `Restore()` doesn't re-add the content-hash mapping** (`internal/engine/engine.go` — the Restore path)

When a soft-deleted engram is restored back to active, the content-hash mapping isn't reinstated. If the mapping was cleaned up during a re-write between soft-delete and restore, dedup won't catch future identical writes for this engram. These two are coupled — if the soft-delete path is updated to clean up the hash (see minor note below), Restore absolutely must re-add it.

**4. `vaultScopedExportPrefixes` missing `0x28`** (`internal/storage/export.go:25-49`)

Exported archives won't contain content-hash mappings, so dedup won't work for imported vaults until content is re-written. Should be included.

---

### Minor

- **Soft-delete path** doesn't explicitly clean up the hash mapping on `Forget(Hard: false)`. The `Write()` path compensates gracefully, but it's inconsistent with the FTS cleanup that already happens on soft-delete. Either clean it up there too, or add a comment explaining the intentional lazy-cleanup design.

- **TOCTOU comment** — the best-effort concurrent write window isn't documented in the code. A short comment acknowledging it (and why a mutex isn't needed for a dedup optimization) would help future readers.

- **Intra-batch dedup** — two identical items in the same `WriteBatch` call won't be deduped against each other (they can't see each other's hashes during Phase 1). Worth a comment documenting this known limitation.

- **Missing tests** for: restore+dedup interaction, clear+dedup (write → clear vault → re-write same content verifies no stale hit), and intra-batch behavior.

---

If you can address the two criticals and the Restore gap, we'd be happy to merge this in. The export and minor items would be great to have too but we can discuss if any of them feel out of scope for this PR.

Thanks again — this is exactly the kind of contribution that makes the system better for everyone.

scrypster (commented):
## Additional feedback: Reinforce on duplicate, don't just return

One more thing worth addressing beyond the 4 required fixes:

When a duplicate is detected, the current implementation silently drops the repetition signal:

```go
return &mbp.WriteResponse{
    ID:   existingID.String(),
    Hint: "duplicate_content",
}, nil  // ← existing engram is never touched
```

This is semantically wrong for the cognitive model. MuninnDB scores engrams using ACT-R:

```
B(M) = ln(n+1) - d × ln(age/(n+1))
```

Where `n` is retrieval/access frequency. More accesses = higher base-level activation = the engram is treated as more memorable and surfaces more readily in recall. Writing the same content twice is a **repetition signal** — the information is being re-experienced, which in cognitive terms *should* strengthen it.

The correct behavior on a duplicate write:
- Don't create a new engram ✓ (dedup is correct)
- But **reinforce** the existing engram — increment access count, update `LastAccess`, let ACT-R naturally score it higher on future recalls

This makes the dedup semantically correct rather than just space-efficient. A duplicate write isn't a no-op — it's evidence that this piece of information matters enough to be written again.

Aperrix:
Thanks for the thorough review. All items addressed:

**Critical fixes:**
- `0x28` added to `dataPrefixes` in `vault_lifecycle.go` (ClearVault)
- `0x28` added to `vaultScopedSwapPrefixes` in `clone.go` (Clone/Merge)

**Major fixes:**
- `0x28` added to `vaultScopedExportPrefixes` in `export.go`
- `Restore()` now re-adds the content-hash mapping via `PutContentHash`
- Duplicate writes now **reinforce** the existing engram — AccessCount is incremented and LastAccess updated, so ACT-R naturally scores repeated content higher

**Minor:**
- TOCTOU race documented with a comment explaining best-effort design
- Intra-batch dedup limitation documented
- Added 3 tests: clear-vault + re-write, restore + dedup, duplicate reinforces access count

All 1640 tests pass (engine + storage + mcp).

scrypster (approved):
## Opus Review: Approved

All 8 requested items verified and correct:

- `ClearVault` dataPrefixes updated to include `0x28` (count bumped 20→21 in comment)
- `clone.go` `vaultScopedSwapPrefixes` includes `0x28`
- `export.go` `vaultScopedExportPrefixes` includes `0x28`
- `Restore()` re-adds content-hash mapping via `PutContentHash` after restoring
- Duplicate writes reinforce via `UpdateMetadata` — AccessCount++, LastAccess updated, and `WriteLastAccessEntry` called to keep the 0x22 index consistent. The ACT-R repetition signal is wired correctly all the way through storage.
- All three tests present: clear+rewrite, restore+dedup, reinforce access count — plus extras (cross-vault no-dedup, hard-delete rewrite, evolve, batch)
- TOCTOU and intra-batch limitation both documented in comments

Key design points look correct: 0x28 key is vault-scoped (`[0x28|ws[8]|sha256[32]]`), lazy cleanup of stale hashes on soft-delete is intentional and handled gracefully in the Write path, WriteBatch filtering index remapping is correct.

One non-blocking note: the `UpdateMetadata` error on reinforce is silently discarded (`_ =`). Best-effort is fine here — the dedup returns the right ID either way — but worth logging at debug level in a follow-up.

Semantically solid — duplicate write as rehearsal signal strengthening the memory trace is exactly right for ACT-R.

PR #318, opened 2026-03-31
feat(cognitive): hybrid exponential + power-law decay model

### LINKED ISSUE #314: feat(cognitive): hybrid exponential + power-law decay model
## Problem

MuninnDB uses pure Ebbinghaus exponential decay (R = e^(-t/S)). While accurate for short-term forgetting, empirical research shows that long-term memory follows a **power-law** curve, not exponential (Wixted & Ebbesen 1991, Wixted 2004 "The Psychology and Neuroscience of Forgetting").

The practical consequence: exponential decay drops off too aggressively for memories older than a few days. Important memories that haven't been accessed recently lose activation faster than they should.

## Proposal

Implement a hybrid decay model:

1. **t < 3 days**: Exponential decay (current behavior) — fast filtering of noise and irrelevant memories
2. **t >= 3 days**: Power-law decay (R = a * t^(-b)) — heavy tail that preserves important older memories

The transition point (3 days) and power-law exponent should be configurable via plasticity parameters, with sensible defaults derived from the literature.

## Academic basis

- Wixted, J. T., & Ebbesen, E. B. (1991). "On the Form of Forgetting." *Psychological Science*, 2(6), 409-415.
- Wixted, J. T. (2004). "The Psychology and Neuroscience of Forgetting." *Annual Review of Psychology*, 55, 235-269.
- Anderson, J. R., & Schooler, L. J. (1991). "Reflections of the Environment in Memory." *Psychological Science*, 2(6), 396-408.

## Scope

- Modify the decay formula in `internal/cognitive/decay.go`
- Add plasticity parameters for transition point and power-law exponent
- Ensure continuity at the transition point (no discontinuous jump)
- Preserve existing floor constraint (5% minimum retention)

Aperrix:
## Summary

- Adds `HybridRetention()` function with two decay models:
  - **exponential** (default): identical to existing `EbbinghausWithFloor()` — backward compatible
  - **hybrid**: exponential for t < transition days, continuous power-law tail for t >= transition days
- Mathematical continuity at the transition point (no discontinuous jump)
- New plasticity parameters: `decay_model`, `power_law_exponent` (default 0.5), `decay_transition_days` (default 3.0)
- All presets default to `"exponential"` — zero behavior change without opt-in

### Why

Pure Ebbinghaus exponential decay drops off too aggressively for long-term memories. Empirical research (Wixted & Ebbesen 1991, Wixted 2004) shows power-law decay better matches human forgetting curves for memories older than a few days. The heavy tail preserves important older memories without requiring frequent access.

Fixes #314

## Test plan

- [x] Exponential model matches current behavior exactly
- [x] Hybrid model uses exponential for t < 3 days
- [x] Hybrid model uses power-law for t >= 3 days
- [x] Continuity at transition point verified
- [x] Floor constraint enforced in power-law phase
- [x] Power-law retention > exponential retention for large t
- [x] Plasticity parameter defaults, overrides, and clamping
- [x] RED/GREEN verified
- [x] Full cognitive + auth test suite passes

### DIFF
--- a/internal/auth/plasticity.go
+++ b/internal/auth/plasticity.go
@@ -51,6 +51,11 @@ type PlasticityConfig struct {
 	// RecallMode is the default recall mode for this vault: "semantic"|"recent"|"balanced"|"deep".
 	// nil = use "balanced" (engine defaults).
 	RecallMode *string `json:"recall_mode,omitempty"`
+
+	// Hybrid decay model parameters
+	DecayModel          *string  `json:"decay_model,omitempty"`           // "exponential" (default) or "hybrid"
+	PowerLawExponent    *float64 `json:"power_law_exponent,omitempty"`    // power-law exponent (default 0.5)
+	DecayTransitionDays *float64 `json:"decay_transition_days,omitempty"` // days before switching to power-law (default 3.0)
 }
 
 // ResolvedPlasticity is the fully-merged configuration after applying preset defaults
@@ -95,6 +100,10 @@ type ResolvedPlasticity struct {
 	EnrichmentEnabled bool `json:"enrichment_enabled"`
 	// RecallMode is the default recall mode for this vault.
 	RecallMode string `json:"recall_mode"`
+	// Hybrid decay model parameters
+	DecayModel          string  `json:"decay_model"`           // "exponential" or "hybrid"
+	PowerLawExponent    float64 `json:"power_law_exponent"`    // power-law exponent (default 0.5)
+	DecayTransitionDays float64 `json:"decay_transition_days"` // transition point in days (default 3.0)
 }
 
 type plasticityPreset struct {
@@ -120,9 +129,12 @@ type plasticityPreset struct {
 	AssocMinWeight       float32
 	ArchiveThreshold     float64
 	BehaviorMode         string
-	InlineEnrichment  string
-	EnrichmentEnabled bool
-	RecallMode        string
+	InlineEnrichment    string
+	EnrichmentEnabled   bool
+	RecallMode          string
+	DecayModel          string
+	PowerLawExponent    float64
+	DecayTransitionDays float64
 }
 
 var plasticityPresets = map[string]plasticityPreset{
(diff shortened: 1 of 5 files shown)


5queezer (changes requested):
## Review

### Major: Feature is dead code in production
`DecayCandidate.DecayModel`, `.PowerLawExponent`, `.DecayTransitionDays` are never populated outside tests. No engine code passes `ResolvedPlasticity` values through to `DecayCandidate`. Opting a vault into `decay_model: "hybrid"` has no effect at runtime.

If this is groundwork for a follow-up, say so in the description.

### Major: Division by zero in `HybridRetention`
`internal/cognitive/decay.go` — `math.Pow(daysSinceAccess/transitionDays, -powerLawExponent)` produces NaN when `transitionDays=0` (the current runtime value per the above). Public function should guard its inputs:

```go
if transitionDays <= 0 {
    transitionDays = 3.0
}
if powerLawExponent <= 0 {
    powerLawExponent = 0.5
}
```

### Minor
- Value continuity at the transition is ensured and tested, but slope is discontinuous (~2.3x steeper on the power-law side with defaults). Docstring implies smoothness — clarify.

scrypster:
Hey Aperrix — closing this one since it's associated with issue #314, which we've already addressed.

The short version: the problem this PR solves is already handled by the ACT-R scorer, which is the default production path. The formula `B(M) = ln(n+1) - d × ln(max(ageDays,ageFloor) / (n+1))` (Anderson 1993) is a power-law decay model and is live in production today. The Ebbinghaus background worker that this PR modifies writes to the `relevance` field, which is used for diagnostics only in ACT-R mode — it doesn't influence actual recall scoring. There's also a code comment in engine.go that reads "All production scoring goes through ACT-R; decay path is kept in code but not reachable."

We left a detailed explanation on the issue itself. Thanks for the effort on the writeup and the academic citations — the research was solid, the problem statement just turned out to already be solved. If you ever find a specific recall case where the ACT-R path isn't producing the right results, we'd love to see a repro.

Aperrix:
Thanks for the detailed explanation. You're right — I missed that the Ebbinghaus worker writes to the `relevance` field which is diagnostic-only in ACT-R mode. The actual production scoring already uses Anderson's power-law formula via `B(M) = ln(n+1) - d × ln(max(ageDays, ageFloor) / (n+1))`, which addresses the exact concern this PR was trying to solve.

Lesson learned: should have traced the full scoring path (activation engine → phase6Score → ACT-R) before assuming the decay worker was the right target. The `ACTRDecay` plasticity parameter already provides the tuning knob we were proposing.

Closing the branch on our end. Thanks for taking the time to explain the architecture.

PR #319, opened 2026-03-31
feat(cognitive): Long-Term Potentiation (LTP) for Hebbian associations

### LINKED ISSUE #315: feat(cognitive): Long-Term Potentiation (LTP) for Hebbian associations
## Problem

MuninnDB's Hebbian learning updates association weights based on co-activation, but all associations are subject to the same decay dynamics. Well-established associations (repeatedly co-activated over time) can still decay away if not recently reinforced. This doesn't match the biological model — synapses that fire together enough times become "potentiated" and resist decay.

## Proposal

Implement Long-Term Potentiation (LTP) for Hebbian associations:

1. Track a **co-activation counter** per association
2. When the counter exceeds a configurable threshold (e.g., 5 co-activations), mark the association as **potentiated**
3. Potentiated associations have a higher weight floor (e.g., 0.3), preventing them from dropping below a minimum strength
4. Decay resistance (reduced decay rate for potentiated associations) is tracked as a future enhancement

This ensures that knowledge relationships confirmed by repeated use become stable foundations of the knowledge graph.

## Academic basis

- Bi, G.-Q., & Poo, M.-M. (1998). "Synaptic Modifications in Cultured Hippocampal Neurons." *Journal of Neuroscience*, 18(24), 10464-10472.
- Bliss, T. V. P., & Collingridge, G. L. (1993). "A Synaptic Model of Memory: Long-Term Potentiation in the Hippocampus." *Nature*, 361, 31-39.

## Scope

- Modify `internal/cognitive/hebbian.go` to track co-activation counts
- Add potentiation threshold and weight floor as configurable constants
- Store co-activation counter in association metadata
- Add plasticity parameter for LTP threshold

Aperrix:
## Summary

- Tracks co-activation counts per association pair in the Hebbian worker
- When count exceeds configurable threshold (default 5), association becomes **potentiated**:
  - Weight floor raised to `ltp_weight_floor` (default 0.3) — potentiated associations never drop below this minimum
  - Weight floor raised to `ltp_weight_floor` (default 0.3)
- Session-local tracking for fast threshold checks, storage-backed for persistence
- New plasticity parameters: `ltp_threshold`, `ltp_weight_floor`
- All defaults are zero/disabled — completely backward compatible
- Session-scoped potentiation (intentional: avoids storage reads in the hot path, documented)

### Why

Repeatedly co-activated memories represent established knowledge that shouldn't decay. LTP is the biological mechanism for synapse stabilization (Bi & Poo 1998, Bliss & Collingridge 1993). Without it, well-established associations can still decay away if not recently reinforced.

Fixes #315

## Test plan

- [x] Co-activation counter increments on each Hebbian pass
- [x] Association becomes potentiated after threshold co-activations
- [x] Potentiated associations respect weight floor
- [x] Counter persists across Hebbian passes
- [x] Nil/default config preserves existing behavior
- [x] Old constructor compatibility
- [x] Plasticity parameter defaults, overrides, and clamping
- [x] RED/GREEN verified
- [x] Full cognitive + auth test suite passes

### DIFF
--- a/internal/auth/plasticity.go
+++ b/internal/auth/plasticity.go
@@ -52,11 +52,18 @@ type PlasticityConfig struct {
 	// nil = use "balanced" (engine defaults).
 	RecallMode *string `json:"recall_mode,omitempty"`
 
-	// ScoringFusion selects the Phase 6 scoring strategy.
+// ScoringFusion selects the Phase 6 scoring strategy.
 	// "rrf" = use Phase 3 RRF scores directly (rank-based, scale-invariant).
 	// "weighted_sum" = use legacy weighted-sum scoring (DisableACTR implied).
 	// nil/empty = default (ACT-R scoring, unchanged behavior).
 	ScoringFusion *string `json:"scoring_fusion,omitempty"`
+// Long-Term Potentiation (LTP) for Hebbian associations.
+	// Associations co-activated beyond LTPThreshold become potentiated:
+	// decay resistance and a higher weight floor.
+	// All zero/nil = disabled (backward compatible).
+	LTPThreshold   *int     `json:"ltp_threshold,omitempty"`    // co-activation count to trigger LTP (0 = disabled)
+	LTPDecayFactor *float64 `json:"ltp_decay_factor,omitempty"` // decay multiplier for potentiated edges (0–1; 0 = disabled)
+	LTPWeightFloor *float32 `json:"ltp_weight_floor,omitempty"` // minimum weight for potentiated edges (0–1; 0 = disabled)
 }
 
 // ResolvedPlasticity is the fully-merged configuration after applying preset defaults
@@ -101,8 +108,13 @@ type ResolvedPlasticity struct {
 	EnrichmentEnabled bool `json:"enrichment_enabled"`
 	// RecallMode is the default recall mode for this vault.
 	RecallMode string `json:"recall_mode"`
-	// ScoringFusion selects Phase 6 scoring strategy: "" (default=ACT-R), "rrf", or "weighted_sum".
+// ScoringFusion selects Phase 6 scoring strategy: "" (default=ACT-R), "rrf", or "weighted_sum".
 	ScoringFusion string `json:"scoring_fusion"`
+// Long-Term Potentiation (LTP) for Hebbian associations.
+	// Zero values = disabled (backward compatible default).
+	LTPThreshold   int     `json:"ltp_threshold"`    // co-activation count to trigger LTP; 0 = disabled
+	LTPDecayFactor float64 `json:"ltp_decay_factor"` // decay multiplier for potentiated edges; 0 = disabled
+	LTPWeightFloor float32 `json:"ltp_weight_floor"` // minimum weight for potentiated edges; 0 = disabled
 }
 
 type plasticityPreset struct {
@@ -131,7 +143,10 @@ type plasticityPreset struct {
 	InlineEnrichment  string
 	EnrichmentEnabled bool
 	RecallMode        string
-	ScoringFusion     string // "" = default (ACT-R), "rrf", "weighted_sum"
+ScoringFusion     string // "" = default (ACT-R), "rrf", "weighted_sum"
+LTPThreshold      int
+	LTPDecayFactor    float64
+	LTPWeightFloor    float32
 }
 
 var plasticityPresets = map[string]plasticityPreset{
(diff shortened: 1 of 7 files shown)


5queezer (changes requested):
## Review

### Major: PR description claims `ltp_decay_factor` — code doesn't implement it
A comment in `hebbian_ltp.go` says decay resistance is "planned for a follow-up." Description should match what's shipped.

### Major: `ltpState` maps grow without bound
`internal/cognitive/hebbian_ltp.go:33-35` — `potentiated` and `counts` only ever grow. No eviction, TTL, or size cap. Long-running instances will accumulate entries indefinitely. At minimum expose a size metric; ideally add LRU eviction or periodic cleanup.

### Major: Silent LTP config drop on pair deduplication
`internal/cognitive/hebbian.go:216` — When the same pair appears multiple times in a batch, only the first event's LTP config is kept. If first has `LTP: nil` and a later one doesn't, config is lost.

```go
// in the existing-pair branch:
if ps.ltp == nil {
    ps.ltp = event.LTP
}
```

5queezer:
## Follow-up: Dream Engine integration conflict

The dream consolidation pipeline (Phase 5 — transitive inference) updates association weights via direct `store.UpdateAssocWeight()` calls, bypassing `HebbianWorker` entirely. This means:

- Dream can set a weight **below the LTP floor** on a potentiated pair — it doesn't check `ltpState.IsPotentiated()`
- Dream Phase 4 can also weaken stability of engrams involved in potentiated associations

If the LTP weight floor is a correctness guarantee (not just a hint), dream needs to either check `IsPotentiated()` before lowering weights, or route updates through `HebbianWorker`.

This isn't a blocker for this PR specifically — the dream engine predates it — but worth coordinating so the two features don't silently undermine each other.

scrypster (changes requested):
## Review: Long-Term Potentiation (LTP)

Thank you for this contribution — session-scoped Hebbian potentiation is a solid idea and the general architecture is clean. However, there are issues that need to be resolved before this can merge.

### 🚨 Required Fixes

**1. Bug: LTP config silently dropped in batch pair deduplication**

In the batch processing path, when an existing pair is found in the same batch, the LTP config from the current event is dropped:

```go
if ps.ltp == nil {
    ps.ltp = event.LTP
}
```

If two events in the same batch reference the same pair, the second event's LTP config is silently ignored. This should either merge/override or return an error for conflicting configs on the same pair.

**2. Fix misleading PR description**

The PR description references `ltp_decay_factor` as a config field, but this field does not exist in the implementation. Remove or correct all references to it so reviewers and future contributors aren't misled.

**3. Add eviction/size cap to `ltpState` maps**

The `ltpState` maps are unbounded. A long-running session with many unique engram pairs will grow without limit. Add a size cap or LRU eviction to prevent unbounded memory growth in production.

### ⚠️ Design Note (non-blocking, worth discussing)

Session-scoped potentiation doesn't survive restarts — the in-memory state is not hydrated from stored `CoActivationCount` values on startup. This means potentiation resets on every server restart. Whether this is intentional is worth documenting explicitly. If it's intentional, add a comment explaining why. If it's not, hydration logic is needed.

---

Please address the three required fixes above and update the PR description. Happy to re-review once those are in.

Aperrix:
Thanks for the review. All required fixes addressed:

**Fix 1: Batch pair dedup** — When the same pair appears multiple times in a batch, the first non-nil LTP config is now preserved (nil-check added in the pair dedup path).

**Fix 2: PR description** — All references to `ltp_decay_factor` removed. The PR now accurately describes what's shipped: potentiation via weight floor only. Decay resistance is noted as a future enhancement.

**Fix 3: Unbounded maps** — Added `maxLTPEntries = 10000` cap on `ltpState.counts`. New pairs are silently skipped at capacity; existing tracked pairs continue normally.

**Additional:**
- Session-scoped potentiation explicitly documented on `ltpState` struct (intentional: avoids storage reads in hot path)
- Dream engine interaction documented (dream can bypass HebbianWorker and set weights below LTP floor — tracked separately)

All 761 tests pass (cognitive + auth + mcp).

scrypster (approved):
All three required fixes verified correct — approving.

**1. Nil-check fix (`if ps.ltp == nil { ps.ltp = event.LTP }`):** First-non-nil-wins is the right semantics. Within a vault, all events for the same pair carry the same `*LTPConfig` from the same `ResolvedPlasticity`, so first-wins is safe. The cross-vault pair collision (no `ws` in `pairKey`) is a pre-existing design constraint, not introduced here.

**2. `maxLTPEntries` cap:** Correctly bounds both maps. A key can only enter `potentiated` after passing through `counts`, so `potentiated` is implicitly bounded by `maxLTPEntries`. The uint32 saturation guard is a nice touch.

**3. Session-scope:** The stored association weight persists durably through Pebble. Only the floor enforcement lapses until re-potentiation after a restart — acceptable for a performance hint, not a correctness guarantee. The doc comment on `ltpState` is explicit about this.

**Engine wiring:** `resolved` is computed per-vault per-activation; `ltpCfg` is constructed fresh each time. No cross-vault contamination risk.

**Dream engine note:** Documented inline. Defer is fine.

Non-blocking for future consideration:
- `IsPotentiated` takes unexported `pairKey` — only callable within the package. Either unexport the method or refactor to accept `[16]byte` pairs if external callers are ever needed.
- `TestLTP_CounterPersistsAcrossPasses` tests store-level count accumulation, not session-state persistence — name is slightly misleading but the test is valid.

PR #320, opened 2026-03-31
feat(activation): Reciprocal Rank Fusion (RRF) as scoring alternative

### LINKED ISSUE #316: feat(activation): Reciprocal Rank Fusion (RRF) as scoring alternative
## Problem

MuninnDB combines BM25, HNSW vector, and graph traversal candidates via phase3RRF, then scores them using ACT-R (the default production path). The legacy weighted-sum path exists but is not reachable without explicit DisableACTR. This approach is sensitive to score scale differences between signals — BM25 scores, cosine similarity, and graph weights operate on different scales and distributions. When one signal is noisy or missing (e.g., no embeddings configured), the weighted sum can produce suboptimal rankings.

## Proposal

Add Reciprocal Rank Fusion (RRF) as an alternative scoring method:

```
score(d) = Σ 1/(k + rank_i(d))
```

where k=60 (standard constant from Cormack et al. 2009) and rank_i(d) is the rank of document d in signal i.

RRF is **rank-based and scale-invariant** — it doesn't care about absolute scores, only relative ranking within each signal. This makes it inherently robust to:
- Score scale mismatches between BM25 and cosine similarity
- Missing signals (a document not found by one signal simply has no rank contribution from it)
- Noisy signals (extreme scores don't distort the fusion)

## Design

- Add RRF as an option alongside the existing weighted sum
- Expose as a plasticity parameter: `scoring_fusion: "weighted_sum" | "rrf"`
- When RRF is active, automatically adjust the activation threshold (RRF scores are in the 0-0.05 range, much lower than ACT-R)
- Default to existing weighted sum for backward compatibility
- RRF uses per-signal k constants (FTS=60, HNSW=40, etc.) matching the existing phase3RRF implementation

## Academic basis

- Cormack, G. V., Clarke, C. L. A., & Buettcher, S. (2009). "Reciprocal Rank Fusion Outperforms Condorcet and Individual Rank Learning Methods." *SIGIR '09*.

## Scope

- Add RRF scoring in `internal/engine/activation/`
- Add plasticity parameter for fusion method selection
- Keep existing weighted sum as default

Aperrix:
## Summary

- Adds RRF as an alternative Phase 6 scoring strategy alongside ACT-R, CGDN, and weighted-sum
- Provides a simpler, parameter-free scoring alternative to ACT-R for use cases where access frequency data is sparse (fresh vaults, low-traffic memories). Reuses existing Phase 3 RRF scores as the fusion basis, applies cognitive boosts multiplicatively after fusion
- Rank-based and scale-invariant -- eliminates score magnitude mismatches between BM25, cosine similarity, and graph scores
- New plasticity parameter: `scoring_fusion` (`"rrf"` or `"weighted_sum"`)
- Default behavior unchanged (ACT-R remains default)
- When RRF is active, threshold automatically lowered to 0.001 (RRF scores are in 0-0.05 range)
- Explicit guard when both RRF and CGDN are enabled: RRF takes precedence, CGDN is disabled with a warning log

### Why

Weighted sum scoring is sensitive to score scale differences between retrieval signals. RRF (Cormack et al. 2009) operates on ranks rather than raw scores, making it robust to scale mismatches and missing signals. Particularly useful when one signal is unavailable (e.g., no embeddings configured).

Fixes #316

## Test plan

- [x] RRF produces correct scores for known rankings
- [x] Scale invariance verified
- [x] Multi-signal > single-signal scoring
- [x] k=60 expected values validated
- [x] Missing signal contributes zero
- [x] Default config uses ACT-R (backward compatible)
- [x] Hebbian/transition boost and confidence modulation applied correctly
- [x] Non-negativity invariant
- [x] resolveWeights flag propagation
- [x] RRF returns results with default threshold (regression test for threshold bug)
- [x] RRF+CGDN conflict resolved correctly (RRF wins)
- [x] RED/GREEN verified
- [x] Full test suite passes

### DIFF
--- a/internal/auth/plasticity.go
+++ b/internal/auth/plasticity.go
@@ -57,9 +57,6 @@ type PlasticityConfig struct {
 	// "weighted_sum" = use legacy weighted-sum scoring (DisableACTR implied).
 	// nil/empty = default (ACT-R scoring, unchanged behavior).
 	ScoringFusion *string `json:"scoring_fusion,omitempty"`
-	// RRF_K is the RRF smoothing constant (default 60). Only used when ScoringFusion="rrf".
-	// Higher k values compress rank differences; lower k values amplify top-rank advantage.
-	RRF_K *int `json:"rrf_k,omitempty"`
 }
 
 // ResolvedPlasticity is the fully-merged configuration after applying preset defaults
@@ -106,8 +103,6 @@ type ResolvedPlasticity struct {
 	RecallMode string `json:"recall_mode"`
 	// ScoringFusion selects Phase 6 scoring strategy: "" (default=ACT-R), "rrf", or "weighted_sum".
 	ScoringFusion string `json:"scoring_fusion"`
-	// RRF_K is the RRF smoothing constant (default 60). Only meaningful when ScoringFusion="rrf".
-	RRF_K int `json:"rrf_k"`
 }
 
 type plasticityPreset struct {
@@ -137,7 +132,6 @@ type plasticityPreset struct {
 	EnrichmentEnabled bool
 	RecallMode        string
 	ScoringFusion     string // "" = default (ACT-R), "rrf", "weighted_sum"
-	RRF_K             int    // default 60
 }
 
 var plasticityPresets = map[string]plasticityPreset{
@@ -166,7 +160,7 @@ var plasticityPresets = map[string]plasticityPreset{
 		InlineEnrichment:     "caller_preferred",
 		EnrichmentEnabled:    true,
 		RecallMode:           "balanced",
-		RRF_K:                60,
+
 	},
 	"reference": {
 		HebbianEnabled:       true,
@@ -193,7 +187,7 @@ var plasticityPresets = map[string]plasticityPreset{
 		InlineEnrichment:     "caller_preferred",
 		EnrichmentEnabled:    true,
 		RecallMode:           "balanced",
-		RRF_K:                60,
+
 	},
 	"scratchpad": {
 		HebbianEnabled:       false,
@@ -220,7 +214,7 @@ var plasticityPresets = map[string]plasticityPreset{
 		InlineEnrichment:     "caller_preferred",
 		EnrichmentEnabled:    true,
 		RecallMode:           "balanced",
-		RRF_K:                60,
+
 	},
 	"knowledge-graph": {
 		HebbianEnabled:       true,
@@ -247,7 +241,7 @@ var plasticityPresets = map[string]plasticityPreset{
 		InlineEnrichment:     "caller_preferred",
 		EnrichmentEnabled:    true,
 		RecallMode:           "balanced",
-		RRF_K:                60,
+
 	},
 }
 
@@ -290,7 +284,6 @@ func ResolvePlasticity(cfg *PlasticityConfig) ResolvedPlasticity {
 		EnrichmentEnabled:    p.EnrichmentEnabled,
 		RecallMode:           p.RecallMode,
 		ScoringFusion:        p.ScoringFusion,
-		RRF_K:                p.RRF_K,
 	}
 
 	if cfg == nil {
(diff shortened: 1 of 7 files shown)


5queezer (changes requested):
## Review

### Major: Threshold floor makes RRF return empty results in practice
`Run()` clamps threshold to ≥0.05, `activateCore` defaults to 0.1. But rank #1 in both FTS and HNSW scores `1/61 + 1/41 ≈ 0.04` — below both floors. **Most real callers will get zero results.** The integration test sidesteps this with `Threshold: 0.001`.

Fix: lower the default threshold when `UseRRFFusion` is set, or normalize RRF scores into the ACT-R range.

### Major: No guard when both `UseRRFFusion` and `UseCGDN` are set
RRF branch jumps past CGDN via `goto cgdnDone`. CGDN is silently ignored. Add a warning log or error.

### Minor
- PR description claims `rrf_k` plasticity parameter exists — it doesn't. k values are hardcoded.

scrypster (changes requested):
## Review: Reciprocal Rank Fusion (RRF)

Thanks for this contribution — RRF is a well-established rank fusion technique and an interesting alternative scoring path. However, there is a ship-blocking bug that must be fixed before this can merge.

### 🚨 Ship-Blocking Bug: RRF scores fall below default activation threshold

The maximum possible RRF score with typical result sets is approximately **0.04** (e.g., `1/(60+1) ≈ 0.016` per ranker, summed across rankers). The default activation threshold is **0.1**.

This means in production, with default settings, **all RRF results are filtered out** and users get zero results. This is a silent failure — no error is returned, just empty results.

This must be fixed before merging. Options:
- When `UseRRFFusion` is enabled, automatically use an RRF-appropriate threshold (e.g., lower default or skip threshold filtering entirely)
- Document the required config change prominently and add a validation warning if the threshold is too high for RRF mode
- At minimum, the existing threshold must not silently discard all RRF results

### ⚠️ Required: Guard against `UseRRFFusion` + `UseCGDN` conflict

These two modes are mutually exclusive — both control the final scoring/ranking step. If both are enabled simultaneously, the behavior is undefined. Add a config validation error or clear precedence rule, and document it.

### Minor
- Add a test that verifies RRF results are actually returned with default config (this would have caught the threshold issue)
- Consider a note in the docs/config about when to choose RRF vs ACT-R

---

The threshold bug is a hard blocker — it makes RRF non-functional with default config. Please fix that and the CGDN conflict guard, then request a re-review.

Aperrix:
Thanks for catching the threshold bug — that was a ship-blocker. All items addressed:

**Ship-blocker fix:** When `UseRRFFusion` is active, the activation threshold is automatically lowered to 0.001 in `Run()`. RRF scores live in the 0-0.05 range (max ~0.04 for 2 signals at rank 1), well below the default 0.05 threshold. This is now handled transparently.

**CGDN conflict guard:** Added explicit detection in `phase6Score()` — if both `UseRRFFusion` and `UseCGDN` are enabled, a warning is logged and RRF takes precedence (CGDN disabled).

**Regression tests:**
- `TestRRF_ReturnsResultsWithDefaultThreshold` — passes `Threshold: 0` (triggers 0.05 default) with RRF enabled, verifies non-empty results. This is the test that would have caught the original bug.
- `TestRRF_CGDNConflict_RRFTakesPrecedence` — enables both flags, verifies RRF-scale scores returned.

**PR description:** Updated to remove stale `rrf_k` references and document the threshold adjustment.

All 953 tests pass (activation + auth + mcp).

scrypster (approved):
Both blockers resolved cleanly. The threshold auto-lowering is transparent and practical — `Run()` detects RRF mode and floors to 0.001 before any phase runs, and the regression test would have caught the original bug. CGDN conflict guard is explicit with a warning log so operators can see the misconfiguration. `computeRRFScore` formula is clean, wiring through `activateCore` is correct, and the test suite is thorough. CI all green. Merging.

PR #333, opened 2026-04-01
fix(storage): ClearEmbedFlagsForVault silently skips freshly imported engrams

To3Knee:
## Problem

The Bug 3 fix introduced in #290 (commit `152b242`) is self-defeating. The intent was to write a zero digest record for engrams that have no existing `0x11` record (freshly imported via vault import), so the `RetroactiveProcessor` can see them as pending embedding.

The implementation sets `raw = 0` for missing records but immediately hits a guard that fires for any zero value:

```go
raw, err := ps.getDigestFlagsRaw(id)
if err != nil {
    raw = 0  // intent: fall through and write zero record
}
if raw&embedMask == 0 {
    continue  // BUG: raw=0 makes this always true — zero write never reached
}
```

Result: freshly imported engrams still have no `0x11` record after `ClearEmbedFlagsForVault` runs. The `RetroactiveProcessor` silently skips them, so a vault import followed by reembed never actually embeds the imported engrams.

## Fix

Track whether the record was absent and only skip when an *existing* record already has both flags clear:

```go
noRecord := err != nil
if noRecord {
    raw = 0
}
// Only skip if existing record has both flags clear.
// If no record: fall through to write the zero record.
if !noRecord && raw&embedMask == 0 {
    continue
}
```

## Tests added

- `TestClearEmbedFlagsForVault_NoRecord` — engram with no digest record is counted and gets a zero record written; second call is idempotent (`cleared=0`)
- `TestClearEmbedFlagsForVault_DigestEmbedFailed` — `DigestEmbedFailed` (0x80) is cleared (regression coverage for Bug 4)

All storage tests pass: `go test ./internal/storage/...`

## Related

Fixes the silent skip regression introduced by the Bug 3 guard in #290.

### DIFF
--- a/internal/storage/embed_migration.go
+++ b/internal/storage/embed_migration.go
@@ -2,6 +2,7 @@ package storage
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/cockroachdb/pebble"
@@ -76,7 +77,10 @@ func (ps *PebbleStore) ClearEmbedFlagsForVault(ctx context.Context, ws [8]byte) 
 		copy(id[:], k[9:25])
 
 		raw, err := ps.getDigestFlagsRaw(id)
-		noRecord := err != nil
+		noRecord := errors.Is(err, pebble.ErrNotFound)
+		if err != nil && !noRecord {
+			return cleared, fmt.Errorf("clear embed flags: get digest: %w", err)
+		}
 		if noRecord {
 			raw = 0
 		}
--- a/internal/storage/embed_migration.go
+++ b/internal/storage/embed_migration.go
@@ -76,14 +76,18 @@ func (ps *PebbleStore) ClearEmbedFlagsForVault(ctx context.Context, ws [8]byte) 
 		copy(id[:], k[9:25])
 
 		raw, err := ps.getDigestFlagsRaw(id)
-		if err != nil {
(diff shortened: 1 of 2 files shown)


scrypster (changes requested):
## Review: ClearEmbedFlagsForVault bug fix

Great catch — the bug diagnosis is correct and the fix logic is sound. The self-defeating guard is real: `raw=0` → `raw&embedMask==0` always true → `continue` always fires → zero record never written for freshly imported engrams. The `noRecord` boolean cleanly solves it.

All CI passes. One thing to tighten before merge:

### Error conflation

`getDigestFlagsRaw` returns an error for **any** Pebble failure — not just `ErrNotFound`. The current code (both the original and your fix) treats all errors as "no record":

```go
noRecord := err != nil  // catches I/O errors, corruption, everything
```

A real storage error silently becomes a zero-record write, masking the underlying problem. Since you're already touching exactly this error handling, now is the right time to tighten it:

```go
noRecord := errors.Is(err, pebble.ErrNotFound)
if err != nil && !noRecord {
    return cleared, fmt.Errorf("clear embed flags: get digest: %w", err)
}
```

The `pebble` import is already present. This makes the function fail loudly on real errors rather than silently treating corruption as a missing record.

Everything else looks good — fix is correct, tests cover the regression and the `DigestEmbedFailed` (0x80) case, idempotency is verified. One small change and this is ready to merge.

PR #334, opened 2026-04-01
fix(enrich): expand entity type allowlist and pass through unknown types

To3Knee:
## Problem

 in  has two related bugs.

### Bug 1 — Allowlist is out of sync with the UI colour map

The function only recognises **8 types**:

But  in  has **14 named types**:

The six types present in the UI but absent from the allowlist — , , , , ,  — are silently coerced to . This causes:

- **Wrong graph colours**: e.g. a  entity (should be indigo ) displays as  (cyan )
- **Corrupted semantic metadata**: stored type no longer reflects what the LLM extracted

### Bug 2 — Unknown types silently coerced to "service"

Any type not in the allowlist is mapped to  with no warning. An LLM returning , , , , or any other valid semantic type has its output silently overwritten with a meaningless default.

## Fix

1. Expand the allowlist to all 14 types recognised by  — keeping the two sources in sync
2. Return the normalised (lowercase, trimmed) string for unrecognised types instead of coercing to  — this preserves LLM semantic intent and is forward-compatible with future type additions

## Tests

- Updated  to cover previously-missing UI types and correct the  case
- Added  as a regression guard

All enrich tests pass: 

## Note

The UI colour map () and the allowlist in  should be kept in sync. Consider a shared constant or comment cross-reference to prevent future drift.

### DIFF
--- a/internal/plugin/enrich/parse.go
+++ b/internal/plugin/enrich/parse.go
@@ -267,10 +267,18 @@ func validateAndDedupeEntities(entities []plugin.ExtractedEntity) []plugin.Extra
 	return result
 }
 
-// normalizeEntityType validates and normalizes entity type strings.
+// normalizeEntityType normalizes entity type strings to lowercase and
+// validates against the known types recognised by the UI colour map.
+// Known types are returned as-is after normalisation. Unknown types are
+// returned as their normalised string rather than being silently coerced
+// to "service", which would corrupt semantic information and cause the
+// graph UI to display incorrect colours for any type not in the original
+// eight-item allowlist (e.g. "technology", "location", "concept", "event").
 func normalizeEntityType(t string) string {
 	t = strings.ToLower(strings.TrimSpace(t))
 
+	// Allowlist mirrors the entity types recognised by the UI colour map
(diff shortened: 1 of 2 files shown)


scrypster (approved):
Both bugs confirmed real and the fixes are correct. The allowlist drift was causing silent type corruption and wrong graph colors for 6 valid types — good catch. Changing the fallback to pass-through is the right call; preserving LLM semantic intent beats overwriting with a meaningless default.

Tests are solid — covers the missing types, pass-through behavior, and the `t.Fatalf` → `t.Errorf` fix is a nice bonus so all cases run on failure.

The cross-reference comment pointing back to `getEntityTypeColor` in `app.js` is exactly the right guard against future drift. Approved.

scrypster:
Merged — thank you again @To3Knee. Four PRs in, and each one has been clean, well-reasoned, and well-tested. This one in particular saved users from silent data corruption that would have been frustrating to track down. Really appreciate the attention to detail — the cross-reference comment is exactly the kind of thing that prevents the same bug from coming back. Hope to keep seeing contributions from you.

Issue #327, opened 2026-03-31
Various UI Issues


Invincibear:
Hi there,

Been running this for a bit now, great potential, still working out how to get the most out of it.

I'm opening this issue as a holder of various little UI issues where things don't quite render properly, overlapping items, incorrect behaviors, etc. I'll submit a series of small PRs to fix them referencing this issue if that's okay

scrypster:
Hey Corey — thanks for raising this and for jumping in with fixes. We've been actively reviewing and merging your PRs:

- #328 (search filter button group) — ✅ merged
- #329 (consistent content tabs) — ✅ merged
- #330 (nav bar cut-off) — pipes passed, just needs a quick rebase to resolve a conflict with develop, then it's going in too

Really appreciate the quality of the work. Clean, focused, well-documented changes.

Once #330 is merged, would you mind closing this issue? We'll leave it to you since you likely have the best sense of whether there are more UI items you're planning to address under this ticket or if it's wrapped up. Either way, thanks again — this kind of polish work makes a real difference.

Invincibear:
@scrypster I have a few more PRs coming in but will close it when done.

I just noticed now that the Observability page just goes to the Dashboard page. I assume you have goals of a detailed Observability page compared to the overview but haven't gotten to it yet?
Maybe we should remove one of those redundant links?
~~Was also thinking of making the MuninnDB icon a link to the official website (not the text, just the icon), thoughts?~~ https://github.com/scrypster/muninndb/pull/344

PR #335, opened 2026-04-01
fix(ui): copyToClipboard always fails on HTTP installs with "Copy failed" toast

To3Knee:
## Problem

 in  uses  directly with no fallback:

The [Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API) requires a **secure context** — HTTPS or . Any installation accessed over plain HTTP on a LAN IP (e.g. , which is the common self-hosted setup) has . The  call throws a , which the catch block surfaces as a **"Copy failed"** error toast on every single click.

### Affected surfaces
- API key copy button ( display after creating a key)
- Connect URL / endpoint copy buttons
- Any other caller of the shared  helper

### Already fixed elsewhere — inconsistently
 (line ~911) already has the correct pattern —  never received the same treatment.

## Fix

Mirror the pattern from :

1. Check  availability before calling 
2. Fall back to a  textarea approach when the Clipboard API is unavailable or throws
3. Both paths set  and show the correct success/error notification

The  fallback works on HTTP, in older browsers, and in sandboxed iframes.

## Testing

Reproduced on a self-hosted instance at  — every Copy click showed the error toast before the fix, and worked correctly after. No automated test added (clipboard interaction requires a browser context).

Ref #327

### DIFF
--- a/web/static/js/app.js
+++ b/web/static/js/app.js
@@ -2080,13 +2080,41 @@ document.addEventListener('alpine:init', () => {
     },
 
     async copyToClipboard(text) {
+      // navigator.clipboard requires a secure context (HTTPS or localhost).
+      // Installations accessed over plain HTTP (e.g. a LAN IP like
+      // 192.168.x.x:8476) have navigator.clipboard === undefined, causing
+      // the previous implementation to always throw and show an error toast.
+      // Fall back to the legacy execCommand path for non-secure contexts.
+      if (navigator.clipboard && navigator.clipboard.writeText) {
+        try {
+          await navigator.clipboard.writeText(text);
+          this.connectCopied = true;
+          setTimeout(() => { this.connectCopied = false; }, 2000);
+          this.addNotification('success', 'Copied to clipboard');
+          return;
+        } catch (_) {
+          // Fall through to execCommand fallback.
+        }
+      }
+      // execCommand fallback — works on HTTP and older browsers.
       try {
-        await navigator.clipboard.writeText(text);
-        this.connectCopied = true;
-        setTimeout(() => { this.connectCopied = false; }, 2000);
-        this.addNotification('success', 'Copied to clipboard');
+        const ta = document.createElement('textarea');
+        ta.value = text;
+        ta.style.position = 'fixed';
+        ta.style.opacity = '0';
+        document.body.appendChild(ta);
+        ta.select();
+        const ok = document.execCommand('copy');
+        document.body.removeChild(ta);
+        if (ok) {
+          this.connectCopied = true;
+          setTimeout(() => { this.connectCopied = false; }, 2000);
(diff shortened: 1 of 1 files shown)


scrypster (changes requested):
## Review: copyToClipboard HTTP fallback

Bug is confirmed real and the fix logic is correct — `navigator.clipboard` is `undefined` on plain HTTP installs, the guard + execCommand fallback is the right pattern, and the existing `_copyFallback()` at line ~929 already does the same thing so this is consistent with the codebase. All CI passes.

Three things to address before merge:

### 1. `copyToken()` has the same bug and isn't fixed (medium)

Line 2453:
```js
navigator.clipboard.writeText(this.clusterToken.token).catch(() => {});
```

No guard, no fallback. On HTTP this throws a `TypeError` synchronously — `.catch()` doesn't help because the throw happens before the promise is created. Cluster token copy is broken for all HTTP users. Please apply the same fix here.

### 2. `Closes #327` closes the wrong scope (low)

Issue #327 is an umbrella UI issues ticket. Merging this will close the entire issue, which may be premature if other items under it are still open. Please change to `Ref #327` or `Partially addresses #327`.

### 3. DOM leak if `execCommand` throws (low)

```js
document.body.appendChild(ta);
ta.select();
const ok = document.execCommand('copy');
document.body.removeChild(ta);  // skipped if above throws
```

If `execCommand` throws (rare, but possible under some browser security policies), `removeChild` is never reached. Wrap in try/finally:

```js
document.body.appendChild(ta);
try {
  ta.select();
  const ok = document.execCommand('copy');
  if (ok) {
    this.connectCopied = true;
    setTimeout(() => { this.connectCopied = false; }, 2000);
    this.addNotification('success', 'Copied to clipboard');
  } else {
    this.addNotification('error', 'Copy failed — please select and copy manually');
  }
} finally {
  document.body.removeChild(ta);
}
```

Note: the existing `_copyFallback()` at line ~929 has the same leak — feel free to fix both while you're here.

---

Fix items 1 and 2, and this is ready to merge. Item 3 is a nice-to-have.

scrypster:
Merged — great turnaround on the feedback @To3Knee. You fixed all three: `copyToken()` guarded with proper fallback, `_copyFallback()` cleaned up with try/finally, and the issue reference corrected. Clean work.

To3Knee:
Thanks for the thorough review — all three items addressed in ad064ef:

1. **`copyToken()` same bug** — added `navigator.clipboard` availability guard and full `execCommand` fallback matching the `copyToClipboard()` pattern.

2. **`Closes #327` scope** — updated to `Ref #327` in the PR description.

3. **DOM leak (try/finally)** — fixed in both `copyToClipboard()` and `_copyFallback()`: textarea is now created before the `try` block and removed in `finally`, so it's always cleaned up even if `execCommand` throws.

PR #337, opened 2026-04-01
fix(engine): resolve ACT-R score saturation in fresh/small vaults

### LINKED ISSUE #331: ACT-R score saturation: fresh engrams clamp to raw=1.0, destroying ranking
## Summary

`computeACTR()` in `internal/engine/activation/engine.go:1452-1456` clamps raw scores to [0, 1]:

```go
raw := contentMatch * contextualPrior / actrDenominator
if raw > 1.0 {
    raw = 1.0  // ← saturation point
}
```

For freshly written engrams in small-to-medium vaults, `contextualPrior` (softplus of base-level activation) is high because `B(M) = ln(n+1) - d × ln(ageDays / (n+1))` produces large values when `ageDays` is small. The result: every engram with any content relevance saturates to `raw = 1.0`, producing `Final = 1.0 × confidence` for all of them. Ranking becomes arbitrary tie-breaking.

## Evidence

Benchmark with 13 test engrams in an isolated vault:
- All engrams score **exactly 0.9000** (confidence=0.9)
- Semantic similarity varies meaningfully underneath: Red=0.8976, Blue=0.7168, Padraig=0.4495
- But final scores are identical — ranking is noise
- Controlled variance test (3 runs, vault cleared between each, temperature=0): **zero score variance**, content ordering changes randomly

This was discovered during benchmark work for #311. Full report: https://gist.github.com/5queezer/3fde45014b5550a5520189068e4b0052

## Impact

- Temporal ordering fails (latest colour update not prioritized over older ones)
- Irrelevant engrams tie with relevant ones in top-N results
- All existing ACT-R benchmarks measured tie-breaking artifacts, not retrieval quality
- Dream engine consolidation impact cannot be measured until this is fixed

## Suggested fix

**Per-query normalization** (least invasive): divide all raw scores by the query's max raw score before applying confidence. Preserves [0,1] contract, preserves relative ranking, no behavioral change for mature vaults where scores already spread below 1.0.

Other options:
- Use raw scores for ranking, apply confidence only for threshold filtering
- Increase `actrDenominator` to account for fresh-engram activation levels
- Remove the clamp entirely

## Context

- The comment at line 1452 says "Clamp to [0, 1] for contract" — this is intentional normalization, not a bug
- It works correctly for mature vaults with age diversity where scores naturally spread below 1.0
- It breaks for fresh engrams (ageDays < 1) where softplus(B(M)) saturates far above 1.0
- Blocked: benchmark validation of PRs #317, #318, #320, and dream engine phases 2-6

## Also found during this work

Separate bug fixed in my fork: `activate()` only filtered `StateSoftDeleted`, not `StateArchived`. Dream engine archives engrams but they remained visible in recall. Fixed in 10 retrieval paths — will submit as a separate PR.

To3Knee:
## Problem

In vaults with recently written engrams, `computeACTR()` produces raw scores well above 1.0. The previous hard clamp collapsed every saturated score to 1.0 — all engrams scored identically, making ranking arbitrary tie-breaking.

Reproduction (from issue #331 benchmark):
- 13 test engrams, all freshly written
- All final scores: **exactly 0.9000** (confidence=0.9)
- Semantic similarity varied meaningfully underneath: Red=0.8976, Blue=0.7168, Padraig=0.4495
- Controlled variance test (3 runs, vault cleared, temperature=0): **zero score variance**, content ordering changes randomly

**Root cause:** `B(M) = ln(n+1) - d × ln(ageDays/(n+1))` produces large values when `ageDays` is small (fresh engrams). `softplus(B(M))` far exceeds `actrDenominator ≈ 1.693`, so `raw = contentMatch × softplus(B(M)) / actrDenominator >> 1.0`. The clamp collapsed all of these to the same ceiling.

## Fix

Two-pass ACT-R scoring with per-query normalization — the approach suggested in the issue.

**Pass 1:** Collect all raw scores. `computeACTR()` no longer applies an upper clamp, so scores above 1.0 are preserved for the normalization step.

**Pass 2:** If any score exceeded 1.0, divide the whole result set by `maxRaw`. This rescales scores so the highest-activation candidate anchors at 1.0, restoring relative ranking. For mature vaults where `maxRaw ≤ 1.0` the scale factor is 1.0 — the code path is identical to before.

```go
scale := 1.0
if maxRaw > 1.0 {
    scale = 1.0 / maxRaw
}
for _, cc := range actrCands {
    raw := math.Min(cc.components.Raw*scale, 1.0)
    ...
}
```

The `[0,1]` output contract is preserved via `math.Min` after normalization. No formula changes, no new parameters, no behavioural change for mature vaults.

## Tests

- `TestComputeACTR_ScoreClamping` — updated to verify `computeACTR` now returns the unclamped value (> 1.0); clamping is the caller's responsibility.
- `TestACTR_PerQueryNormalization_RankingPreserved` — new regression test: two fresh engrams with different content match both saturate above 1.0; after normalization the higher-relevance one must outscore the lower-relevance one.

All existing ACT-R tests pass.

Fixes #331

### DIFF
--- a/internal/engine/activation/actr_test.go
+++ b/internal/engine/activation/actr_test.go
@@ -26,13 +26,12 @@ func assertNear(t *testing.T, name string, got, want, tol float64) {
 }
 
 // expectedACTRRaw computes the expected Raw score from first principles.
+// Note: no upper clamp — computeACTR now returns the unclamped value so that
+// the ACT-R scoring path can apply per-query normalization (issue #331 fix).
 func expectedACTRRaw(contentMatch, baseLevel, hebScale, hebbianBoost float64) float64 {
 	totalActivation := baseLevel + hebScale*hebbianBoost
 	contextualPrior := softplus(totalActivation)
 	raw := contentMatch * contextualPrior / actrDenominator
-	if raw > 1.0 {
-		return 1.0
-	}
 	if raw < 0.0 {
 		return 0.0
 	}
@@ -184,6 +183,9 @@ func TestComputeACTR_ConfidenceMultiplication(t *testing.T) {
 }
 
 func TestComputeACTR_ScoreClamping(t *testing.T) {
+	// computeACTR no longer applies an upper clamp — clamping to [0,1] is the
+	// caller's responsibility after per-query normalization (issue #331 fix).
+	// Verify that the raw value above 1.0 is returned unchanged.
 	now := time.Now()
 	eng := &storage.Engram{
 		Confidence:  1.0,
@@ -193,13 +195,10 @@ func TestComputeACTR_ScoreClamping(t *testing.T) {
 	}
 	w := actrDefaultWeights()
 
-	// Perfect match + very high FTS + Hebbian + many accesses → exceeds 1.0 before clamp.
+	// Perfect match + very high FTS + Hebbian + many accesses → exceeds 1.0.
 	sc := computeACTR(1.0, 10.0, 1.0, 0.0, eng, 0, now, w)
 
-	if sc.Raw > 1.0 {
-		t.Errorf("Raw=%v, must be clamped to <= 1.0", sc.Raw)
-	}
-	// Verify it actually hit the ceiling (pre-clamp value exceeds 1.0).
+	// Compute expected unclamped value.
 	contentMatch := 0.35*1.0 + 0.25*math.Tanh(10.0)
 	n := 51.0
 	ageDays := 1.0 / (24.0 * 60.0) // floor: 1 minute
@@ -207,9 +206,13 @@ func TestComputeACTR_ScoreClamping(t *testing.T) {
 	totalActivation := baseLevel + 4.0*1.0
 	unclamped := contentMatch * softplus(totalActivation) / actrDenominator
 	if unclamped <= 1.0 {
-		t.Fatalf("Pre-clamp value=%.4f, test requires it to exceed 1.0", unclamped)
+		t.Fatalf("unclamped=%.4f, test requires > 1.0 to be meaningful", unclamped)
 	}
-	assertNear(t, "Raw", sc.Raw, 1.0, 1e-9)
+	// computeACTR must now return the unclamped value — the caller normalises.
+	if sc.Raw <= 1.0 {
+		t.Errorf("Raw=%.4f, expected > 1.0 (no longer clamped in computeACTR)", sc.Raw)
+	}
+	assertNear(t, "Raw (unclamped)", sc.Raw, unclamped, 1e-9)
 }
 
 func TestComputeACTR_ZeroLastAccess(t *testing.T) {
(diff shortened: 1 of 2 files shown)


To3Knee:
Thank you for the kind words on #335 — really glad the work has been landing well. Looking forward to your thoughts on this one and #340 whenever you get a chance.

5queezer (commented):
The two-pass normalization correctly fixes the immediate ranking collapse — fresh vaults will produce differentiated scores again, and the mature-vault path is unchanged. Tests cover both the unclamped return and the normalization round-trip. Clean implementation.

**However, this treats the symptom rather than the root cause.**

The real problem is in `contextualPrior`. With the 1-minute `ageDays` floor:

```
B(M) = ln(n+1) - d × ln(ageDays/(n+1))
```

A freshly written engram with `AccessCount=5` gives `B(M) ≈ 6.3`, `softplus(6.3) ≈ 6.3`, so `contextualPrior / actrDenominator ≈ 3.7×` — guaranteed saturation regardless of content match. The formula diverges as `ageDays → 0`; the floor just picks how far past 1.0 you blow.

The per-query normalization band-aid has a subtle cost: **scores become query-relative, not absolute.** A `threshold=0.3` means something different for a fresh-only query (where the max raw might be 4.0) vs. a mature-vault query (where max raw ≤ 1.0). Cross-query score comparisons (e.g. for downstream ranking or federation) also break.

**Suggested direction:** dampen `contextualPrior` for young engrams at the source — either a smooth ramp on the `ageDays` floor, or cap `B(M)` at the denominator's assumed median activation. That keeps scores absolute, preserves threshold semantics across queries, and still allows the base-level formula to work correctly for engrams with meaningful age.

The normalization could remain as a safety net, but shouldn't be the primary fix.

To3Knee:
@5queezer — you're right on both counts. Revised in 3fa25a1.

**What changed**

Two targeted changes in `computeACTR`, exactly as you suggested:

**1. Smooth ramp on the age floor** — replaced the hard 1-minute floor used in the B(M) computation with a 1-day additive offset (`bmAgeOffset = 1.0`). For sub-day engrams `ageForBM ≈ 1 day` (no divergence). For mature engrams `ageForBM ≈ ageDays` (no behavioural change). The original 1-minute floor is retained for `Recency` and `DecayFactor` reporting only.

**2. Cap B(M) at the derived saturation threshold** — `bLevelCap = ln(exp(actrDenominator) - 1) ≈ 1.489`. This is the unique value where `softplus(B(M)) = actrDenominator`, i.e. where base-level alone would push `raw = contentMatch`. Above this, the formula violates the semantic contract that contentMatch gates the score. The cap is derived from the denominator constant, not arbitrary. Hebbian boosts may still legitimately push `totalActivation` above the cap — that's intentional.

**Two-pass normalization retained as safety net**, per your recommendation. Post-fix it only fires when strong Hebbian boosts push `totalActivation` well past `bLevelCap`. It never fires for the zero-Hebbian fresh-vault case that caused #331.

**Score absoluteness restored**: `threshold=0.3` now means the same thing across fresh-vault and mature-vault queries.

**Tests**

Replaced `TestACTR_PerQueryNormalization_RankingPreserved` (which required `raw > 1.0` as a precondition — no longer true for zero-Hebbian fresh engrams) with three root-cause tests:
- `TestComputeACTR_FreshVault_NoDivergence` — asserts `raw ≤ contentMatch` for n1…n100, zero Hebbian
- `TestComputeACTR_FreshVault_Differentiated` — asserts two fresh engrams with different content-match scores rank correctly without any normalization
- `TestComputeACTR_TwoPassSafetyNet_HebbianCanStillSaturate` — confirms the safety net is still reachable via strong Hebbian boost (so it's not dead code)

To3Knee:
Revised again in ab11bdd — dropped the 1-day age offset.

The offset was causing a secondary problem: for n=1 (freshly written, never accessed), `B(M) = ln(1) - d×ln(1) = 0` regardless of age or decay. With FTS-only content match (~0.23 in environments without an embedder), raw dropped to ~0.09 — below the default `threshold=0.1`, breaking retrieval of newly written engrams entirely.

Cap-only is simpler and sufficient:
- Original 1-minute age floor is restored
- `bLevelCap ≈ 1.489` is applied after the standard B(M) computation
- n=1 fresh: B(M)=3.64 → capped → `raw = contentMatch` (retrievable at any reasonable threshold)
- n=5 fresh: B(M)=6.3 → capped → same ceiling, differentiated by content match
- Old engrams with natural B(M) < 1.489 are untouched

The smooth ramp was appealing but `B(M) = ln(n) - d×ln(ageForBM/n)` evaluates to exactly 0 for n=1 when `ageForBM = n = 1` — a formula identity that kills recency for brand-new engrams. The cap avoids that entirely.

5queezer (commented):
We ran a controlled benchmark (13 engrams, 5 queries, 3 runs, vault cleared between each) to validate this fix. Results confirm the saturation issue is resolved.

| Metric | Before (#331 baseline) | After (this PR) |
|--------|------------------------|-----------------|
| Score range | 0.9000 only | 0.18 – 0.90 |
| Ranking stability | Arbitrary tie-breaking | 12–13/15 positions stable (max_diff ≤ 0.0003) |
| Correct top-1 | Indeterminate | 5/5 |

The two-pronged approach is the right call. The `bLevelCap` addresses the root cause scrypster flagged (dampening `contextualPrior` at the source rather than just normalizing after), and the per-query normalization remains as a safety net for Hebbian-boosted engrams that legitimately exceed 1.0. This directly addresses scrypster's review feedback — nicely done.

### One edge case worth noting (non-blocking)

When dream consolidation archives engrams from a small vault, the candidate pool shrinks enough that an unrelated engram can become the new normalization anchor — flipping top-1 results. We observed "Padraig is what I am called." anchoring at 0.9000 for 4/5 unrelated queries after dedup removed 3 colour variants from a 13-engram vault.

This won't affect mature vaults where removing 3 candidates from hundreds doesn't shift the landscape. But it's worth a comment near the normalization path noting the small-vault sensitivity, and possibly a minimum-candidate-count guard in a follow-up.

Full benchmark report: https://gist.github.com/5queezer/3fde45014b5550a5520189068e4b0052

This is ready to merge.

To3Knee:
Thanks for running the benchmark — those numbers are exactly what we'd hope to see. Good to have controlled validation rather than just unit tests.

The small-vault normalization anchor issue is a good catch. It's a natural consequence of the two-pass path firing when the candidate pool is very small — the safety net works correctly, it just has less to work with. Worth tracking as a follow-up issue so it doesn't get lost, but nothing that should hold up the merge.

Happy to see this go in whenever you're ready.

PR #349, opened 2026-04-02
fix(activation): Dormant flag should be false in ACT-R mode

### LINKED ISSUE #348: bug(activation): Dormant flag uses Ebbinghaus relevance instead of ACT-R score
## Problem

The `Dormant` flag on activation results is computed from `eng.Relevance` (the Ebbinghaus decay worker output):

```go
// internal/engine/activation/engine.go:1347
Dormant: eng.Relevance <= minFloor*1.1,
```

In ACT-R mode (the default production path), `Relevance` is written by the background Ebbinghaus worker using pure exponential decay. But ACT-R scores engrams using a power-law formula that accounts for access frequency:

```
B(M) = ln(n+1) - d × ln(max(ageDays, ageFloor) / (n+1))
```

These two systems can contradict each other:

- **Ebbinghaus** (background): old engram → `Relevance` drops below floor → marked dormant
- **ACT-R** (at recall): same engram accessed 50 times → high base-level activation → should surface

An engram that is old but frequently accessed gets marked `Dormant=true` by Ebbinghaus even though ACT-R would score it highly. The Dormant flag is returned to MCP clients in `ActivationItem.Dormant` and is used by Phase 4.75 (archive restore).

## Reproduction

1. Create an engram
2. Access it frequently over several weeks (high ACT-R `n`)
3. Wait for the Ebbinghaus worker to decay `Relevance` below `minFloor * 1.1`
4. Recall — the engram surfaces (ACT-R scores it well) but with `Dormant=true`

## Expected behavior

`Dormant` should reflect the active scoring model. In ACT-R mode, it should use ACT-R base-level activation, not Ebbinghaus relevance.

## Possible fixes

1. **Compute dormancy from ACT-R inline**: `Dormant: computeACTRBaseLevel(eng) <= threshold`
2. **Skip dormancy in ACT-R mode**: `Dormant: !w.UseACTR && eng.Relevance <= minFloor*1.1`
3. **Have the Ebbinghaus worker write ACT-R base-level to Relevance field** when ACT-R mode is active

Option 2 is the simplest. Option 1 is the most correct.

## Context

This was discovered while investigating issue #314 (hybrid decay). The maintainer correctly noted that ACT-R already handles power-law decay for scoring. However, the `Relevance` field written by the Ebbinghaus worker is not purely diagnostic — it feeds the `Dormant` flag which is exposed to clients and used by Phase 4.75. The inconsistency between the two decay models creates a concrete bug.

Aperrix:
## Summary

One-line fix: `Dormant: !w.UseACTR && eng.Relevance <= minFloor*1.1`

In ACT-R mode (the default), the `Dormant` flag was derived from `eng.Relevance` (Ebbinghaus decay output). This contradicts ACT-R scoring for frequently-accessed old engrams — ACT-R scores them highly but the Ebbinghaus worker has already decayed `Relevance` below the floor, marking them dormant.

In ACT-R mode, dormancy is now always `false` — it's implicit via scoring (low ACT-R base-level = below threshold = not returned). In legacy weighted-sum mode, `Dormant` still reflects `Relevance` as before.

Fixes #348

## Test plan

- [x] ACT-R mode: `Dormant=false` regardless of `Relevance` (2 tests)
- [x] Legacy mode: `Dormant=true` when `Relevance` below floor (2 tests)
- [x] RED/GREEN verified
- [x] Full activation suite passes (235 tests)

### DIFF
--- a/internal/engine/activation/activation_test.go
+++ b/internal/engine/activation/activation_test.go
@@ -323,27 +323,27 @@ func TestActivationLogRingBuffer(t *testing.T) {
 }
 
 // ---------------------------------------------------------------------------
-// Test 4: Engram with very low relevance (0.01) comes back Dormant == true
+// Test 4: Dormant flag respects scoring mode
 // ---------------------------------------------------------------------------
 
-func TestDormantFlagSetWhenRelevanceLow(t *testing.T) {
+func TestDormantFlag_ACTRMode_AlwaysFalse(t *testing.T) {
 	store := newStubStore()
-	dormant := &storage.Engram{
+	eng := &storage.Engram{
 		Concept:    "dormant engram",
 		Content:    "barely alive",
 		Confidence: 1.0,
 		Stability:  30.0,
-		// Relevance 0.01 is well below minFloor*1.1 = 0.05*1.1 = 0.055
-		Relevance: 0.01,
+		Relevance:  0.01, // well below minFloor*1.1 = 0.055
 	}
-	store.writeEngram(dormant)
+	store.writeEngram(eng)
 
-	fts := &stubFTS{results: []activation.ScoredID{{ID: dormant.ID, Score: 0.8}}}
-	hnsw := &stubHNSW{results: []activation.ScoredID{{ID: dormant.ID, Score: 0.8}}}
+	fts := &stubFTS{results: []activation.ScoredID{{ID: eng.ID, Score: 0.8}}}
+	hnsw := &stubHNSW{results: []activation.ScoredID{{ID: eng.ID, Score: 0.8}}}
 
-	eng := newTestEngine(store, fts, hnsw)
+	e := newTestEngine(store, fts, hnsw)
 
-	result, err := eng.Run(context.Background(), &activation.ActivateRequest{
+	// Default mode is ACT-R — Dormant should be false regardless of Relevance.
+	result, err := e.Run(context.Background(), &activation.ActivateRequest{
 		Context:    []string{"barely alive"},
 		Threshold:  0.0,
 		MaxResults: 5,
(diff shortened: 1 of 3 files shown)


scrypster (changes requested):
## Review: Dormant flag in ACT-R mode

The one-line fix is correct, the logic is sound, tests are well-structured and cover both modes. All 7 CI checks pass.

One thing to fix before merge:

### `buildWhy` still emits "dormant" in ACT-R mode

The `buildWhy` function (around line 1647) adds a human-readable annotation independently of the structured `Dormant` field:

```go
if eng.Relevance <= minFloor*1.1 {
    parts = append(parts, "dormant (low decay relevance)")
}
```

This is not gated on `!useACTR`, so in ACT-R mode the `Why` string will say `"dormant (low decay relevance)"` while `Dormant=false`. That's a direct contradiction in the same response — the structured field says not dormant, the explanation says dormant.

Simplest fix — pass the already-computed `Dormant` bool into the caller site, or gate the annotation:

```go
if !useACTR && eng.Relevance <= minFloor*1.1 {
    parts = append(parts, "dormant (low decay relevance)")
}
```

One line, same pattern as the main fix. Everything else looks good — merge-ready once that's in.

Aperrix:
Fixed — `buildWhy` now also gates the dormant annotation on `!useACTR`, same pattern. Added the `useACTR` parameter to the function signature and updated all 6 test callsites.

Also found during the audit: `trigger/system.go:310` (`TriggerScore()`) uses `meta.Relevance` as the `decayFactor` component (20% weight) of the trigger scoring formula. Same Ebbinghaus-vs-ACT-R inconsistency, but different scope — noted in the commit message for future tracking.

scrypster (approved):
The `buildWhy` fix is exactly right — `useACTR bool` added to the signature and the dormant annotation gated on `!useACTR`. Structured field and explanation string are now consistent. Tests split correctly into ACT-R (Dormant=false) and legacy (Dormant=true) cases. Good catch on the `trigger/system.go` Relevance usage — worth a follow-up issue. Merging.

Issue #308, opened 2026-03-29
[feat] agent-manged enrichment


rsubr:
Currently, the enrichment plugin uses server-managed LLM calls that require an API key. This incurs additional API costs, and I would like to use my Claude/ChatGPT/etc. subscription instead.

As LLM providers only permit the use of their own agents when using subscription plans, could we consider add a new feature to the enrichment plugin so enrichment is agent-managed? This way, the enrichment feature would be available as an MCP tool that would be called by Claude/ChatGPT itself, using the agent's LLM for enrichment rather than triggering a separate backend LLM API call.

We could initiate enrichment MCP tool calls periodically using Claude's `/schedule` command. This method could also be used for the new "dream" feature introduced in the latest release.

Claude Code provided me with this [TODO.md](https://github.com/user-attachments/files/26333314/TODO.md) implementation plan.

@scrypster, can you please let me know your thoughts?

Thanks.

scrypster:
Hey Raja — thanks for the detailed writeup and the effort that went into the TODO.md. Really appreciate the thoroughness.

We wanted to check in with you before this goes further, because we think what you're looking for may already be possible with MuninnDB today.

MuninnDB has an enrichment plugin system that supports multiple LLM providers out of the box — including **Anthropic (Claude), OpenAI, Google, and Ollama**. You configure it via `MUNINN_ENRICH_URL` and `MUNINN_ENRICH_API_KEY` at startup, and the server handles enrichment automatically in the background after each write. No changes to your MCP client needed.

For example, to use Claude Haiku for enrichment:

```
MUNINN_ENRICH_URL=anthropic://claude-haiku-3
MUNINN_ENRICH_API_KEY=sk-ant-...
```

We do want to address the cost concern directly — we understand the motivation if you're trying to avoid paying for a separate API key on top of a Claude Pro subscription. Unfortunately, **Anthropic doesn't allow programmatic access via a Claude Pro subscription** — the consumer subscription (claude.ai) and the API are entirely separate products with separate billing. This is Anthropic's policy, not a MuninnDB limitation. The same is true for OpenAI's ChatGPT Plus. There's no OAuth or credential bridge between the two.

The most cost-effective path is probably one of:
- **`enrich-anthropic`** with `claude-haiku-3` — very cheap at ~$0.25/MTok for the kind of short extraction tasks enrichment does
- **`enrich-ollama`** with a local model — completely free if you have the hardware

[...]

PR #358, opened 2026-04-02
[feat] agent-manged enrichment

rsubr:
Sorry... i'm not too good at this PR workflow, please excuse any mistakes.

## Summary

Expose two new MCP tools that let an external agent drive enrichment explicitly, without invoking the built-in enrichment plugin:

- muninn_get_enrichment_candidates: returns active engrams missing one or more enrichment stages (entities, relationships, classification, summary) so a caller can decide which to enrich.
- muninn_apply_enrichment: accepts agent-generated enrichment output (summary, memory type/label, entities, relationships) and persists it, guarded by an optimistic-concurrency check on expected_updated_at to prevent lost-update races (ErrEnrichmentConflict).

Please see https://github.com/scrypster/muninndb/issues/308.

I tested this with ChatGPT and it works. ChatGPT caches `tools/list` and you have to delete the MCP first and then add it again for it to detect the new MCP tools.

## Prompt used to trigger enrichment using the client agent's LLM

```
You have access to a MuninnDB memory store via MCP tools. Your job is to enrich memories that are missing enrichment stages.
  Follow this loop:
  1. Call `muninn_get_enrichment_candidates` with the stages you can produce (e.g. ["summary", "entities", "relationships"]). Use a reasonable limit (50).
  2. For each candidate returned in `items`:
     a. Read its `content` and `missing_stages`.
     b. Generate only the stages listed in `missing_stages`:
        - **summary**: a concise 1-3 sentence summary of the memory's content.
        - **entities**: named entities (people, systems, tools, organizations, concepts) with their type.
        - **relationships**: entity-to-entity relationships with `from_entity`, `to_entity`, and `rel_type`.
        - **classification**: a `memory_type` (e.g. "decision", "fact", "event") and optional `type_label`.
     c. Call `muninn_apply_enrichment` with:
        - `id`: the candidate's `id`
        - `expected_updated_at`: the candidate's `updated_at` (verbatim — this is the optimistic lock)
        - the enrichment output fields you generated
  3. If `muninn_apply_enrichment` returns a conflict error, skip that candidate and continue to the next.
  4. Repeat until all candidates are processed.
  Key constraint the agent must respect: always pass updated_at from the candidate response verbatim as expected_updated_at. If the engram was modified between
  the fetch and the apply, the server rejects the write — the agent should skip and move on, not retry with a stale timestamp.
```


scrypster (commented):
## Review: feat(mcp): agent-managed enrichment via MCP

First — thank you for this. The agent-managed enrichment pattern is genuinely well thought out, and the pivot from the original framing to \"let the agent already in the conversation drive enrichment\" is the right architectural read. The implementation follows the codebase conventions closely (you clearly studied `ReplayEnrichment` and `RetryEnrich`), the optimistic concurrency via `expected_updated_at` with the `-32009` conflict code is exactly the right design for agent retry flows, and the test coverage across engine, handler, and smoke layers is solid for a contribution of this scope.

Before I approve I have three questions — all minor, just want to understand the intent:

---

**1. `GetEnrichmentCandidates` scan pattern**

The method calls `ListByState` (up to `limit` active engrams), then does a per-engram `GetDigestFlags` call to filter out already-enriched ones. For small-to-medium vaults this is fine and consistent with how `ReplayEnrichment` works. But in a large vault where most memories are already enriched, the agent may have to call this tool many times before finding unenriched candidates — there's no pagination cursor or digest-flags-aware index.

Was this a deliberate tradeoff (keep it simple, follow existing patterns) with the intent to revisit at scale? Or do you have something in mind for the pagination case? Happy either way — just want to know if there's a follow-up issue planned or if this is intended as the long-term approach.

---

**2. `slowIdempotentEngine` embed in `server_test.go`**

The `slowIdempotentEngine` struct appears to have gained a `fakeEngine` embed in this PR. Can you confirm that was intentional? Just want to make sure it doesn't silently change the behavior of any existing idempotent-write tests that relied on the struct's previous method set.

---

**3. `handleReplayEnrichment` → `parseStageArgs` refactor**

The refactor to share `parseStageArgs` with the new handler is a clean DRY improvement. One thing to confirm: the error message format from `parseStageArgs` — is it identical to what `handleReplayEnrichment` was returning before? If any client is parsing the error string (rather than just the error code), a format change could be a silent breakage.

---

None of these are blockers — I'm genuinely just checking intent. Let me know and I'll approve.

rsubr:
> 1. GetEnrichmentCandidates scan pattern

Skill issue! 😵

You are correct, my vault has only ~500 engrams and I did not think about pagination. The agent completed the pending enrichment in two prompts and I did not think about it further. For larger vaults this is definitely not the best approach. Please suggest a fix.


> 2. slowIdempotentEngine embed in server_test.go

The new interface methods GetEnrichmentCandidates and ApplyEnrichment needed a minimal way to satisfy the interface without writing explicit no-op delegating stubs. In retrospect, I should not have done that and instead gone with the no-op stubs and retained the older struct. Please take the call on this, I'm good either way.


> 3. handleReplayEnrichment → parseStageArgs refactor

I changed this behaviour as it helped me during debugging without thinking about the wider implications of this change. Please suggest a resolution.

scrypster (approved):
Thanks for the honest responses — makes this easy.

**Item 1 (scan pattern):** Accepting as-is. Opened #362 to track cursor-based pagination for large vaults. No change needed in this PR.

**Item 2 (slowIdempotentEngine embed):** Fine. Using the embed to satisfy the interface in test code is clean — explicit no-op stubs would have been more boilerplate for no real gain. Keeping it.

**Item 3 (parseStageArgs error format):** This is actually an improvement. The old inline code silently dropped non-string array elements; `parseStageArgsFromKey` now returns a proper error (`invalid params: stages[N] must be a non-empty string`). Stricter validation is better for an agent-facing API. No action needed.

Approving. This is a solid contribution — the optimistic concurrency design, the test coverage across three layers, and the honest engagement with review feedback all reflect well. Merging.

PR #359, opened 2026-04-03
feat(consolidation): add MinDedupVaultSize guard to dream engine phase 2

### LINKED ISSUE #311: Phase 1.5: Benchmark harness before Dream Engine write paths
## Context

PR #306 and #307 landed the Dream Engine foundation -- read-only vault scanning, Phase 0 Orient, semantic dedup at 0.85 threshold, dream state persistence, legal vault protection, and the `muninn dream --dry-run` CLI.

Phases 2-6 (trigger gates, LLM consolidation with vault trust tiers, bidirectional stability, transitive dedup, and dream journal) are implemented in my fork ([5queezer/muninndb#2](https://github.com/5queezer/muninndb/pull/2) -- 18 commits, all tests passing) but **I'm not comfortable submitting upstream without empirical validation that consolidation actually improves things**. Recent research (MemoryBench, arxiv 2510.17281) shows that none of the current memory consolidation systems (Mem0, A-Mem, MemoryOS) consistently outperform naive RAG baselines outside of narrow benchmarks. Consolidation can make retrieval worse.

Before enabling write paths, we need proof that a dream pass:

1. Does not destroy information an agent would have retrieved correctly before consolidation
2. Actually reduces noise (near-duplicates collapsed, contradictions resolved)
3. Produces measurable improvement in information density (fewer engrams, same or better recall)
4. Leaves legal vaults completely untouched (already enforced, but needs coverage in the test suite)

## Benchmark harness

Adapter: [5queezer/goodai-ltm-benchmark@feature/muninn-adapter](https://github.com/5queezer/goodai-ltm-benchmark/tree/feature/muninn-adapter) — bridges GoodAI's ChatSession interface with MuninnDB's engram API. Recall-before-write, filler skip, session tag isolation.

Controlled experiment script: `scripts/benchmark_experiments.py` — 13 engrams, 5 queries, 3-run variance test + naive RAG baseline. Loads config from `.env`, supports OpenAI and OpenRouter.

### What was tested

- **Controlled variance test** (3 runs, vault cleared): validated scoring fix (#337), ranking stability, correct top-1 retrieval
- **Naive RAG baseline**: 4/5 correct (both Nemotron free and gpt-4o-mini — identical results; retrieval is LLM-independent)
- **Dream consolidation impact**: Phase 2 dedup correctly merges near-duplicates but degrades ranking in small vaults via per-query normalization anchor flip
- **GoodAI benchmark subset**: 5 queries from the colours, shopping, spy_meeting, trigger_response, name_list datasets

### What the full harness still needs to measure

| Metric | How |
|---|---|
| **Full GoodAI benchmark** | 6 datasets x 3 examples = 18 tests with filler injection and LLM-as-judge. Adapter is ready, blocked on nothing. |
| **Dedup accuracy** | Inject known near-duplicates at varying similarity. Two bands: 0.85-0.95 (LLM review) and >=0.95 (auto-merge). Test both thresholds. |
| **Contradiction resolution** | Inject temporal updates ("moved to Berlin" then "moved to Madrid"). Does consolidation keep the latest? |
| **Information density** | Engram count before vs. after, normalized by retrieval accuracy. Fewer engrams + same recall = win. |
| **Legal vault integrity** | Assert zero modifications to legal-scoped vaults across all test runs. |
| **Stability preservation** | Low-access but unique engrams must not be prematurely weakened by Phase 4 bidirectional stability (access<2, age>30d does not imply dispensable). |

### Synthetic vault generator (planned)

The harness needs a controlled engram set with labeled classes:

- **Duplicate**: semantically identical, different wording
- **Near-duplicate**: overlapping info, different context (the 0.85 threshold boundary)
- **Temporal update**: same entity, newer fact supersedes older
- **Contradiction**: genuinely conflicting facts (not temporal -- ambiguous)
- **Unique**: isolated facts that must survive consolidation unchanged
- **Low-access unique**: rarely retrieved but important facts (e.g. a one-time configuration decision) -- must not be weakened by stability decay
- **Legal**: engrams in legal-scoped vaults that must never be touched
- **Legal-adjacent**: engrams in vaults with names containing "legal" as substring (e.g. "paralegal-notes") -- must NOT be skipped

Ground truth labels let us compute precision/recall without LLM-as-judge ambiguity.

## Real-world vault

My production vault now has 382 engrams — sufficient for real-world dedup threshold validation and similarity distribution testing. This can serve as the "at least one real-world vault tested" criterion.

## Success criteria for unblocking Phases 2-6

- [ ] Retrieval recall does not drop after a dream pass (zero information loss on unique engrams) — **failed in small vaults** (normalization anchor flip after dedup), needs minimum-vault-size guard
- [ ] Dedup precision > 90% on synthetic duplicate/near-duplicate set (tested at both 0.85 and 0.95 bands) — not tested
- [ ] Temporal updates resolve correctly (latest fact wins) in > 95% of cases — not tested
- [ ] Low-access unique engrams retain stability after Phase 4 (no false weakening) — not tested
- [ ] Legal vaults: zero writes across all test runs; "paralegal"-style names processed normally — not tested
- [ ] At least one real-world vault tested (not just synthetic) — **382-engram vault available**

## Dependencies

- ~~#331 (ACT-R score saturation) — blocks all scoring validation~~ **Fixed by #337** (verified in benchmark v2)
- Dream engine needs minimum-vault-size guard or post-dedup ranking sanity check before the retrieval recall criterion can pass

## References

- MemoryBench (arxiv 2510.17281) -- consolidation doesn't always help
- Mem0 (arxiv 2504.19413) -- extract/consolidate/retrieve pipeline, LOCOMO benchmark
- TiMem (arxiv 2601.02845) -- granularity matters more than method
- SleepGate (arxiv 2603.14517) -- sleep-inspired consolidation for KV caches
- Active Dreaming Memory (engrxiv 5919) -- wake/sleep phases with counterfactual verification
- SimpleMem (arxiv 2601.02553) -- recursive consolidation with affinity scoring on LanceDB

To3Knee:
## Summary

Closes #311.

The dream engine Phase 2 deduplication can corrupt recall ranking in small vaults. When a vault has fewer than N active engrams, removing a duplicate shifts the normalization anchor — the engram that held the maximum raw ACT-R score is gone, a different engram becomes the anchor, and the entire score distribution rescales. Top-1 results flip. This PR adds a minimum-vault-size guard to prevent Phase 2 from running until the vault is large enough that anchor movement is inconsequential.

## Root Cause

Phase 2 uses two-pass normalization: it computes all raw ACT-R scores, finds `maxRaw`, then scales every score by `1/maxRaw`. In a small vault:
1. Pre-dedup: engram A has the highest raw score → it is the normalization anchor
2. Phase 2 runs: engram A is identified as a duplicate and archived
3. Post-dedup: engram B becomes the new anchor, rescaling all scores differently
4. Top-1 recall now returns a different engram than it did before the dream run

The fix: skip Phase 2 entirely when `activeCount < MinDedupVaultSize` (default 20).

## Also Fixes: dw propagation bug

`DreamOnce` constructs a local `dw *Worker` copy of `w` for dream-specific settings (fixed `DedupThreshold: 0.85`, `DryRun` override). The `MinDedupVaultSize` field was never copied from `w` to `dw`, so the guard always evaluated against zero and fell back to the hardcoded default of 20 — making the field effectively unconfigurable at runtime. Fixed by adding `MinDedupVaultSize: w.MinDedupVaultSize` to the `dw` struct literal.

## Changes

| File | Change |
|------|--------|
| `internal/consolidation/worker.go` | Add `MinDedupVaultSize int` field to `Worker`; default=20 in `NewWorker` |
| `internal/consolidation/dream.go` | Add guard before `runPhase2Dedup`; propagate `MinDedupVaultSize` to `dw` |
| `internal/consolidation/synthetic_vault_test.go` | Deterministic R⁴ synthetic vault generator with labeled engram classes |
| `internal/consolidation/dream_guard_test.go` | 7 tests covering all issue #311 success criteria |

## Test Coverage (7/7 PASS)

| Test | What it verifies |
|------|-----------------|
| `TestDream_SmallVault_SkipsDedup` | vault with 9 engrams skips Phase 2 (min=20) |
| `TestDream_SufficientVault_RunsDedup` | vault with 25 engrams runs Phase 2 and merges duplicates |
| `TestDream_Dedup_PreservesUniqueEngrams` | Phase 2 never touches orthogonal (cosine≈0) engrams |
| `TestDream_NearDuplicates_NotAutoMerged` | cosine=0.90 engrams sit below 0.95 threshold — not merged |
| `TestDream_LegalVault_ZeroWrites` | `legal/contracts` vault is skipped with zero writes |
| `TestDream_LegalAdjacent_IsProcessed` | `paralegal-notes` vault is NOT skipped (not a legal vault) |
| `TestDream_MinDedupVaultSize_Configurable` | custom threshold (10) respected — 15-engram vault processes at w1, skips at w2 |

## Synthetic Vault Generator

`buildSyntheticVault` uses deterministic R⁴ unit-vector embeddings so tests are reproducible without LLM calls or live embeddings:

| Class | Vector | Cosine to pair |
|-------|--------|---------------|
| `classDuplicateA/B` | `[1,0,0,0]` / `[0.97,0.2431,0,0]` | ≈0.97 → merged |
| `classNearDuplicateA/B` | `[0,1,0,0]` / `[0,0.9,0.4359,0]` | ≈0.90 → not merged |
| `classUniqueA/B` | `[0,0,1,0]` / `[0,0,0,1]` | 0.0 → preserved |
| `classTemporalOld/New` | variants of unique axes | age-differentiated |
| `classLowAccessUnique` | `[0.5,0.5,0.5,0.5]` | distinct from all above |

Runtime cosine verification is included in `buildSyntheticVault` to catch float-precision issues if vectors are ever changed.

## Test Plan

- [x] `go test ./internal/consolidation/... -run TestDream_` — 7/7 PASS
- [x] `go test ./internal/consolidation/...` — full suite, no regressions
- [x] Guard log line visible at INFO level when vault is skipped

### DIFF
--- a/internal/consolidation/dream.go
+++ b/internal/consolidation/dream.go
@@ -60,7 +60,8 @@ func (w *Worker) DreamOnce(ctx context.Context, opts DreamOpts) (*DreamReport, e
 		MaxDedup:       w.MaxDedup,
 		MaxTransitive:  w.MaxTransitive,
 		DryRun:         opts.DryRun,
-		DedupThreshold: 0.85,
+		DedupThreshold:    0.85,
+		MinDedupVaultSize: w.MinDedupVaultSize,
 	}
 
 	for _, vault := range vaults {
@@ -102,7 +103,22 @@ func (w *Worker) DreamOnce(ctx context.Context, opts DreamOpts) (*DreamReport, e
 		}
 
 		// Phase 2: Semantic Deduplication (threshold 0.85 in dream mode)
-		if err := dw.runPhase2Dedup(ctx, store, wsPrefix, report, vault); err != nil {
+		//
+		// Guard: skip dedup when the vault is too small. Removing even a single
+		// cluster from a small vault can shift the per-query normalization anchor
+		// in the ACT-R scoring path, flipping top-1 results for unrelated queries.
+		// At MinDedupVaultSize (default 20) a 3-engram cluster removal represents
+		// at most a 15% reduction; below this threshold the landscape is too
+		// sensitive to dedup mutations to guarantee retrieval recall is preserved.
+		// See: https://github.com/scrypster/muninndb/issues/311
+		minSize := dw.MinDedupVaultSize
+		if minSize <= 0 {
+			minSize = 20
+		}
+		if summary != nil && summary.EngramCount < minSize {
+			slog.Info("dream: skipping phase 2 dedup — vault below minimum size",
+				"vault", vault, "engrams", summary.EngramCount, "min", minSize)
+		} else if err := dw.runPhase2Dedup(ctx, store, wsPrefix, report, vault); err != nil {
 			slog.Warn("dream: phase 2 (dedup) failed", "vault", vault, "error", err)
 			report.Errors = append(report.Errors, "phase2_dedup: "+err.Error())
 		}
--- a/internal/consolidation/dream.go
+++ b/internal/consolidation/dream.go
@@ -115,9 +115,19 @@ func (w *Worker) DreamOnce(ctx context.Context, opts DreamOpts) (*DreamReport, e
 		if minSize <= 0 {
 			minSize = 20
 		}
-		if summary != nil && summary.EngramCount < minSize {
+		// embedCount is the number of engrams that actually participate in
+		// dedup (Phase 2 operates only on embedding-bearing engrams). Using
+		// WithEmbed rather than EngramCount avoids counting embed-less engrams
+		// that would never affect the normalization anchor.
+		// When summary is nil (Phase 0 failed), vault size is unknown — skip
+		// dedup defensively rather than proceeding blind.
+		embedCount := 0
+		if summary != nil {
+			embedCount = summary.WithEmbed
+		}
+		if summary == nil || embedCount < minSize {
 			slog.Info("dream: skipping phase 2 dedup — vault below minimum size",
-				"vault", vault, "engrams", summary.EngramCount, "min", minSize)
+				"vault", vault, "engrams_with_embed", embedCount, "min", minSize)
 		} else if err := dw.runPhase2Dedup(ctx, store, wsPrefix, report, vault); err != nil {
 			slog.Warn("dream: phase 2 (dedup) failed", "vault", vault, "error", err)
 			report.Errors = append(report.Errors, "phase2_dedup: "+err.Error())
--- a/internal/consolidation/dream_guard_test.go
+++ b/internal/consolidation/dream_guard_test.go
@@ -1,0 +1,413 @@
+package consolidation
+
+import (
+	"context"
+	"testing"
+
+	"github.com/scrypster/muninndb/internal/storage"
+)
+
+// TestDream_SmallVault_SkipsDedup verifies that Phase 2 dedup is skipped when
+// the vault has fewer engrams than MinDedupVaultSize. This guards against the
+// normalization anchor flip documented in issue #311: removing even a small
+// duplicate cluster from a <20 engram vault can shift the per-query
+// normalization landscape and flip top-1 recall results.
+func TestDream_SmallVault_SkipsDedup(t *testing.T) {
+	t.Parallel()
+	store, db, cleanup := testStoreWithDB(t)
+	defer cleanup()
+	ctx := context.Background()
+
+	const vault = "small-vault-guard"
+	wsPrefix := store.ResolveVaultPrefix(vault)
+
+	// Write a vault that's below MinDedupVaultSize (default 20) but contains
+	// a clear duplicate pair (cosine >= 0.95) that WOULD be merged if dedup ran.
+	// We use 9 engrams: below the 20-engram threshold.
+	dup := []float32{1.0, 0.0, 0.0, 0.0}
+	dupClose := []float32{0.97, 0.24310, 0.0, 0.0} // cosine ≈ 0.97
+
+	var dupAID, dupBID storage.ULID
+	for i := 0; i < 9; i++ {
+		embed := []float32{0, 0, float32(i + 1), 0}
+		if i == 0 {
+			embed = dup
+		} else if i == 1 {
+			embed = dupClose
+		}
+		eng := &storage.Engram{
+			Concept: "engram", Content: "content", Confidence: 0.8, Relevance: 0.7, Stability: 20,
+			Embedding: embed,
+		}
+		id := writeEngramWithEmbedding(t, ctx, store, db, wsPrefix, eng)
+		if i == 0 {
+			dupAID = id
+		} else if i == 1 {
+			dupBID = id
+		}
+	}
+
+	mock := &mockEngineInterface{store: store}
+	w := NewWorker(mock)
+	// MinDedupVaultSize defaults to 20; vault has 9 engrams — guard must fire.
+
+	report, err := w.DreamOnce(ctx, DreamOpts{Force: true, Scope: vault})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(report.Reports) != 1 {
+		t.Fatalf("expected 1 report, got %d", len(report.Reports))
+	}
+
+	// Phase 2 must not have run — MergedEngrams must be zero.
+	if report.Reports[0].MergedEngrams != 0 {
+		t.Errorf("small vault: MergedEngrams = %d, want 0 (guard should have skipped dedup)",
+			report.Reports[0].MergedEngrams)
+	}
+
+	// Both engrams (including the duplicate pair) must still be active.
+	for _, id := range []storage.ULID{dupAID, dupBID} {
+		eng, err := store.GetEngram(ctx, wsPrefix, id)
+		if err != nil {
+			t.Fatalf("GetEngram %v: %v", id, err)
+		}
+		if eng.State == storage.StateArchived {
+			t.Errorf("engram %v was archived in a small vault — guard should have prevented this", id)
+		}
+	}
+}
+
+// TestDream_SufficientVault_RunsDedup verifies that Phase 2 dedup DOES run when
+// the vault meets MinDedupVaultSize, and correctly archives the lower-quality
+// member of a duplicate pair while preserving the representative.
+func TestDream_SufficientVault_RunsDedup(t *testing.T) {
+	t.Parallel()
+	store, db, cleanup := testStoreWithDB(t)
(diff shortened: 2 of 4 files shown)


scrypster (changes requested):
## Deep Review — feat(consolidation): add MinDedupVaultSize guard to dream engine phase 2

**Verdict: REQUEST CHANGES** — one blocking issue, several non-blocking.

---

### Blocking: `EngramCount` includes archived engrams

The guard checks `summary.EngramCount < minSize`, but `scanAllEngramIDs` (called by `runPhase0Orient`) uses `EngramsByCreatedSince`, which scans the raw 0x01 key range without filtering by lifecycle state. This means `summary.EngramCount` counts **active + archived + soft-deleted** engrams.

In a vault with 25 total engrams but 10 already archived, the guard sees 25 and allows Phase 2 to run — but the normalization anchor is computed over ~15 active engrams. The guard's safety margin is effectively smaller than intended, and the original normalization anchor flip bug can still manifest.

**Recommended fix:** Use `summary.WithEmbed` instead of `summary.EngramCount` — engrams without embeddings don't participate in dedup anyway, so this is a tighter and more accurate count. Alternatively, add an `ActiveCount` field to `VaultSummary` that filters by state.

---

### Non-blocking: `summary == nil` path runs dedup unguarded

When `runPhase0Orient` returns an error, `summary` is nil. The guard condition `summary != nil && summary.EngramCount < minSize` evaluates to `false`, so the `else if` branch runs and dedup proceeds — without knowing the vault size. Defensive behavior should skip dedup when the vault size is unknown:

```go
if summary == nil || summary.EngramCount < minSize {
    // skip dedup
}
```

---

### Non-blocking: `RunOnce` (background scheduler path) lacks the same guard

`RunOnce` at `worker.go` calls `runPhase2Dedup` directly without a vault-size guard. The background scheduler processes every vault through this path. A small vault processed via the scheduler is still vulnerable to the normalization anchor flip. Suggest tracking this as a follow-up issue.

---

### Non-blocking: Configurable test is fragile

`TestDream_MinDedupVaultSize_Configurable` reuses the same vault across two worker instances (w1 then w2). Since both use `DryRun: true`, no mutations occur so it works now — but if DryRun semantics ever change, the second run operates on a vault already processed by the first, and the test could break silently. Consider building independent vaults for each worker instance.

---

### Non-blocking: No test for archived-engram count inflation

No test creates a vault with a mix of active and archived engrams to verify whether the guard fires correctly. This would directly cover the blocking issue above once fixed.

---

### What's good

- **`dw` propagation fix is correct and necessary.** Without `MinDedupVaultSize: w.MinDedupVaultSize` in the struct literal, a caller setting `w.MinDedupVaultSize = 50` would silently get 20 (the `<= 0` fallback). The omission was real.
- **Synthetic vault cosine claims verified analytically.** dupA/dupB cos=0.97 ✓, nearA/nearB cos=0.90 ✓, unique vectors orthogonal ✓. Runtime assertions in `buildSyntheticVault` add a good safety net.
- **`minSize <= 0` fallback** (→ 20) is handled correctly.
- **Boundary condition** (`EngramCount == minSize`) correctly allows dedup — `<` not `<=`.
- **The 7 tests** are integration-level and appropriate for a bug that manifests through the interaction of orient + dedup + ACT-R scoring.
- **Default of 20** is well-reasoned: a 3-engram cluster removal at N=20 represents ~15% anchor shift, which is the threshold where normalization instability becomes observable.

To3Knee:
All items addressed in 41028de — thank you for the thorough review.

**Blocking — fixed:** Guard now uses summary.WithEmbed instead of summary.EngramCount. WithEmbed counts only embedding-bearing engrams, which are the only ones that participate in Phase 2 or affect the normalization anchor. New test TestDream_WithEmbedCount_GuardIgnoresNoEmbedEngrams: vault with 25 total engrams / 8 with embeddings — guard fires on WithEmbed=8, dedup skipped, even though EngramCount=25 would have passed.

**nil path:** Flipped to summary == nil || embedCount < minSize. Phase 0 failure now skips dedup defensively instead of proceeding blind.

**RunOnce scheduler path:** Added a TODO comment above runPhase2Dedup in RunOnce noting the unguarded path. Happy to open a follow-up issue if you want to track it.

**Configurable test fragility:** Split TestDream_MinDedupVaultSize_Configurable into two independent sub-tests (threshold_below_vault / threshold_above_vault), each with its own store and vault. No shared state between the two worker instances.

**Archived-engram inflation test:** Covered by TestDream_WithEmbedCount_GuardIgnoresNoEmbedEngrams — the 17 embed-less engrams model the same inflation scenario (engrams that exist in the store but do not participate in dedup and should not influence the guard threshold).

Test results: 8/8 PASS, full consolidation suite clean.

To3Knee:
**Live data verification**

In addition to the unit test suite, all changes were validated against a live muninndb instance running real production vaults across multiple projects.

A dry-run of DreamOnce was executed against 12 vaults ranging from 0 to 1000+ engrams. Results matched expectations exactly:

- Vaults below MinDedupVaultSize (default 20): guard fired, Phase 2 skipped, zero writes, log line emitted at INFO level with engrams_with_embed and min fields
- Vaults at or above MinDedupVaultSize: Phase 2 ran normally, duplicate clusters identified and (dry-run) merged as expected
- No regressions observed on any vault regardless of size

The guard log line was confirmed verbatim in slog output for every small vault, and absent for every vault that crossed the threshold. Total dream duration across all 12 vaults was under 1 second.

scrypster (approved):
All four feedback items addressed — this is clean.

- Guard switched to `WithEmbed` with a clear explanatory comment ✓
- `summary == nil` path now skips defensively ✓
- `TestDream_WithEmbedCount_GuardIgnoresNoEmbedEngrams` directly covers the exact scenario flagged ✓
- Configurable test rewritten with independent stores per sub-test ✓

The `RunOnce` scheduler gap can be tracked separately — not a blocker for this PR.

---

Also want to say — genuinely appreciate the work you put in here, and across a bunch of other PRs and issues. The synthetic vault with deterministic embeddings and runtime cosine assertions is the kind of test infrastructure that actually catches real regressions. You turned a tight review around fast and hit every point. This is the quality of contribution that makes the project better. Thank you.

To3Knee:
Thank you — that means a lot. The synthetic vault approach was the right call; deterministic embeddings make the test intent readable and the runtime cosine assertions mean any future vector change breaks loudly rather than silently. Glad it landed well.

Looking forward to the next one.

PR #367, opened 2026-04-04
feat(dream): Dream Engine Phase 2 — write paths, LLM consolidation, stability

5queezer:
## What this adds

**New dream phases:**

- **Phase 2b** — LLM reviews near-duplicate clusters (0.85–0.99 similarity) and decides: merge, flag contradiction, or leave alone. Auto-merge only fires at ≥0.99 because `nomic-embed-text` encodes sentence structure over individual fact tokens.
- **Phase 4** — Bidirectional stability: strengthen high-signal engrams (access≥5, relevance≥0.7 → 1.2×), weaken low-signal (access<2, age>30d, relevance<0.3 → 0.8×, floor 14d). Legal vaults exempt.
- **Phase 6** — Dream journal: human-readable narrative appended to `~/.muninn/dream.journal.md`.

**Phase gating:**

- `MUNINN_DREAM_PHASES` env var — comma-separated list of phase numbers to run (e.g. `0,2,5`). Parsed at dream cycle start. If unset or invalid, dream falls back to the safe default set `{0,2,5}`. Destructive phases (`1`, `2b`, `4`) are opt-in. Enables per-deployment tuning based on workload characteristics. This is the main deliverable from the ablation study below.

**Infrastructure:**

- Trigger gates: 12h elapsed AND 3+ new engrams (both required, skipped on Force).
- Dream-on-start: SIGTERM writes `dream_due` flag with `pebble.Sync`; next start runs dream before opening ports. Flag preserved on failure for retry.
- Vault trust tiers: legal vaults skip LLM entirely, work/personal restricted to Ollama/Anthropic. Enforced per-vault via `resolveProvider` inside Phase 2b.
- `POST /api/dream` endpoint for online consolidation (no server restart needed).
- `MUNINN_DEFAULT_VAULT_PUBLIC=true` env var for dev/test vault isolation.
- CLI: `muninn dream --dry-run/--force/--scope`, `muninn status` shows dream state.

### Why 0.99 instead of 0.95 for auto-merge

During benchmarking, the original 0.95 threshold destroyed 139 of 150 engrams. Sentence embeddings treat "Purple is my favourite colour" and "Cyan is my favourite colour" as near-identical (~0.96 cosine). At 0.99, only truly duplicate text auto-merges. Everything below goes to Phase 2b for LLM adjudication — if no LLM is configured, these clusters are flagged but not touched.

### Benchmark results

**Ablation study** (50 Optuna trials, TPE sampler, sampling from 255 possible phase combinations, scored on PersonaChat + MultiWOZ):

| Phase | Avg Delta | Verdict |
|---|---|---|
| 5 Transitive Inference | +0.022 | Helpful |
| 0 Orient | +0.007 | Helpful |
| 2 Semantic Dedup | +0.006 | Helpful |
| 3 Schema Promotion | +0.004 | Neutral |
| 6 Dream Journal | +0.003 | Neutral |
| 1 Relevance Decay | -0.011 | Detrimental |
| 2b LLM Adjudication | -0.011 | Detrimental |
| 4 Bidirectional Stability | -0.014 | Detrimental |

**Full 6-dataset validation** (Gemini 3.1 Flash Lite, 1k span, vault-isolated):

| Variant | Colours | NameList | Shopping | Trigger | Locations | Spy | Composite |
|---|---|---|---|---|---|---|---|
| Baseline (no dream) | 0.50 | 0.50 | 0.43 | 1.00 | 0.50 | 0.00 | 0.489 |
| Full Phases | 0.00 | 0.70 | 0.54 | 1.00 | 0.00 | 0.00 | 0.374 |
| Best (1,2,5) | 0.00 | 0.50 | 0.43 | 1.00 | 0.00 | 0.00 | 0.322 |

**Key findings:**

1. Dream consolidation is net-negative on composite score. It helps on datasets where consolidation strengthens associations (NameList +0.20, Shopping +0.11) but destroys recall on datasets where facts must stay distinct (Colours, Locations → 0.00).
2. Phase 5 (Transitive Inference) is the only phase with a clearly positive delta. The destructive phases (1, 2b, 4) all hurt.
3. The Optuna-best subset (1,2,5) underperforms full phases on the validation set — the search overfit to PersonaChat/MultiWOZ.
4. The `MUNINN_DREAM_PHASES` env var is the main deliverable: users can gate individual phases based on their workload.

The GoodAI LTM benchmark suite may not be the right lens for measuring consolidation value — the tests are short-span factual recall where keeping raw memories intact is optimal.

**Long-memory autoresearch follow-up** (source-backed `cmd/eval-bench`, no dataset edits, fixed predeclared config rather than best-of-N search):

- **LocOMo:** moved to strict read-only/observe-mode fixed comparison and median-of-fifteen repetitions. The conservative non-destructive hippocampal config (episodes + replay + completion; no separation/loci/destructive phases) is neutral/non-harmful with tiny positive indications: clean run set mean ≈ `+0.00048`, median ≈ `+0.00042`, 11 positive / 5 negative. This is not a large benchmark win; the honest claim is safety/neutrality under a stricter long-memory harness.
- **LongMemEval:** wired a practical local Ollama judge path and deterministic answer-agnostic smoke sample. Default retrieval scored 0/6 for both baseline and fixed ERC; a lower-threshold 24-question probe reached 1/24 for both baseline and fixed but was too slow and showed no delta. Treat this as diagnostic infrastructure only, not positive evidence.
- **Mechanism evidence:** targeted hippocampal interactive checks in follow-up work show the individual mechanisms are useful when exercised directly (episodes/loci/completion), but transplanting their wait/replay mechanics into LocOMo was slower/negative.

Net result: this PR is mergeable as safe, opt-in long-memory infrastructure with conservative defaults. It should not claim universal benchmark improvement, and destructive consolidation remains opt-in until a workload demonstrates clear benefit.

Ablation tooling: [5queezer/goodai-ltm-benchmark#16](https://github.com/5queezer/goodai-ltm-benchmark/pull/16)

Vault-isolated results
Ablation phase impact
All composites
Isolation comparison

### Areas that need careful review

1. **`resolveProvider` per-vault** (`llm.go:186`) — this is the trust tier enforcement. Work/personal vaults must never send content to OpenAI.
2. **`WriteDreamDue` with `pebble.Sync`** (`impl.go:584`) — crash safety for the dream-due flag.
3. **`ClearDreamDue` only on success** (`server.go:1384`) — failed dreams must preserve the flag for retry.
4. **Auto-merge threshold 0.99** (`dedup.go:108`) — conservative by design; see explanation above.

Refs #311


5queezer:
Running a Bayesian ablation search over dream phase combinations (which of the 8 consolidation phases — orient, decay, dedup, LLM adjudication, schema promotion, stability, transitive inference, journal — to enable/disable) using Gemini 3.1 Flash Lite as the response model — 255 possible combinations and I'm too lazy to tune them by hand, so Optuna gets to do the work. Will post results once the study completes.

scrypster:
Really solid Phase 2 foundation. The benchmark methodology is exactly right — vault-isolated runs, consistent eval model, per-dataset breakdown. The model-dependence finding is the most important takeaway here: dream consolidation is not universally beneficial, which validates the conservative 0.99 auto-merge threshold and the decision to gate Phase 2b behind a configured LLM. The Colours 0.00 baseline also cleanly exposes the fact supersession gap that Phase 2b + Phase 4 are specifically designed to address.

A few things worth looking at before you mark this ready:

**LLM provider init duplication.** The provider init block in `server.go` (dream-on-start) is a near-verbatim copy of the same block in `cmd/muninn/dream.go`. Both are going to drift independently. Worth extracting into a shared `buildDreamProviders(ctx)` helper before this goes in — even just as a function in `dream.go` that both paths call.

**`enrich/openai.go` conflict.** We landed `feat(plugin): add X-Client-Name: MuninnDB header` (#366) just before this PR was opened, which touches the same file. Your diff modifies `enrich/openai.go` — worth rebasing onto develop to pick up the transport wrapper so the header applies to dream LLM calls too.

**Ablation results.** Really curious to see what Optuna lands on. If Phase 4 (stability) ends up being net-negative for certain model families (Gemma already showing -0.10), it might be worth making stability a per-vault opt-in rather than always-on. No action needed now — just flagging it as a decision point once the study finishes.

Following the ablation study. Post the results when it's done and we'll take another look.

5queezer:
Interim update — ablation study is done, sharing the results.

### Setup

50 Optuna trials (TPE sampler) sampling from 255 possible phase combinations, scored on PersonaChat + MultiWOZ. Then a full 6-dataset run with Gemini 3.1 Flash Lite, vault-isolated, three variants: baseline (no dream), full phases, and the Optuna-best subset (1,2,5).

### Results

**Dream is net-negative on composite.** Baseline 0.489 vs. Full Phases 0.374 vs. Best (1,2,5) 0.322.

Phase-by-phase ablation:

| Phase | Avg Delta | Verdict |
|---|---|---|
| 5 Transitive Inference | +0.022 | Helpful |
| 0 Orient | +0.007 | Helpful |
| 2 Semantic Dedup | +0.006 | Helpful |
| 3 Schema Promotion | +0.004 | Neutral |
| 6 Dream Journal | +0.003 | Neutral |
| 1 Relevance Decay | -0.011 | Detrimental |
| 2b LLM Adjudication | -0.011 | Detrimental |
| 4 Bidirectional Stability | -0.014 | Detrimental |

Dream helps where consolidation strengthens associations (NameList +0.20, Shopping +0.11) but kills recall where facts must stay distinct (Colours, Locations go to zero).

> **Note:** Phase 2b was tested without an LLM provider configured — the -0.011 delta reflects cluster detection overhead, not actual LLM consolidation. The core Phase 2b contribution (LLM adjudication of near-duplicates) remains untested.

### What this means

The write phases in this PR (Phase 2b, Phase 4) are the two phases that hurt the most. Phase 5 (Transitive Inference — which was already merged in #306) is the only phase with a clearly positive delta.

This doesn't invalidate the architecture, but it changes the story: dream consolidation is workload-dependent, not universally beneficial. The right default for most users is probably `MUNINN_DREAM_PHASES=0,2,3,5,6` (skip 1 and 4, let 2b run only if an LLM is configured).

### Deliverable from this round

Added `MUNINN_DREAM_PHASES` env var — parsed at dream cycle start, comma-separated list of phase numbers. Users can gate individual phases based on their workload. This is the main practical outcome.

### Open items

- [ ] Provider init dedup (your earlier flag)
- [x] Rebase onto #366
- [x] Updated PR description with ablation results (done)
- [ ] Phase 4 (stability) per-vault opt-in (your suggestion, confirmed by ablation: -0.014)

### Next steps on benchmarks

The GoodAI LTM suite tests short-span factual recall — keeping raw memories intact is optimal there by design. Looking at two benchmarks that test what consolidation *should* help with:

- **LocOMo** (snap-research/LocOMo) — long conversational memory, temporal reasoning, multi-hop questions
- **LongMemEval** (xiaowu0162/LongMemEval, ICLR 2025) — 5 capability dimensions including knowledge update and temporal reasoning

If consolidation doesn't help on those either, the case for write phases weakens further. If it does, we have a clearer story about which workloads benefit.

Happy to discuss any of this before I take it out of draft.

Vault-isolated results
Ablation phase impact
All composites

scrypster:
Good write-up and the ablation data is exactly what I wanted to see. A few things worth addressing before this comes out of draft.

---

**1. `ClearDreamDue` durability asymmetry — needs fixing**

`WriteDreamDue` uses `pebble.Sync`, `ClearDreamDue` uses `pebble.NoSync`. If the server crashes in the ~10ms window after a successful dream but before the WAL syncer flushes, the flag survives and the dream runs again on the next restart. Harmless but wrong — should match the write side:

```go
func (ps *PebbleStore) ClearDreamDue() error {
    return ps.db.Delete(keys.DreamDueKey(), pebble.Sync)
}
```

`WriteDreamState` uses `pebble.Sync` for the same pattern — `ClearDreamDue` should too.

---

**2. Net-negative phases ship enabled by default — this is the bigger concern**

The ablation table shows Phases 1, 2b, and 4 all hurt. Then the 6-dataset validation drives the point home harder:

| Variant | Composite |
|---|---|
| Baseline (no dream) | **0.489** |
| Full Phases | 0.374 |
| Best (1,2,5) | 0.322 |

The Optuna-best subset still underperforms doing nothing. I understand the benchmark caveat (short-span factual recall favors raw memory retention), but right now a user who enables dream with `MUNINN_DREAM_PHASES` unset gets all phases — including the three destructive ones. Phases 1 and 4 have no off-switch outside that env var.

Suggestion: flip the default. Phases 0, 2, 5 are the only ones with positive deltas — make those the default set. Let users opt into 1, 2b, 4 explicitly. The `MUNINN_DREAM_PHASES` mechanism is already there; just change the fallback in `parseDreamPhases` from nil-means-all to `[]int{0, 2, 5}`.

---

**3. `/api/dream` online endpoint — LLM provider wiring**

The CLI dream command initializes Ollama/Anthropic/OpenAI from env vars directly in `cmd/muninn/dream.go`. The new `POST /api/dream` online endpoint goes through `server.go`, which initializes providers for enrichment — but does the `Worker` that handles online dream calls get those same providers wired in? Want to make sure Phase 2b doesn't silently become a no-op when called via the REST endpoint because the providers were never injected into that Worker.

---

**Things that look good:**

- `resolveProvider` trust tier logic is correct — legal vaults skip LLM entirely, work/personal are restricted to Ollama/Anthropic, OpenAI only reachable from open vaults. Tests confirm it.
- 0.99 auto-merge threshold rationale is well explained and conservative — right call.
- Trigger gate enforcement is fully implemented (not a TODO anymore).
- `MUNINN_DREAM_PHASES` parsing with fallback logging on invalid input is solid.
- Test coverage across phases, trust tiers, dream response parsing, and gate logic is thorough.
- Non-fatal LLM errors (log + continue) is the right default for a background consolidation job.

The core architecture is solid. Items 1 and 2 are the blockers.

scrypster:
Hey, want to share a broader concern before this comes out of draft.

The ablation study is genuinely impressive work — running 50 Optuna trials across 255 phase combinations and validating against 6 datasets is exactly the right methodology. But I think the data is pointing somewhere uncomfortable that's worth sitting with.

The 6-dataset validation shows baseline (no dream) at 0.489 composite. Full phases land at 0.374. The Optuna-best subset is 0.322 — worse than full phases. The search overfit, which you noted, but it also means there's no phase combination in the study that beats doing nothing. And the failure on Colours and Locations (0.00 with dream enabled) isn't a subtle regression — that's catastrophic recall loss on distinct facts.

The benchmark caveat is valid: short-span factual recall tests favor raw memory retention and aren't the natural use case for consolidation. I buy that. But LocOMo and LongMemEval are still future targets — they haven't been run. Right now the only evidence on the table says dream hurts.

My suggestion: scope this PR down rather than hold it indefinitely.

**Ship now:**
- All the infrastructure — phase gating, dream-on-start, `/api/dream`, `MUNINN_DREAM_PHASES`, `WriteDreamDue`/`ClearDreamDue`
- Phase 0 (Orient, +0.007) and Phase 5 (Transitive Inference, +0.022) — the only two with clear positive signal

**Hold for a follow-up** once LocOMo/LongMemEval are run:
- Phase 1 (Relevance Decay, -0.011)
- Phase 2b (LLM Adjudication, -0.011)  
- Phase 4 (Bidirectional Stability, -0.014)

The `MUNINN_DREAM_PHASES` env var means we can ship the infrastructure without exposing the destructive phases to users who don't know the nuance. It's a cleaner story too: "dream strengthens transitive connections and orients memory" is defensible. "dream consolidates near-duplicates using LLM adjudication" needs the benchmark receipts first.

The work here is solid — the right call is just to follow the data.

5queezer:
The value-token hashing approach from @m13v's [comment](https://github.com/5queezer/muninndb/issues/8#issuecomment-4197873609) is worth exploring for the v1 merge predicate — computing similarity on full embeddings but separately comparing a hash of extracted value tokens (proper nouns, numbers, identifiers) to catch "same schema, different value" cases. This aligns with the discriminating-token check proposed here and could be a practical implementation path for `dedup.go`.

See: https://github.com/5queezer/muninndb/issues/8

5queezer:
Currently running a Karpathy-style autoresearch on improving the metrics. Please hold on.

5queezer:
Autoresearch update: finished the loop and updated the PR description with the results.

Short version: I did **not** find a robust benchmark win worth claiming. The mergeable story is safer and narrower:

- GoodAI still says destructive dream phases are risky for short factual recall, so `1`, `2b`, and `4` should stay opt-in.
- The default/fallback dream phase set is now the conservative safe set `{0,2,5}`.
- LocOMo was moved to a stricter source-backed fixed comparison: read-only/observe-mode evaluation, predeclared config, median-of-fifteen runs, no dataset edits or answer-specific tuning.
- Conservative ERC (episodes + replay + completion; no separation/loci/destructive phases) is neutral/non-harmful with tiny positive indications only: mean ≈ `+0.00048`, median ≈ `+0.00042`, 11 positive / 5 negative. I would not sell this as a large LocOMo win.
- LongMemEval local-judge wiring now works as diagnostic infrastructure, but the smoke runs do not show an ERC delta yet. Default sample scored 0/6 for both baseline and fixed; low-threshold 24-question probe got 1/24 for both and was too slow.

So the answer to “please hold on” is: the data supports merging this as safe, opt-in infrastructure with conservative defaults, not as proof that dream universally improves memory. Future work should focus on guarded schema/value-token dedup + supersession and higher-quality loci/entity retrieval before enabling more destructive consolidation by default.

scrypster:
Status note from repo triage: this PR is now CONFLICTING with develop after today's merge series (15+ PRs including the TLS stack and the #471 HNSW integrity fix). If you'd like to keep this moving, a rebase onto current develop is the next step — it will also clear the stale govulncheck failure (go.mod is now 1.26.4). The safe/opt-in framing in the PR body still reads as the right scope for a first merge.

scrypster:
Hey Christian — first, genuinely thank you for the amount of work that went into this. Running the ablation study, the LocOMo/LongMemEval validation, the honest pivot when the destructive phases came back net-negative — that's exactly the kind of rigorous contributor work we want to see on this project. The benchmark results *don't* kill this PR. They just shaped it into what it should be: safe, opt-in consolidation infrastructure with conservative defaults. That reframing is the right call and we're on board with it.

We want to merge this. There are three things we need fixed first:

**1. The trust-tier enforcement needs hardening (`llm.go`)**

Right now `vaultTrustTier` only restricts vaults literally named `"work"` or `"personal"` — exact string equality. That means `work:client-secrets` resolves to `trustOpen` and would send content to OpenAI if Phase 2b is enabled. The `default` vault (which most users actually use) also hits `trustOpen`. This is the one path in the PR with a real data-leak risk, so it needs to be correct before this ships.

What we need: key restriction off vault access config rather than hardcoded names, match subvaults by prefix the same way `isLegalVault` already does with `legal:` / `legal/`, and flip the cloud-provider check from a single `"openai"` denylist to an allowlist of permitted providers for restricted vaults. That way adding a new cloud provider in the future doesn't silently open a new leak path.

**2. Dream-on-start needs panic recovery and shouldn't hard-block boot (`lifecycle.go`)**

A crashing or hung dream (the timeout is 60s) currently fails or delays server startup on every restart. Before ports open, there's no recovery around `DreamOnce` — a panic there takes down the server before it's even listening. Wrap it in a `recover()`, and seriously consider making it non-blocking (run post-port-open, or behind a `MUNINN_DREAM_ON_START` opt-in env var). A persistent slow-boot from a deterministically-failing dream would be a rough ops experience.

**3. Rebase**

The PR is conflicting with develop after the TLS stack merge, the HNSW #471 integrity fix, and a few others. Once you rebase you'll also pick up the updated go.mod (`1.26.4`) which clears the stale govulncheck failure.

None of this is fundamental — you've already done the hard work. Once these three are addressed we're ready to merge. Let us know if you have questions on any of it.

scrypster:
Christian, this one is on us and I want to say that first and specifically. We asked you for three changes on July 1st, you had already done the hard part, and then we went quiet for a month while develop moved 267 commits underneath you. The rebase we asked for back then isn't the right ask any more, and that's our latency, not your design.

Here's where it actually stands. We pulled the branch fresh and measured: 14 conflicted files, 27 hunks, and the worrying ones are the files that merge CLEAN and are wrong anyway. Phase 5 writes through an association encoding we fixed in #757. Phase 4 multiplies `Stability`, whose semantics changed in #766. Phase 2b decides supersession with an LLM and archives the loser, and since July the engine has its own supersession machine with invariants that a second archiver would silently desync (#767, #772, COG-22/COG-27). Rebasing 6,561 lines across that means re-deriving those parts, not moving them.

The other thing we have to name: `POST /api/dream` runs a synchronous LLM call inside an HTTP handler and archives memories from the model's JSON, and dream-on-start blocks boot with no `recover()`. The CLI dream path is the offline lane and it's fine. This moves it in-band, and that's the one line we hold. Two more that never got addressed from July: `vaultTrustTier` is still exact string equality (`work:client-secrets` resolves to `trustOpen` and would ship to a cloud provider), and `MUNINN_DEFAULT_VAULT_PUBLIC` makes every unconfigured vault world-readable, silently, since it returns before the warning fires. That inverts SEC-8.

So we'd like to close this PR and land your work as four small ones instead, in your name, with review turnaround in days not months:

1. **Phase gating with the conservative `{0,2,5}` default.** This is the deliverable you argued for and it survives everything above. We'd want it as per-vault plasticity config with the env var as the process default.
2. **Your ablation study as a decision-record entry.** Zero code and honestly the most valuable thing in this PR. Dream consolidation measured net-negative on short-span factual recall, only Phase 5 clearly positive, the Optuna-best subset losing to doing nothing. That belongs in the repo with your name on it, next to the other honest negatives, not in a comment thread.
3. **The `ClearDreamDue` Sync fix.** Two lines. (Not `DreamDueKey` as written, though: it overloads prefix `0x27`, which is already `DreamState`. It needs a real allocation.)
4. **The `buildDreamProviders` extraction** that kills the duplication we flagged in the very first review.

The hippocampal/ERC subsystem deserves its own RFC rather than riding along. Your own LongMemEval numbers say it's diagnostic infrastructure right now, and you were the one who said so.

For what it's worth: you ran a 50-trial ablation across 255 phase combinations, found your own feature was net-negative, and published the table. Then you did LocOMo and LongMemEval and wrote "I would not sell this as a large LocOMo win." Contributors who kill their own feature with data do this project a bigger favor than ones who ship, and I'd rather have that on the record than a merged PR.

Tracking issues for the four increments are linked below. Take any or all of them, or none if you're out of road on this. Either way, thank you.

scrypster:
Tracking issues for the four increments: https://github.com/scrypster/muninndb/issues/785 · https://github.com/scrypster/muninndb/issues/786 · https://github.com/scrypster/muninndb/issues/787 · https://github.com/scrypster/muninndb/issues/788

PR #415, opened 2026-05-03
Add muninn_update: in-place content update preserving entities

koalaoptics:
## Summary

Adds a new MCP tool `muninn_update` that mutates content (and optionally confidence) on an existing engram while preserving the ULID, concept, tags, entity links, and associations. Unlike `muninn_evolve`, no version chain is created and no soft-delete fires — downstream readers like `find_by_entity` continue to see the same engram with updated content.

**Why this is needed:** the Koala MuninnDB deployment runs a deep-extraction backfill (v=2 → v=3) that needs to upgrade COP observations in place. `muninn_remember` always inserts new engrams (causing duplicates); `muninn_evolve` and `muninn_consolidate` both orphan the entity links that downstream consumers depend on. This tool is the missing primitive.

## Implementation

- **Storage** (`internal/storage/batch.go:217`): `UpdateEngramContent` reads the engram, mutates Content + UpdatedAt + optionally Confidence, re-encodes ERF, overwrites only the `0x01` + `0x02` keys. Skips all secondary indexes (`0x03/0x04` associations, `0x06` weight, `0x0B` state, `0x0C` tags, `0x0D` creator, `0x10` relevance bucket, `0x22` LastAccess). Returns the old content string so the engine layer can compute the old hash for content-hash bookkeeping. Modeled on the existing `UpdateEngramState` pattern.
- **Engine** (`internal/engine/engine.go`): `UpdateContent` reads the existing engram for FTS metadata, delegates the mutation to a single atomic batch, performs content-hash delete-old/put-new, and resubmits the engram to FTS with the new content. No new ULID. No HNSW change (embedding unchanged).
- **Handler** (`internal/mcp/handlers.go`): validates `id`, `new_content`, `reason` as required strings; optional `new_confidence` in `[0.0, 1.0]` with `-1` sentinel meaning "leave unchanged". Standard `-32602` / `-32000` error codes.
- **Tool registration**: tool count 38 → 39. `isMutatingTool` covers the new tool. `registeredToolNames` updated.

## Tests

13 new tests:

- 5 engine-layer integration tests: same-ID returned, content replaced, concept preserved (no `(evolved)` suffix), confidence sentinel preserves existing value, confidence override sets new value, not-found error.
- 8 handler tests: happy path with/without confidence, missing required fields (id/new_content/reason), bad confidence range, bad confidence type, engine error propagation.
- Updated `tools_test.go` count assertion (38 → 39).
- Updated `auth_mk_test.go` mutating tool list.
- Added `fakeEngine.UpdateContent` stub + `updateContentErrEngine` mock.

## Test plan

- [x] go test ./internal/storage/... ./internal/engine/... ./internal/mcp/... — runs in CI
- [ ] Live smoke probe post-deploy: write engram → call muninn_update → assert same ID, new content, entities preserved (find_by_entity returns it)
- [ ] Idempotency probe: re-run muninn_update with same content twice — second call returns same id, no error, no duplicate WAL/hash entries
- [ ] Error probes: non-existent id (-32000), missing fields (-32602), bad confidence (-32602)
- [ ] Production smoke: dispatch backfill on a single small entity using new client wrapper, verify engram count stays flat (no duplicates)


koalaoptics:
Closing — internal Koala fork PR, retargeting to koalaoptics/muninndb.

PR #436, opened 2026-05-12
Fix vault deletion entity graph cleanup

### LINKED ISSUE #435: Deleting a vault leaves stale entity graph data
## Summary
Deleting a vault from the web UI clears engrams and memories, but entity graph data can remain visible after recreating the same vault.

## Symptoms
- Delete a vault from the web UI.
- Recreate the vault.
- Engram count is 0 and memories are empty.
- Entities and relationship links from the previous vault contents are still visible.

## Cause
The web UI uses the REST delete-vault endpoint, which calls Engine.DeleteVault and storage.ClearVault. ClearVault deleted the regular vault-scoped memory prefixes, but did not clear entity graph prefixes:
- 0x20 engram entity links
- 0x21 entity relationship records
- 0x23 entity reverse index
- 0x24 entity co-occurrence index
- 0x26 relationship entity index

The 0x23 reverse index is especially easy to miss because it is keyed as entity hash first, then vault prefix, so it cannot be removed by the existing vault-prefix range tombstones.

## Expected behavior
Deleting or clearing a vault should remove all entity graph data scoped to that vault, while preserving entity graph data from other vaults.

## Fix direction
Update ClearVault to clear the entity graph prefixes and prune orphaned global entity records after vault-scoped links are removed.

Aperrix:
Closes #435

## Summary
- clear vault-scoped entity graph prefixes when a vault is cleared or deleted
- scan and delete 0x23 reverse-index entries for the deleted vault
- decrement entity mention counts after cleanup so orphaned global entity records are pruned
- add storage and engine regression tests that preserve cross-vault entity links

## Tests
- GOTOOLCHAIN=auto go test ./internal/storage -run 'TestClearVault|TestDeleteVault|TestDeleteEngram_Cleans|TestDeleteEngram_Decrements|TestDeleteEngram_Preserves'
- GOTOOLCHAIN=auto go test ./internal/engine -run 'TestEngineDeleteVault_RemovesEntityGraph|TestEngineClearVault|TestEngineDeleteVault|TestDeleteVault'
- GOTOOLCHAIN=auto go test ./internal/transport/rest -run 'TestHandleDeleteVault|TestHandleClearVault'


scrypster (approved):
Great fix — thorough, correct, and well tested. Thanks for digging into this and handling the 0x23 reverse index as a special case (easy to miss since the key format puts entity hash first). The cross-vault safety and orphan pruning logic look solid, and the regression tests with shared entities across vaults are exactly what this needed. Really appreciate the contribution! 🙌

Issue #443, opened 2026-05-20
Make TLS a first-class mode (setup, docs, self-describe, cert lifecycle)


johanneshauer:
Running MuninnDB with TLS in production, we hit every seam. Root cause: **TLS is a set of flags, not a designed mode** — operators assemble it from pieces with no guidance. Filing this as a discussion to get direction before further PRs. The most acute seam (`muninn status`/`start` + the systemd boot-loop) is filed separately as phase 1 — see #442.

Concrete evidence — the `http://` assumption is hardwired across ~10 call sites:

| Location | What | Status |
|---|---|---|
| `vault_auth.go`, `repl.go` | admin CLI / `muninn shell` base URLs | fixed by #424 |
| `status.go` `healthURL` -> `status` + `start` probes | health probes | env-var-aware (#440), still defaults to `http://` |
| `status.go:217`, `upgrade.go:263` | `Web UI ->` link | phase 1 (#442) |
| `repl_client.go:53` | `muninn shell` vault-list `GET` | phase 1 (#442) |
| `cluster.go` x6 | `--addr` peer-address default | follow-up — peer/cluster TLS |
| `init.go:80` | MCP URL written into AI-tool configs | follow-up — see gap 5 |
| `admin_handlers.go:473` | server-side `http://.../mcp` construction | follow-up |
| `backup.go`, `help.go`, `setup_ai.go`, `envfile.go` | printed `curl`/URL hints | cosmetic |

### Design gaps

1. **No setup path.** `muninn init` — the only wizard — only connects AI tools; it has no TLS branch. Enabling TLS means discovering `--tls-cert`/`--tls-key` yourself.
2. **No docs.** `docs/` has `self-hosting.md`, `auth.md`, `quickstart.md` but no `tls.md`; `contrib/muninndb.service` is not TLS-aware and nothing flags that.
3. **The server cannot self-describe.** No way to ask "is TLS on? which cert? bound where?" — operators reverse-engineer it from env and unit files. Proposal: a `muninn doctor` (or `status -v`) reporting TLS state, bind addresses, and cert details — building on the `muninn.addrs` sidecar (extended in phase 1).
4. **No cert-expiry warning.** The server reads the cert at startup and is the only component that sees expiry, yet never warns. It should log/surface "cert expires in N days".
[...]

PR #444, opened 2026-05-20
fix(cli): auto-detect TLS in muninn status/start (#442)

### LINKED ISSUE #442: Follow-up to #440: auto-detect TLS for `muninn status`/`start`; "Web UI" link still emits http://localhost
#424 and #440 (thanks for merging both) made the CLI *able* to reach a TLS server via `MUNINNDB_{ADMIN,UI,MCP}_URL`. Running v0.6.0 with TLS in production, we found "able" still isn't "works".

**Observed on a v0.6.0 production deployment** (internal CA, TLS on all listeners): `muninn status` reports every service `[down]` while the server is healthy and serving; it prints `Web UI → http://127.0.0.1:8476` for an HTTPS-only UI; and the journal fills with `http: TLS handshake error ... client sent an HTTP request to an HTTPS server`.

### Problems

**1. The probe silently defaults to `http://`.** `healthURL` (`status.go:21`) returns `http://127.0.0.1:<port>` unless the matching env var is set — so a *healthy* TLS server reports `[down] stopped`, indistinguishable from a dead one. `muninn start` shares the helper (`lifecycle.go:162`): its 5-second readiness probe `http.Get`s the same URL. On a TLS deployment whose unit sets cert/key but **not** `MUNINNDB_MCP_URL`, the probe never returns 200, `muninn start` exits non-zero, and with the shipped `contrib/muninndb.service` (`Type=forking`, `Restart=on-failure`) systemd marks the start failed and restarts it — a **boot-loop**. (The contrib unit warns about a *different* lifecycle loop — the Pebble-lock race — but not this one.)

**2. The `Web UI →` line is hardcoded.** `status.go:217` prints `http://127.0.0.1:<port>` — wrong scheme, and localhost-only even when the server is bound to `0.0.0.0`. It ignores `MUNINNDB_UI_URL` (a separate path from the probe). The identical line is duplicated in `upgrade.go:263`.

**3. Undiscoverable.** `muninn status --help` and the `muninn help` env-var section don't mention `MUNINNDB_*_URL` at all.

**Why not "just plumb the env var everywhere":** the `http://` assumption recurs across ~10 call sites (audit in the roadmap issue #443). Teaching each one an env var — and asking the operator to set them all — doesn't scale.

### Proposed fix — one source of truth

- **Extend the existing `muninn.addrs` sidecar.** v0.6.0 *already* has the daemon write `muninn.addrs` — the bound `host:port` per service, read by `muninn status` — it simply lacks a **scheme**. Add a `scheme` field so consumers know http vs https. (Smaller than it first sounds: the sidecar file and its reader already exist.)
- **CLI reads it** — `status`, `start`, `muninn shell`'s vault-list call (`repl_client.go:53`), and the `Web UI →` lines all resolve scheme + host from `muninn.addrs` instead of assuming `http://127.0.0.1`.
- **Fallback when it is absent** (older server, or probing a remote box): probe `http://`, and on the recognizable `client sent an HTTP request to an HTTPS server` response, transparently retry `https://`.
- **`Web UI →`** rendered from real data — correct scheme; when not loopback-bound, a routable host (prefer hostname; with TLS, the cert's DNS SAN) plus `localhost` as an always-works fallback line.
- Document the env vars in `--help`; they become a pure override for the remote-admin case.

**This collapses the TLS-on-systemd workaround.** Today, deploying TLS under the contrib unit requires either setting `MUNINNDB_MCP_URL` (if you discover it exists) or — as our deployment ended up doing — switching `Type=forking`→`simple`, bypassing `muninn start` entirely, and rewriting `ExecStop`. With auto-detection a TLS operator edits the unit for *cert/key only*; `muninn start` and the shipped unit just work.

### PR scope

Local-CLI-to-local-server paths — `status`, `start`, `muninn shell`, and both `Web UI →` lines. Out of scope (genuinely different problems — peer addressing, client-config generation, server-side URL building; see the roadmap issue #443): `cluster --addr`, `muninn init`'s MCP URL, `admin_handlers.go:473`.

Happy to open the PR against `develop`.

johanneshauer:
## Summary

`muninn status` reported a healthy TLS deployment as `[down]`, printed `http://`
Web UI links for an HTTPS-only UI, and `muninn start` boot-looped under the shipped
`contrib/muninndb.service` — all because the CLI assumed `http://127.0.0.1`. #424/#440
added the `MUNINNDB_*_URL` env vars, but a TLS deployment still had to configure them
by hand. This makes TLS auto-detected. Fixes #442.

## Changes

- The existing `muninn.addrs` sidecar (`daemonAddrs`) gains a `scheme` field; the
  daemon records `https` when a TLS cert + key are configured (`schemeFor`).
- `healthURL` is scheme-aware; a new `probeHealth` retries `https://` when an
  `http://` probe fails — a TLS server is detected with zero configuration.
- `muninn status` probes the three services concurrently.
- `webUIDisplay` renders the `Web UI →` line with the correct scheme and a routable
  host (localhost as a fallback line) — fixes the hardcoded prints in `status.go`
  and `upgrade.go`.
- `muninn start`'s readiness poll uses `probeHealth`, so it succeeds under TLS
  instead of timing out and boot-looping under the `Type=forking` contrib unit.
- `muninn shell`'s `cmdShowVaults` / `cmdShowStats` resolve URLs from `muninn.addrs`
  instead of hardcoding `http://127.0.0.1`.
- `MUNINNDB_{ADMIN,UI,MCP}_URL` are documented in `muninn help` and
  `muninn status --help`.

`muninn.addrs` stays backward-compatible — an absent `scheme` is treated as `http`,
so old/new daemon and CLI interoperate. The `MUNINNDB_*_URL` env vars remain as an
explicit override (e.g. for remote admin).

Against a daemon that predates this change (no `scheme` in `muninn.addrs`), `muninn
status` still gets it right: the probe's http→https retry discovers the scheme, and
that discovered scheme is used for the `Web UI →` line too.

Verified on a live internal-CA TLS deployment: `muninn status` reports `[up]` with no
env vars set and prints `Web UI → https://…`; `muninn start` under TLS exits 0 (no
boot-loop); a daemon from this branch writes `"scheme":"https"` to `muninn.addrs`.

The broader "make TLS a first-class mode" design is tracked in #443.

## Release Checklist

- [x] ~~`CHANGELOG.md`~~ — maintainers update this at release time, no action needed
- [x] OpenAPI spec — N/A (no REST API routes changed)
- [x] SDK types — N/A (no request/response schemas changed; `muninn.addrs` is internal)
- [x] `docs/` updated — `MUNINNDB_*_URL` added to `docs/self-hosting.md`
- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test -race ./cmd/muninn/...` passes (full `./...` suite runs in CI)

### DIFF
--- a/cmd/muninn/help.go
+++ b/cmd/muninn/help.go
@@ -93,7 +93,10 @@ var subcommandHelp = map[string]func(){
 	},
 	"status": func() {
 		printSubcommandUsage("status", "show service health and ports", "muninn status", nil,
-			[]string{"muninn status"})
+			[]string{
+				"muninn status",
+				"MUNINNDB_ADMIN_URL=https://host:8475 muninn status   # TLS: override probe URLs",
+			})
 	},
 	"shell": func() {
 		printSubcommandUsage("shell", "interactive memory shell", "muninn shell", nil,
@@ -344,6 +347,9 @@ func printHelp() {
 	fmt.Println("  MCP endpoint: http://127.0.0.1:8750/mcp")
 	fmt.Printf("  %-28s %s\n", "MUNINN_MCP_URL", "Override MCP server URL (also used by 'muninn mcp' proxy)")
 	fmt.Printf("  %-28s %s\n", "MUNINNDB_DATA", "Override default data directory")
+	fmt.Printf("  %-28s %s\n", "MUNINNDB_ADMIN_URL", "TLS: base URL for 'muninn status'/admin REST probes")
+	fmt.Printf("  %-28s %s\n", "MUNINNDB_UI_URL", "TLS: base URL for 'muninn status' Web UI probe")
+	fmt.Printf("  %-28s %s\n", "MUNINNDB_MCP_URL", "TLS: base URL for 'muninn status'/'start' MCP probe")
 	fmt.Println()
 
 	fmt.Println(bold("PORTS"))
--- a/cmd/muninn/lifecycle.go
+++ b/cmd/muninn/lifecycle.go
@@ -2,9 +2,7 @@ package main
 
 import (
 	"fmt"
-	"io"
 	"net"
-	"net/http"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -157,20 +155,14 @@ func runStart(webEnabled bool) error {
 
 	// Wait for health check (up to 5s).
 	// Use the actual MCP port from daemon args — may differ from defaultMCPPort
-	// when the user passed --mcp-addr. Honours MUNINNDB_MCP_URL for TLS
-	// deployments (see #410 / #424 for the admin-CLI equivalents).
-	mcpHealthURL := healthURL("MUNINNDB_MCP_URL", mcpPortFromArgs(args)) + "/mcp/health"
+	// when the user passed --mcp-addr. probeHealth's http→https retry makes the
+	// poll succeed against a TLS deployment with no MUNINNDB_MCP_URL set, so
+	// `muninn start` no longer times out (and boot-loops) under TLS.
+	mcpHealthURL := healthURL("MUNINNDB_MCP_URL", "http", mcpPortFromArgs(args)) + "/mcp/health"
 	deadline := time.Now().Add(5 * time.Second)
 	for time.Now().Before(deadline) {
 		time.Sleep(200 * time.Millisecond)
-		resp, err := http.Get(mcpHealthURL)
-		if err != nil {
-			continue
-		}
-		// Always drain and close body to allow connection reuse and prevent leaks.
-		io.Copy(io.Discard, resp.Body)
-		resp.Body.Close()
-		if resp.StatusCode == 200 {
+		if up, _ := probeHealth(mcpHealthURL); up {
 			fmt.Printf("muninn started (pid %d)\n", cmd.Process.Pid)
 			fmt.Println()
 			printStatusDisplay(true)
--- a/cmd/muninn/pid.go
+++ b/cmd/muninn/pid.go
@@ -14,12 +14,25 @@ import (
 // the startup health poll can probe the correct ports when non-default
 // --*-addr flags are used.
 type daemonAddrs struct {
+	// Scheme is "http" or "https" — the scheme the daemon's client-facing
+	// listeners serve. An empty Scheme (e.g. a sidecar written by an older
+	// daemon that predates this field) is treated as "http" by readers.
+	Scheme   string `json:"scheme"`
 	RestAddr string `json:"rest_addr"`
 	MCPAddr  string `json:"mcp_addr"`
 	UIAddr   string `json:"ui_addr"`
 }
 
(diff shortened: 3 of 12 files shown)


scrypster:
Thanks for the thorough work on this — the PR description, the backward-compat story, and especially the test coverage are all excellent. This fixes a real pain point for TLS deployments and we're happy to merge it once one thing is addressed.

**One change requested — `cmdShowVaults` in `repl_client.go`**

The `InsecureSkipVerify: true` in `cmdShowVaults` sends the session cookie over an unverified TLS connection. Unlike `probeOnce` (credential-free liveness probe — justified), this path attaches a credential. The comment says "localhost admin REPL talking to its own server" but the code doesn't enforce that constraint — `apiURL` can come from `MUNINNDB_ADMIN_URL` pointing off-host.

This is also inconsistent with `loginAdmin` in `vault_auth.go`, which uses the default cert-verifying client for the same credential flow.

Suggested fix — gate the skip on a loopback URL check:

```go
if strings.HasPrefix(apiURL, "https://") && isLoopbackURL(apiURL) {
    client.Transport = &http.Transport{
        TLSClientConfig: &tls.Config{InsecureSkipVerify: true}, //nolint:gosec
    }
}
```

Or simply drop `InsecureSkipVerify` from `cmdShowVaults` entirely — consistent with `loginAdmin`, and operators running TLS with an internal CA can install the cert into the system trust store.

Everything else looks great. Address that and we'll merge.

johanneshauer:
Thanks for the careful review — good catch on the credential exposure.

Addressed in `54f7f9b`. I went with the **loopback-gate** option rather than dropping `InsecureSkipVerify` outright. Dropping it entirely would break `show vaults` in the local admin REPL on internal-CA TLS deployments unless the operator first installs the CA into the system trust store — which isn't currently a documented step. The gate keeps the zero-config local case working while closing the off-host hole.

Changes:
- New `isLoopbackURL()` in `status.go` — true only for empty host / `localhost` / a loopback IP; unparseable input returns `false`, so the caller fails closed (verification stays on).
- `cmdShowVaults` now skips cert verification only when `apiURL` is `https://` **and** loopback. An off-host `MUNINNDB_ADMIN_URL` verifies the certificate, consistent with `loginAdmin` in `vault_auth.go`.

One thing slightly beyond your request — happy to split it out if you'd prefer: now that verification is enforced for off-host URLs, an untrusted cert would surface through the old error path as a misleading "Is muninn running?". I added `isTLSCertError()` so a verification failure prints an actionable hint (install the CA / use a loopback address) instead. It's in the same commit — say the word and I'll move it to its own.

Tests: `TestIsLoopbackURL` (loopback v4/v6, `127/8`, localhost, off-host, malformed input) and `TestIsTLSCertError`. `go build` / `go vet` / `gofmt` clean; full `cmd/muninn` package green.

scrypster (approved):

PR #445, opened 2026-05-22
style: gofmt-align literals in repl_client_test.go

johanneshauer:
## What

`gofmt -w` on `cmd/muninn/repl_client_test.go`. The file has carried gofmt-misaligned composite literals since `1f1279a` (2026-02) — map keys and trailing comments in the `TestFormatVaultTable*` helpers were never column-aligned.

The change is **pure whitespace** — `git diff -w` produces an empty diff. No logic touched.

## Why it went unnoticed

CI currently has no `gofmt -l` / `golangci-lint` step, so formatting drift doesn't fail any check. The practical cost is diff noise: any contributor who runs `gofmt -w` (or whose editor formats on save) while editing this file drags these unrelated reformats into their PR.

Surfaced while working on #444 — kept out of that PR deliberately to avoid mixing a formatting cleanup into a security fix.

## Suggestion (optional)

Consider adding a `gofmt -l` gate to the CI workflow (fail if it lists any file) so this can't recur. Happy to open a separate PR for that if you'd like.

## Note

Touches the same file as #444, but in different regions (`TestFormatVaultTable*` here vs. a new test function there) — no overlap; whichever merges second needs no manual conflict resolution.

### DIFF
--- a/cmd/muninn/repl_client_test.go
+++ b/cmd/muninn/repl_client_test.go
@@ -205,10 +205,10 @@ func TestCmdShowVaultsEmptyList(t *testing.T) {
 func TestFormatVaultTableMissingFields(t *testing.T) {
 	// Vaults with missing optional fields should not panic
 	vaults := []map[string]any{
-		{"name": "vault-no-count"},                    // no memory_count
-		{"name": "vault-no-time", "memory_count": float64(5)},  // no last_active
-		{"memory_count": float64(3)},                  // no name
-		{},                                             // completely empty
+		{"name": "vault-no-count"},                            // no memory_count
+		{"name": "vault-no-time", "memory_count": float64(5)}, // no last_active
+		{"memory_count": float64(3)},                          // no name
+		{},                                                    // completely empty
 	}
 	out := captureStdout(func() {
 		formatVaultTable(vaults)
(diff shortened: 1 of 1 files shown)


scrypster (approved):
Pure gofmt cleanup — no logic changes. Verified  is empty. Happy to take the  CI gate PR when you're ready.

scrypster (approved):

johanneshauer:
Thanks for the standing invitation — picking this up now and hit a sizing question before opening the PR.

I ran `gofmt -l .` on `develop` tip (commit ad7908b): **144 files are dirty**, not just the one #445 fixed. PR #445's verification only confirmed `repl_client_test.go`. The formatting drift is repo-wide.

Three sequencing options for the CI gate — which would you prefer?

1. **Cleanup PR first, then gate PR.** One big pure-whitespace `gofmt -w` sweep across all 144 files (zero logic), then a second small PR adding the workflow. Cleanest endstate; two narrow review scopes.
2. **Single bundled PR.** Sweep + workflow together. Smaller round-trip but mixes a giant diff with the workflow change.
3. **Soft gate first.** Add the workflow with `continue-on-error: true` so it surfaces findings without blocking, chip away at the 144 files over time, then flip to hard-fail later. Never blocks your other PRs.

Also — scope: gofmt only, or extend to `go vet` (clean now) + a `golangci-lint` config? Lint will surface more drift.

Happy with whichever path you prefer.

PR #456, opened 2026-05-27
feat(tls): warn at startup when client-facing or cluster cert nears expiry (#443 gap 4)

johanneshauer:
## Summary

Addresses [issue #443](https://github.com/scrypster/muninndb/issues/443) gap 8 — "Cert read at boot, expiry never surfaced."

Adds `internal/tlsutil.CheckCertExpiry` and wires it into the two TLS cert-load paths so operators see expiry at startup instead of discovering it via a failed handshake.

### Behavior

| Cert state            | Log level | Existing flow |
|-----------------------|-----------|---------------|
| Expired               | Error     | Server still starts — failing loud, not shut |
| < 30 days to expiry   | Warn      | Server starts normally |
| ≥ 30 days             | (silent)  | Server starts; expiry fields added to existing Info log |

Threshold of 30 days matches Let's Encrypt's renewal window.

### Wiring

- `cmd/muninn/server.go` — client-facing TLS, user-provided cert path. Existing `tls: client-facing TLS enabled` Info log now also carries `subject`, `not_after`, `days_remaining`.
- `internal/replication/tls.go` — cluster node TLS, both:
  - user-provided cert path (line ~91)
  - auto-generated 1-year cert path (`generateNodeCert`, line ~268), where the check uses the cert template directly (no re-parse)

### Notes

- Parse failure on a just-loaded cert is treated as best-effort: a Warn is logged and startup continues. The cert load itself succeeded; blocking startup because the expiry banner can't be drawn would be wrong.
- `tlsutil.DaysRemaining` uses ceiling-up so a cert with 23h59m left reports `days_remaining=1`, not `0` (which reads as expired).
- The pre-existing gofmt drift on `internal/replication/tls.go:126` (CA template literal alignment) is intentionally untouched — separate from this PR's logic change, per the convention from #445. Will be swept up either by the CI-gate cleanup (see the open question on #445) or by future PRs that touch those lines.

### Out of scope

This PR closes gap 8 only. The remaining #443 gaps (1–7, 9) will land as separate small PRs.

### Tests

- `internal/tlsutil/expiry_test.go` — 5 test functions, including table-driven coverage for `DaysRemaining` boundaries.
- Existing `cmd/muninn` and `internal/replication` test suites pass unchanged; the replication suite exercises the new auto-gen check path via `generateNodeCert`.

### Verification

- `go build ./...` ✓
- `go vet ./...` ✓
- `gofmt -l` clean on all changed regions
- Full test suite green on touched packages

### DIFF
--- a/cmd/muninn/server.go
+++ b/cmd/muninn/server.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"crypto/tls"
+	"crypto/x509"
 	"encoding/json"
 	"flag"
 	"fmt"
@@ -42,6 +43,7 @@ import (
 	"github.com/scrypster/muninndb/internal/replication"
 	"github.com/scrypster/muninndb/internal/storage"
 	"github.com/scrypster/muninndb/internal/storage/migrate"
+	"github.com/scrypster/muninndb/internal/tlsutil"
 	grpcpkg "github.com/scrypster/muninndb/internal/transport/grpc"
 	"github.com/scrypster/muninndb/internal/transport/mbp"
 	"github.com/scrypster/muninndb/internal/transport/rest"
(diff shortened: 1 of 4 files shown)


johanneshauer:
Cross-referencing from a #443 review: there's an **init-side** complement to this daemon-startup expiry warning — `muninn init` accepts an expired (or not-yet-valid) TLS cert and prints the `✓ TLS enabled` / `You're live` banner with no warning, because `validateTLSPair` (via `tls.LoadX509KeyPair`) and `warnIfCertNotLoopback` (via `VerifyHostname`) both ignore the validity period.

Filed as #475 so an operator learns at config time, not just at the next daemon start. The two are complementary: #456 guards the serving daemon, #475 guards the init banner. Flagging here so the validity-window check can be shared/kept consistent between the two paths.

scrypster (approved):
Verified: additive observability only — cert expiry surfaced at startup, no behavior change, fail-loud-not-shut semantics are right. Tests included, all checks green.

PR #461, opened 2026-06-02
fix(ui): normalize created_at unit so memory detail panel shows a valid date

### LINKED ISSUE #460: Dashboard: memory detail panel shows "Created: Invalid Date" for search results
# Dashboard: memory detail panel shows "Created: Invalid Date" for search results

## Summary

In the dashboard Web UI (port 8476), opening a memory's detail panel can show
`Created: Invalid Date` instead of the creation timestamp. It reproduces
reliably for memories opened from **search** results, and for **session
timeline** entries that are expanded to their full engram.

## Steps to reproduce

1. Open the dashboard → **Memories** tab.
2. Run a search (any mode) so the list is populated from search.
3. Click a result to open the detail panel.
4. The **Created** field reads `Invalid Date`.

Memories opened from the plain **Browse All** list render the date correctly,
which is the clue to the root cause.

## Root cause

The REST API is **not consistent** about the unit of the `created_at` integer
it returns, and the frontend assumes a single unit everywhere.

Backend serialization units by endpoint:

| Endpoint | Frontend use | Serialized as | Unit |
|---|---|---|---|
| `GET /api/engrams` (list) | Browse list | `time.Time.Unix()` | seconds |
| `POST /api/activate` (search) | Search results | `time.Time.UnixNano()` | nanoseconds |
| `GET /api/engrams/{id}` (read) | Open full engram | `time.Time.UnixNano()` | nanoseconds |
| `GET /api/session` | Session timeline | `time.Time.Unix()` | seconds |

The frontend formats every `created_at` the same way:

```js
new Date(selectedMemory.createdAt * 1000).toLocaleString()
```

- For **seconds** (list, session): `seconds * 1000` = milliseconds → correct.
- For **nanoseconds** (search, single read): `nanoseconds * 1000` is ~10^21,
  far outside JS `Date`'s valid range → `Invalid Date`.

So the visible bug is a frontend display defect, but the underlying defect is
that the REST JSON contract uses two different units for the same field name.

## Relevant code

- Frontend formatting: `web/templates/index.html` (the "Created" field in the
  detail panel) — multiplies `createdAt` by 1000.
- Frontend mapping sites: `web/static/js/app.js` — list, search (`/api/activate`),
  open-full (`/api/engrams/{id}`), and session mappings each copy `created_at`
  straight through without normalizing the unit.
- Backend units: `internal/transport/rest/engine_adapter.go` (`.Unix()` for
  list/session) vs `internal/engine/engine.go` (`.UnixNano()` for activate and
  read responses).

## Proposed fix

Normalize the `created_at` unit so the frontend renders a single, well-defined
value. The display fix lives in the frontend (it already knows which endpoint
each value came from): convert each source to epoch **milliseconds** at the
mapping site and render with `new Date(ms)`.

Optionally, follow up by making the REST JSON contract self-consistent (pick one
unit for `created_at` across all REST endpoints) so future API consumers don't
hit the same trap. That change is broader because the activate/read response
types are shared with other transports, so it is tracked separately.

## Acceptance criteria

- The detail panel shows a valid, correct local timestamp for memories opened
  from search, browse, and session views.
- No `Invalid Date` for any valid stored memory.
- A regression test covers unit normalization for the affected paths.

timharsch:
# fix(ui): normalize created_at unit so memory detail panel shows a valid date

Closes #460

## Problem

The dashboard memory detail panel showed `Created: Invalid Date` for memories
opened from search results (and for expanded session entries). Memories opened
from the plain browse list rendered correctly.

## Root cause

The REST API returns the `created_at` integer in **inconsistent units**, while
the frontend assumed a single unit everywhere:

| Endpoint | Used by | Serialized as | Unit |
|---|---|---|---|
| `GET /api/engrams` | browse list | `time.Time.Unix()` | seconds |
| `GET /api/session` | session timeline | `time.Time.Unix()` | seconds |
| `POST /api/activate` | search | `time.Time.UnixNano()` | nanoseconds |
| `GET /api/engrams/{id}` | open full engram | `time.Time.UnixNano()` | nanoseconds |

The frontend rendered every value as `new Date(createdAt * 1000)`:

- seconds × 1000 → milliseconds → correct
- nanoseconds × 1000 → ~10^21, outside JS `Date`'s valid range → `Invalid Date`

## Fix

Frontend-only, lowest blast radius (the activate/read response types are shared
with other transports, so the backend types are deliberately untouched):

- New `web/static/js/time-utils.js` with `secondsToMillis` / `nanosToMillis`,
  exposed via `globalThis.MuninnTime`, mirroring the existing
  `plugin-config-utils.js` pattern.
- Each `created_at` mapping site now normalizes to epoch **milliseconds** at
  ingest, choosing the converter that matches that endpoint's unit (browse +
  session = seconds; search + single-read = nanoseconds; the `memory_added`
  websocket feed = seconds).
- Every template render site drops the `* 1000` and renders `new Date(value)`.

The invariant is now "frontend `createdAt` is always epoch-milliseconds", applied
once at ingest and never re-scaled at render — no double-conversion.

## Tests

- New `web/time-utils.test.js` (vitest): seconds input, nanoseconds input
  (including sub-millisecond flooring), zero/null/undefined, NaN, negative
  inputs, the `|| fallback` chain semantics, and a valid-`Date` assertion for
  real inputs. 16 tests.
- Unit suite green: `npx vitest run plugin-config-utils.test.js time-utils.test.js`
  → 33 passed.

## Notes / out of scope

- The deeper asymmetry is in the backend (`.Unix()` vs `.UnixNano()` across
  endpoints). A follow-up could make the REST JSON contract emit a single unit
  so the frontend can drop the dual-converter logic; that touches response types
  shared with other transports and is tracked separately.
- `npm test` (bare `vitest run`) currently also tries to collect the Playwright
  `e2e/*.spec.ts` files and errors at collection — a pre-existing vite config
  gap unrelated to this change (run e2e via the Playwright script). Not modified
  here.

### DIFF
--- a/web/static/js/app.js
+++ b/web/static/js/app.js
@@ -616,7 +616,10 @@ document.addEventListener('alpine:init', () => {
         // ends of the array confuse Alpine's DOM anchor tracking and produce the
         // "can't access property 'after', v is undefined" crash.
         if (!this.liveFeed.some(item => item.id === msg.data.id)) {
-          const next = [msg.data, ...this.liveFeed];
+          // The memory_added event carries createdAt in SECONDS (backend
+          // .Unix()) → normalize to ms so the frontend convention holds.
+          const item = { ...msg.data, createdAt: this.secondsToMillis(msg.data.createdAt) };
+          const next = [item, ...this.liveFeed];
           this.liveFeed = next.length > 20 ? next.slice(0, 20) : next;
         }
       }
@@ -634,6 +637,24 @@ document.addEventListener('alpine:init', () => {
         throw new Error(res.status + ': ' + text);
       }
       return res.json();
+    },
+
+    // ── created_at unit normalization ────────────────────────────────────────
+    // The REST API returns created_at in INCONSISTENT units depending on the
+    // endpoint (the backend types are shared with the MCP/gRPC transports, so
+    // they cannot be changed here):
+    //   • /api/engrams and /api/session use .Unix()      → SECONDS
+    //   • /api/activate and /api/engrams/{id} use UnixNano() → NANOSECONDS
+    // To keep the frontend consistent, every mapping site normalizes created_at
+    // to epoch MILLISECONDS via these helpers, and templates render the value
+    // directly with `new Date(createdAt)` (no `* 1000`). The pure converters
+    // live in time-utils.js (exposed as globalThis.MuninnTime) so they can be
+    // unit-tested under Vitest.
+    secondsToMillis(value) {
+      return MuninnTime.secondsToMillis(value);
+    },
+    nanosToMillis(value) {
+      return MuninnTime.nanosToMillis(value);
     },
 
     // ── Dashboard ──────────────────────────────────────────────────────────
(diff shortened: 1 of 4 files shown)


scrypster (approved):
Verified: root cause confirmed (mixed seconds/nanoseconds across endpoints), normalization applied once at ingest, no remaining *1000 render sites (grepped the branch), 16 vitest unit tests. All checks green except the stale pre-#464 govulncheck.

PR #463, opened 2026-06-09
feat(cli): add 'muninn doctor' to self-describe TLS state (#443 gap 3)

johanneshauer:
## What

Adds **`muninn doctor`** — a TLS/connectivity self-describe command implementing **gap 3 of #443** ("the server cannot self-describe its TLS state"). Operators currently reverse-engineer "is TLS on? which cert? bound where? when does it expire?" from env vars and unit files.

```
  muninn doctor

  server     ●  running
  TLS        ●  enabled (https)
  bind       database  0.0.0.0:8475
             mcp       0.0.0.0:8750
             web ui    0.0.0.0:8476

  certificate  (from live socket)
    subject    CN=muninn.example
    issuer     CN=Example Root CA
    dns sans   muninn.example, muninn
    valid      2026-05-01 → 2027-05-01 (UTC)
    expiry     ●  326 days remaining
```

`-v` adds IP SANs, serial, signature algorithm, the negotiated TLS version/cipher, and the certificate chain.

## How

- **Cert source priority:** (a) the **live TLS socket** — ground truth, what clients actually receive, dialed on loopback since the recorded bind host may be `0.0.0.0` (or absent under systemd, in which case it falls back to the default REST port); (b) the **`MUNINN_TLS_CERT` file** — cert-only PEM parse (never the private key), so a *stopped* server's configured cert can still be inspected offline.
- **Reuses `internal/tlsutil`** (`DaysRemaining`, `ExpiryWarnWindow`) for expiry classification — deliberately **not** `CheckCertExpiry`, which logs via slog and would corrupt the formatted stdout.
- Factors two shared helpers out of `status.go` (`resolveScheme`, `portFromAddr`) now used by **both** `status` and `doctor`, removing a verbatim scheme-recovery duplication.
- Pure gather/format split (`gatherDoctor` does the only I/O; `formatDoctor` is pure) with an injectable dial seam — see tests.

## Tests

`cmd/muninn/doctor_test.go`: formatter matrix (TLS off / healthy / expiring / expired / unobtainable / verbose / non-TTY ASCII degradation), `gatherDoctor` with stubbed probe+dial (asserts the `0.0.0.0`→loopback rewrite, the default-port fallback when the sidecar is absent, and that `MUNINN_TLS_CERT` errors surface even with the server stopped), PEM leaf-first chain selection, and a live smoke test against `httptest.NewTLSServer`. `gofmt`/`vet` clean; verified manually against a live TLS server.

## ⚠️ Stacked on #456

This branches off `feat/443-cert-expiry-warn` for the `internal/tlsutil` dependency, so this PR's diff currently **includes #456's commit** (`feat(tls): warn at startup when cert nears expiry`). **Review only the doctor commits**; rebase onto `develop` once #456 merges.

### DIFF
--- a/cmd/muninn/doctor.go
+++ b/cmd/muninn/doctor.go
@@ -1,0 +1,300 @@
+package main
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"errors"
+	"flag"
+	"fmt"
+	"net"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/scrypster/muninndb/internal/tlsutil"
+)
+
+// doctorReport is the gathered, render-free view of the daemon's TLS and
+// connectivity state. Populated by gatherDoctor (the only I/O); rendered by the
+// pure formatDoctor.
+type doctorReport struct {
+	svcs       []serviceStatus
+	scheme     string              // "https" | "http" | "" (unknown / server stopped)
+	addrs      daemonAddrs         // real bound addresses — displayed verbatim
+	cert       *x509.Certificate   // leaf; nil when TLS off or cert unobtainable
+	chain      []*x509.Certificate // additional chain certs (cert-file path only)
+	certSource string              // "live socket" | "cert file" | ""
+	tlsVersion uint16              // negotiated version (live path only; 0 otherwise)
+	cipher     uint16              // negotiated cipher (live path only; 0 otherwise)
+	certErr    string              // why the cert could not be inspected, if applicable
+}
+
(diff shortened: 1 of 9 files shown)


scrypster (approved):
Verified: 'muninn doctor' is cmd/muninn-only (the internal/tlsutil content in the file list is #456's commit, byte-identical to develop). Pure gather/format split with injectable dial seam, 332 lines of tests; the InsecureSkipVerify in dialServedCert is inspect-not-trust with the dial pinned to 127.0.0.1; cert-file fallback parses only CERTIFICATE blocks. Vuln-scan red is the stale pre-#464 advisory.

PR #464, opened 2026-06-09
chore: bump go directive to 1.26.4 to clear govulncheck stdlib advisories

johanneshauer:
## Problem

The **Vulnerability scan** CI job (`govulncheck ./...`) is failing on `develop` and on every open PR — `govulncheck@latest` now reports two Go **standard-library** advisories, both reachable from existing code:

| Advisory | Package | Fixed in | Reachable from |
|---|---|---|---|
| [GO-2026-5037](https://pkg.go.dev/vuln/GO-2026-5037) | `crypto/x509` (inefficient candidate hostname parsing) | go1.26.4 | `internal/mcp/server.go`, `internal/transport/grpc/engine_adapter.go` |
| [GO-2026-5039](https://pkg.go.dev/vuln/GO-2026-5039) | `net/textproto` (arbitrary inputs unescaped in errors) | go1.26.4 | `internal/mcp/server.go` |

`go.mod` pins `go 1.26.3`, and CI's `setup-go` uses `go-version-file: go.mod`, so the scan builds against the vulnerable stdlib.

## Fix

Bump the `go` directive `1.26.3 → 1.26.4`. CI then builds against the patched standard library. No code changes; no new `toolchain` directive.

## Verification

```
$ govulncheck ./...   # before: go.mod @ 1.26.3
Your code is affected by 2 vulnerabilities from the Go standard library.  (exit 3)

$ govulncheck ./...   # after: go.mod @ 1.26.4
No vulnerabilities found.   (exit 0)
```

`go build ./...` succeeds under go1.26.4.

PR #465, opened 2026-06-09
feat(cli): TLS setup in 'muninn init' + scheme-aware client URLs (#443 gaps 1 & 5)

johanneshauer:
## What

Implements **gaps 1 & 5 of #443** in the `muninn init` wizard:

- **Gap 1 — no TLS setup path.** `init` now has `--tls-cert`/`--tls-key` flags and an interactive *"Serve over TLS?"* prompt. Bring-your-own-cert — the wizard does **not** generate certificates (operator's job, per the issue).
- **Gap 5 — client onboarding over TLS is unowned.** `init` hardcoded `http://127.0.0.1:8750/mcp` into every generated AI-tool config, so a TLS deployment's clients got an unusable `http://` config. The URL is now scheme-aware.

## How

- **Validate up front:** the pair is checked with `tls.LoadX509KeyPair` before the daemon starts, so a bad cert gives a clear CLI error instead of a 5s health-probe-timeout hang (the daemon `os.Exit(1)`s on a bad cert). An explicit-but-invalid pair is **fatal** in both the flag and non-interactive paths — an explicit TLS request is never silently downgraded to plaintext.
- **Reach the daemon:** `init` sets `MUNINN_TLS_CERT/_KEY` in its env before `runStart` forks the daemon (which inherits the env and enables TLS via the existing fallback) — no `buildDaemonArgs` change, keeping cert paths out of `ps`, consistent with how the bearer token is handled.
- **Persist:** the cert/key are written into `~/.muninn/muninn.env` as **active** lines (new `upsertEnvFileVar`, which flips the commented template lines — handles commented/active/`export` forms, prefix-collision safe) so future `muninn start`/systemd restarts keep https.
- **Scheme-aware URLs:** `clientMCPURL()` returns `MUNINN_MCP_URL` (operator-advertised/routable) if set, else `<scheme>://127.0.0.1:8750/mcp` with the scheme following the TLS env. The success-banner MCP/UI URLs are scheme-aware too. **With no TLS the output is byte-identical to before** — no behaviour change for existing setups.

`envfile.go` also factors a shared `atomicWriteFile` primitive out of `writeEnvFileTo`.

## Follow-up in this PR (commit 54ee39c) — scheme-aware generated configs + admin calls under TLS

Closes the remaining init-under-TLS seams so the configs the wizard writes *and* the admin calls it makes are correct:

- **OpenClaw `SKILL.md` REST URL** is now scheme-aware (new `clientRESTURL()`), and the `muninn.env` `MUNINN_MCP_URL` template reflects the scheme + `defaultMCPPort`.
- **`alignLocalAdminBasesToDaemon()`** — after `runStart`, the wizard points its *own* admin/UI base URLs at the daemon's actual scheme + port read from the `muninn.addrs` sidecar (the ground truth `cmdShowVaults`/`doctor` already use), so `loginAdmin` and the plasticity `PUT` reach a daemon just started with TLS (or on a non-default port) instead of failing against `http://127.0.0.1:8475`. `MUNINNDB_ADMIN_URL`/`MUNINNDB_UI_URL` overrides are respected.
- **Loopback cert handling:** `loginAdmin` and the plasticity client skip verification *only* for a loopback `https` target (same gate as `cmdShowVaults`); off-host stays fully verified. These three sites unify onto `httpClientForURL` once #468 lands.

Without this, `muninn init --tls` set up TLS but then silently failed to persist the default vault's behaviour mode (login `http→https`) and emitted `http://` in the generated SKILL.md / env template. **No behaviour change for non-TLS setups** (the `http` output is byte-identical).

## Tests

`cmd/muninn/init_tls_test.go`: `clientMCPURL`/`clientUIURL` (default http, https from env, `MUNINN_MCP_URL` override + trim, cert-without-key stays http), `validateTLSPair` (both/one/neither + invalid), and `upsertEnvFileVar` (creates-when-absent, activates a commented template line, updates an active line, recognizes a commented `export` line without duplicating, prefix-collision leaves a longer-named key untouched). Verified end-to-end against a throwaway self-signed cert in a sandboxed `HOME`: TLS run writes `https://…` into the config + active `muninn.env` lines; no-TLS run writes `http://`. `gofmt`/`vet` clean.

## Scope / notes

- Independent of #456/#463 — reuses only phase-1 infra (the `muninn.addrs` scheme field, `schemeFor`).
- Out of scope (separate #443 gaps): the server-side `/api/admin/mcp-info` http construction and `docs/tls.md`. The OpenClaw `SKILL.md` REST URL and `muninn.env` `MUNINN_MCP_URL` template are now handled (commit 54ee39c). The wizard's *own* admin/UI calls now source scheme + port from `muninn.addrs`; the generated *client* config URLs (`clientMCPURL`/`clientUIURL`/`clientRESTURL`) still hardcode the default ports, so honouring a non-default `--rest-addr`/`--ui-addr` in those remains a small follow-up.
- The repo-wide **Vulnerability scan** CI job is currently red on every branch (a go1.26.3 stdlib advisory); see #464, which bumps the toolchain. Not related to this change.

### DIFF
--- a/cmd/muninn/coverage_hardening_test.go
+++ b/cmd/muninn/coverage_hardening_test.go
@@ -786,7 +786,7 @@ func TestRunNonInteractiveInit(t *testing.T) {
 	t.Setenv("MUNINNDB_DATA", filepath.Join(home, "data"))
 
 	out := captureStdout(func() {
-		runNonInteractiveInit("http://127.0.0.1:8750/mcp", "manual", "", false, true, true)
+		runNonInteractiveInit("manual", "", false, true, true, "", "")
 	})
 	if !strings.Contains(out, "Manual") && !strings.Contains(out, "manual") &&
 		!strings.Contains(out, "mcpServers") && !strings.Contains(out, "curl") {
@@ -1417,7 +1417,7 @@ func TestRunNonInteractiveInit_WithTools(t *testing.T) {
 	t.Setenv("MUNINNDB_DATA", filepath.Join(home, "data"))
 
 	out := captureStdout(func() {
-		runNonInteractiveInit("http://127.0.0.1:8750/mcp", "vscode,manual", "", true, true, true)
+		runNonInteractiveInit("vscode,manual", "", true, true, true, "", "")
 	})
 	_ = out
 }
@@ -1429,7 +1429,7 @@ func TestRunNonInteractiveInit_WithToken(t *testing.T) {
 	t.Setenv("MUNINNDB_DATA", filepath.Join(home, "data"))
 
 	out := captureStdout(func() {
-		runNonInteractiveInit("http://127.0.0.1:8750/mcp", "", "mdb_customtoken", false, true, true)
+		runNonInteractiveInit("", "mdb_customtoken", false, true, true, "", "")
 	})
 	if !strings.Contains(out, "Token") || !strings.Contains(out, "mcp.token") {
 		t.Logf("output: %s", out)
@@ -1444,7 +1444,7 @@ func TestRunNonInteractiveInit_GenerateToken(t *testing.T) {
 	t.Setenv("MUNINNDB_DATA", filepath.Join(home, ".muninn", "data"))
 
 	out := captureStdout(func() {
-		runNonInteractiveInit("http://127.0.0.1:8750/mcp", "", "", false, true, true)
+		runNonInteractiveInit("", "", false, true, true, "", "")
 	})
 	_ = out
 }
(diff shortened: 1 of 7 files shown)


scrypster (approved):
Deep-reviewed: TLS flags + interactive prompt in init, upfront keypair validation, TLS handed to the forked daemon via env (cert paths kept out of ps), atomic env-file persistence, scheme-aware generated configs with byte-identical non-TLS output. InsecureSkipVerify strictly loopback-gated, matching the httpClientForURL precedent — and the merge resolution adopted develop's httpClientForURL unification this PR's own comment anticipated. 546-line test file covers the new paths; CI fully green post-merge with develop.

PR #466, opened 2026-06-09
docs(tls): TLS guide + TLS-aware systemd unit (#443 gap 2)

johanneshauer:
## What

Implements **gap 2 of #443**: "No docs — `docs/` has self-hosting.md, auth.md, quickstart.md but no `tls.md`; `contrib/muninndb.service` is not TLS-aware and nothing flags that."

- **New `docs/tls.md`** — an operator guide for **native** client-facing TLS: what TLS protects, enabling it, obtaining a certificate, verifying, connecting clients, systemd, a reverse-proxy alternative, and troubleshooting.
- **`contrib/muninndb.service`** — gains a commented `# --- TLS (optional) ---` block.
- **`docs/index.md`** links the new guide; **`docs/auth.md`** and **`docs/self-hosting.md`** are reworded so their "run behind a TLS-terminating proxy" advice no longer contradicts native TLS.

## Accuracy notes (verified against `develop`, runtime-tested)

- **TLS is configured via `MUNINN_TLS_CERT`/`MUNINN_TLS_KEY` env**, not `muninn start --tls-cert` — `buildDaemonArgs` does not forward those flags to the forked daemon, so the guide leads with the env vars (and systemd `Environment=`).
- **All client-facing listeners** (REST `:8475`, MCP `:8750`, Web UI `:8476`, MBP `:8474`, gRPC `:8477`) share the one cert; the **metrics** endpoint is plain HTTP/no-auth and must be loopback-bound — called out in a `> **Note:**`.
- **systemd `$HOME` trap:** the shipped unit has no `User=`, so `~/.muninn/muninn.env` does not load (`os.UserHomeDir()` errors); the doc and unit both state TLS must go in `Environment=`/`EnvironmentFile=`, and that the two file formats are not interchangeable.
- The self-signed `openssl` one-liner **sets a SAN** (modern clients reject SAN-less certs); the verification steps were run end-to-end against a real TLS daemon in a sandbox (`muninn status` auto-detects https, `openssl s_client`, `curl -k`). The guide is explicit that OpenSSL clients reject a self-signed cert even via `--cacert` (use a CA-issued cert, or `-k` for local testing).

## Scope

Documents only what's on `develop` — **no** references to the unmerged init TLS wizard (#465), `muninn doctor` (#463), or cert-expiry (#456); a follow-up will document those once they land. Cross-links self-hosting/auth/cluster/Traefik rather than duplicating them.

> The repo-wide **Vulnerability scan** CI job is currently red on every branch (a go1.26.3 stdlib advisory); see #464. Unrelated to this docs change.

### DIFF
--- a/contrib/muninndb.service
+++ b/contrib/muninndb.service
@@ -59,6 +59,15 @@ Environment=MUNINN_ANTHROPIC_KEY=sk-ant-...
 # Environment=MUNINN_ENRICH_URL=openai://gpt-4o-mini
 # Environment=MUNINN_ENRICH_API_KEY=sk-...
 
+# --- TLS (optional) ---
+# Serve clients over https. Set BOTH or neither. See docs/tls.md.
+# NOTE: ~/.muninn/muninn.env is NOT loaded under systemd (no $HOME), so TLS
+# for a systemd deployment MUST be set here via Environment= (or EnvironmentFile=).
+# Environment=MUNINN_TLS_CERT=/etc/muninndb/cert.pem
+# Environment=MUNINN_TLS_KEY=/etc/muninndb/key.pem
+# Alternatively load from a separate file (systemd EnvironmentFile format is
+# KEY=VALUE, NOT muninn.env format): EnvironmentFile=-/etc/muninndb/muninndb.env
+
 # --- Auth ---
 # The MCP bearer token is passed via --mcp-token flag.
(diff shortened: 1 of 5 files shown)


scrypster (approved):
Docs + commented systemd block only; verified accurate against develop (TLS is env-var-only in buildDaemonArgs, metrics plain-HTTP caveat real). Vuln-scan red is the stale pre-#464 stdlib advisory.

PR #467, opened 2026-06-09
feat(cli): derive Web UI URL host from the cert DNS SAN (#443 gap 6)

johanneshauer:
## What

Implements **gap 6 of #443**. When the Web UI is bound to a routable address (`0.0.0.0`/LAN) under TLS, `webUIDisplay` printed `https://<os.Hostname()>:8476` — a host the certificate is unlikely to certify, so the printed URL **fails TLS verification**. It now uses a name the cert actually covers.

## How

- The **daemon** (which has the cert) records the cert's **first routable, non-wildcard DNS SAN** into the `muninn.addrs` sidecar via a new `cert_host` field (`certRoutableHost` in pid.go, next to `schemeFor`). `muninn status` stays a cheap file read — no new TLS dial.
- `webUIDisplay` prefers `addrs.CertHost` in the routable branch under https, so the printed URL passes verification. Precedence is unchanged elsewhere: `MUNINNDB_UI_URL` env override still wins outright; loopback binds and http are unaffected; it falls back to `os.Hostname()` when there's no usable SAN.
- **Wildcards are skipped** (`*.example.com` would make a broken URL); single-label SANs are kept (the cert certifies them, so the URL verifies). `localhost` is skipped via the existing `hostIsRoutable`.
- `certRoutableHost` is **best-effort** — any error (no TLS, unreadable/parse failure, one-of-pair) yields `""` and the display falls back to `os.Hostname()`. The authoritative cert load + validation (`os.Exit(1)` on a bad cert) at startup is untouched; the sidecar write stays where it is (a deliberate best-effort second parse, commented).

## Notes

- The new `cert_host` field is JSON-additive and backward-compatible: old sidecars read with it empty (Go ignores unknown fields; no `DisallowUnknownFields` in the tree), so an old daemon simply falls back to today's `os.Hostname()` behavior.
- **Value-add framing:** operators could already force a verifying URL with `MUNINNDB_UI_URL`; this makes the **default** (no env var) print one automatically.
- **Trade-off:** the printed `https://<SAN>:port` is what *verifies* — the operator is responsible for making `<SAN>` resolve (DNS/hosts), vs today's resolvable-but-maybe-unverifiable hostname. That's the intended #443 behavior ("guarantee the printed URL passes TLS verification").
- **Staleness:** `cert_host` is recorded at startup; a cert rotated in place without a restart leaves it stale (documented on the field). Display-only; self-heals on `muninn restart`.

## Tests

`cmd/muninn/webui_san_test.go`: `certRoutableHost` over generated SAN certs (routable SAN; localhost-skip; wildcard-skip; wildcard-then-hit; only-localhost; IP-only; empty/one-of-pair/nonexistent paths); the `muninn.addrs` round-trip with `cert_host` + an old-format sidecar; and `webUIDisplay` (uses the SAN under https-routable; falls back to `os.Hostname()` without a SAN; ignores `CertHost` for http and for loopback binds; `MUNINNDB_UI_URL` wins). Verified end-to-end in a sandbox: a `0.0.0.0`-bound TLS daemon with a `DNS:muninn.example` cert → `Web UI → https://muninn.example:<port>`. `gofmt`/`vet` clean.

> The repo-wide **Vulnerability scan** CI is currently red on every branch (a go1.26.3 stdlib advisory); see #464. Unrelated to this change.

### DIFF
--- a/cmd/muninn/pid.go
+++ b/cmd/muninn/pid.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"crypto/tls"
+	"crypto/x509"
 	"encoding/json"
 	"fmt"
 	"os"
@@ -21,6 +23,12 @@ type daemonAddrs struct {
 	RestAddr string `json:"rest_addr"`
 	MCPAddr  string `json:"mcp_addr"`
 	UIAddr   string `json:"ui_addr"`
+	// CertHost is a routable DNS SAN from the client-facing TLS certificate, if
+	// any — so 'muninn status' can print a Web UI URL that passes TLS
+	// verification instead of an os.Hostname() that the cert may not cover.
+	// Empty for http, for an older daemon, or for a cert with no routable DNS SAN.
+	// Recorded at startup; a cert rotated in place without a restart leaves it stale.
+	CertHost string `json:"cert_host"`
 }
 
 const addrsFileName = "muninn.addrs"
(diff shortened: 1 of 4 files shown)


scrypster (approved):
Verified: best-effort, display-only derivation of the Web UI host from the cert's first routable non-wildcard DNS SAN; JSON-additive cert_host field, env override precedence tested, rotation staleness documented. Keep-both pid.go conflict with #468 resolved; full cmd/muninn suite green locally and in CI.

PR #468, opened 2026-06-09
feat(cli): reflect the TLS scheme in internal & printed URLs (#443 follow-ups)

johanneshauer:
## What

Finishes the last *functional* rows of #443's gap-map — internal/printed URLs that hardcoded `http://` and broke or misled under TLS. (The numbered design gaps 1–6 are already shipped.)

- **Server-side MCP URL** — `/api/admin/mcp-info` now returns `https://…/mcp` when the daemon serves TLS.
- **`muninn mcp` stdio proxy** — defaulted to `http://`, so it spoke plaintext to a local **TLS** daemon (a real bug). Now scheme-aware.
- **`muninn cluster --addr` default** — defaulted to `http://127.0.0.1:8475`; now reflects the local daemon's scheme.
- Plus the `muninn help` MCP-endpoint line and the `muninn backup` curl hint.

## How

- **Server:** `handleMCPInfo` derives the scheme from `s.tlsConfig != nil` — the exact condition `Serve` wraps the listener with, and more reliable than `r.TLS` (which is `nil` when this endpoint is reached over loopback behind the UI reverse-proxy). The `MUNINN_MCP_URL` override is unchanged.
- **CLI:** a new `localScheme()` reads the scheme the daemon recorded in the `muninn.addrs` sidecar (the same source `muninn status` uses); `muninn mcp` and `muninn cluster` default to it.
- **The crux — a loopback-TLS-aware client.** Flipping a default to `https://` is wrong unless the client trusts a loopback self-signed/internal-CA cert. New `httpClientForURL(rawURL, timeout)` skips verification **only for loopback https** (`isLoopbackURL`), keeping remote https fully verified — generalizing the pattern already inline in `probeOnce`/`repl_client` (and `repl_client` is migrated to it). This also fixes a *pre-existing* latent bug: the cluster/mcp clients couldn't reach a loopback https daemon even with an explicit https URL.
- Adds the `defaultRESTPort` const (used at the touched sites; not a repo-wide literal sweep).

**Non-TLS behaviour is byte-identical**, and `MUNINN_MCP_URL` / an explicit `--addr` still override everything.

## Security

`InsecureSkipVerify` is scoped to **loopback only** (`isLoopbackURL`, which uses `u.Hostname()` — userinfo tricks like `https://127.0.0.1@evil.com/` resolve to `evil.com` and stay verified). The credential-bearing callers (MCP bearer token, admin session cookie) never send creds to a *remote* host with verification disabled — remote endpoints aren't loopback, so verification stays on.

## Tests

`cmd/muninn/scheme_urls_test.go`: `localScheme`/`defaultMCPProxyURL`/`clusterAddrDefault` (https sidecar ⇒ https, absent ⇒ http — each `t.Setenv(MUNINNDB_DATA)`), and `httpClientForURL` (loopback https ⇒ skip-verify transport; remote https ⇒ verified; http ⇒ verified). New `TestMCPInfo_TLS` (https when `tlsConfig != nil`; the 3 existing mcp-info tests stay http/green). Verified end-to-end in a sandbox: a TLS daemon makes `muninn cluster info -h` show an `https://` default and the `help` line `https://…/mcp`; a no-TLS sandbox stays `http://`. `gofmt`/`vet` clean.

## Scope

Out of scope (logged as follow-ups): `muninn init`-generated `SKILL.md`/`muninn.env` URLs (moot on `develop` — init has no TLS branch yet, that's the unmerged #465); the full `8475`/`8476` literal sweep; and ~30 other admin/vault CLI `http.Client`s (`vault.go`, `api-key.go`, `admin.go`) that have the same latent loopback-TLS gap `httpClientForURL` now solves — a clean follow-up.

> The repo-wide **Vulnerability scan** CI is currently red on every branch (a go1.26.3 stdlib advisory); see #464. Unrelated to this change.

### DIFF
--- a/cmd/muninn/backup.go
+++ b/cmd/muninn/backup.go
@@ -82,7 +82,7 @@ func runBackup(args []string) {
 		fmt.Fprintf(os.Stderr, "error: muninn is running (pid %d) — cannot perform offline backup\n", pid)
 		fmt.Fprintln(os.Stderr, "")
 		fmt.Fprintln(os.Stderr, "Use the online backup endpoint instead:")
-		fmt.Fprintf(os.Stderr, "  curl -X POST http://127.0.0.1:8475/api/admin/backup -d '{\"output_dir\": \"%s\"}'\n", outputDir)
+		fmt.Fprintf(os.Stderr, "  curl -X POST %s://127.0.0.1:%s/api/admin/backup -d '{\"output_dir\": \"%s\"}'\n", localScheme(), defaultRESTPort, outputDir)
 		osExit(1)
 		return
 	}
--- a/cmd/muninn/cluster.go
+++ b/cmd/muninn/cluster.go
@@ -57,14 +57,14 @@ func printClusterHelp() {
 	fmt.Println("  disable       Disable cluster mode on the running node")
 	fmt.Println()
 	fmt.Println("Flags:")
-	fmt.Println("  --addr <url>   Server address (default: http://127.0.0.1:8475)")
+	fmt.Printf("  --addr <url>   Server address (default: %s)\n", clusterAddrDefault())
 	fmt.Println("  --json         Output raw JSON")
 }
 
 // runClusterInfo displays cluster topology and node status.
 func runClusterInfo(args []string) {
 	fs := flag.NewFlagSet("cluster info", flag.ContinueOnError)
-	addr := fs.String("addr", "http://127.0.0.1:8475", "Server address")
+	addr := fs.String("addr", clusterAddrDefault(), "Server address")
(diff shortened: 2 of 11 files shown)


scrypster (approved):
Reviewed in depth: scheme-aware URLs with InsecureSkipVerify strictly gated to loopback (userinfo-trick safe, fail-closed on unparseable URLs), CheckRedirect refuses off-loopback redirects and re-imposes the 10-hop cap, tests pin all of it. Transport-value change only, no contract change. Noted trade-off: loopback trust is skip-verify, not CA-anchored — acceptable for a localhost CLI boundary, future hardening welcome. Vuln-scan red is the stale pre-#464 advisory.

PR #469, opened 2026-06-09
fix(cli): route admin/vault HTTP clients through httpClientForURL for loopback TLS (#443)

johanneshauer:
## What

The admin/vault CLI clients in `cmd/muninn` built a plain `&http.Client{}` (or used the bare `http.Get`/`http.Post` helpers), so they failed x509 verification against a TLS-enabled daemon reached over loopback — e.g. `MUNINNDB_ADMIN_URL` / `MUNINNDB_UI_URL` / `MUNINNDB_MCP_URL` = `https://127.0.0.1:…`. This routes the 26 affected call sites through the existing `httpClientForURL` helper (`cmd/muninn/status.go`), which skips verification **only** for a loopback https target and keeps full verification for any off-host endpoint.

> **Stacked on #468**, which introduces `httpClientForURL`. Until #468 merges, this PR's diff against `develop` also shows #468's commit — review the final commit (`feat/443-cli-loopback-tls`) in isolation; the base will collapse once #468 lands.

## Sites retrofitted (26)

- **`vault.go` ×16** — create/delete/clear/clone/rename/merge/reindex-fts/reembed/export/export-markdown/import/job-status/recall-mode/behavior, plus the two bare `GET /api/vaults`
- **`api_key.go` ×3** — list / create / revoke
- **`admin.go` ×1** — change password
- **`repl.go` ×2 + `vault_auth.go` ×1** — the admin/shell **login** POSTs (credentials; now loopback-only skip, fully verified off-host)
- **`repl_client.go` ×3** — MCP health check, `cmdShowStats` health, `mcpCall`

## Behavior preserved

- Every timeout carried through unchanged (10s / 30s / 30m / 5s / 2s); bare calls keep no-timeout (`httpClientForURL(url, 0)`).
- No new `crypto/tls` imports — the helper encapsulates the loopback decision.

## Deliberately excluded

- **`upgrade.go`** — its two clients target the GitHub releases API and the binary-download URL, not the daemon; they must keep full verification.
- **`init.go`'s plasticity client** — `init.go` is rewritten on the `feat/443-init-tls` branch (#465); it's retrofitted there to avoid a cross-branch conflict.

## Merge note — one expected conflict with #465

Excluding `init.go` keeps the two PRs apart there, but **`vault_auth.go`'s `loginAdmin` is edited by both** and will conflict whichever lands second: this PR converts it to `httpClientForURL`, while #465 (based on `develop`, before the helper exists) adds an equivalent inline `https:// && isLoopbackURL → InsecureSkipVerify` block. The two are behaviorally equivalent, so resolve to **this PR's `httpClientForURL` version** (#465's inline comment already says "unifies to `httpClientForURL` once #468 lands").

⚠️ One non-obvious step: after taking the helper version, also **delete the now-unused `"crypto/tls"` import** at the top of `vault_auth.go` — it auto-merges in from #465's side, outside the conflict markers, and otherwise breaks the build (`crypto/tls imported and not used`). The full `#465 + #468 + #469 + #470` tree builds and `go test ./cmd/muninn/` passes once resolved this way.

After both land, one inline skip-verify copy still survives by design in #465's `applyBehaviorToVault` (plasticity client); unifying it onto `httpClientForURL` is the natural follow-up to the cleanup below.

## Follow-up (not in this PR)

~23 of these sites share the same admin-request shape (build URL → client → `NewRequest` → `addSessionCookie` → `Do` → status switch). A `doAdminRequest` helper could own `httpClientForURL` once; deferred to keep this PR a focused, mechanical TLS retrofit.

## Testing

`go build`, `go vet`, and `go test ./cmd/muninn/` all pass; `gofmt` clean. Adds `cli_loopback_tls_test.go` — behavioral tests that run the credential-bearing paths (`loginAdmin`, `shellValidateAdmin`, `autoAuth`) plus `mcpHealthCheck` against an `httptest` TLS server on `127.0.0.1` with a self-signed cert; a verifying client fails the handshake, so they pin the loopback skip-verify path end-to-end (cookie capture included, and bad credentials still rejected). The repo-wide `Vulnerability scan` job stays red until #464 (go 1.26.4) merges — unrelated to this change.

### DIFF
--- a/cmd/muninn/admin.go
+++ b/cmd/muninn/admin.go
@@ -128,10 +128,9 @@ func runAdminChangePassword(username string, args []string) {
 		return
 	}
 
-	client := &http.Client{Timeout: 10 * time.Second}
-	req, err := http.NewRequest("PUT",
-		fmt.Sprintf("%s/api/admin/password", vaultAdminBase),
-		bytes.NewReader(bodyBytes))
+	reqURL := fmt.Sprintf("%s/api/admin/password", vaultAdminBase)
+	client := httpClientForURL(reqURL, 10*time.Second)
+	req, err := http.NewRequest("PUT", reqURL, bytes.NewReader(bodyBytes))
 	if err != nil {
 		fmt.Printf("Error: %v\n", err)
 		return
--- a/cmd/muninn/api_key.go
+++ b/cmd/muninn/api_key.go
@@ -139,10 +139,9 @@ func runAPIKeyCreate(args []string) {
 		return
 	}
 
-	client := &http.Client{Timeout: 10 * time.Second}
-	req, err := http.NewRequest("POST",
-		fmt.Sprintf("%s/api/admin/keys", vaultAdminBase),
-		bytes.NewReader(bodyBytes))
+	reqURL := fmt.Sprintf("%s/api/admin/keys", vaultAdminBase)
+	client := httpClientForURL(reqURL, 10*time.Second)
+	req, err := http.NewRequest("POST", reqURL, bytes.NewReader(bodyBytes))
 	if err != nil {
 		fmt.Printf("Error: %v\n", err)
 		return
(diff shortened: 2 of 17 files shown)


scrypster (approved):
Verified: purely mechanical retrofit of 26 CLI call sites onto httpClientForURL — every timeout preserved, zero new skip-verify code, trust decisions stay centralized in the one helper. End-to-end loopback-TLS credential tests included; upgrade.go correctly excluded so GitHub endpoints stay verified. Vuln-scan red is the stale pre-#464 advisory.

PR #470, opened 2026-06-09
test(cli): use an RFC5737 documentation IP in TestIsLoopbackURL

johanneshauer:
`TestIsLoopbackURL`'s non-loopback case used a specific private LAN address as its sample value. Test/documentation code should use the reserved ranges from [RFC 5737](https://datatracker.ietf.org/doc/html/rfc5737) instead, so this swaps it to `203.0.113.10` (TEST-NET-3) and renames the case `"lan ip"` → `"non-loopback ip"` to match its intent (a non-loopback IP literal, as opposed to the adjacent hostname case).

Pure test-data change — no production code touched. `go test ./cmd/muninn/ -run TestIsLoopbackURL` passes.


scrypster (approved):
Test-only data change, verified trivial. The vuln-scan failure is the pre-#464 govulncheck stdlib advisory, now fixed on develop — admin-merging rather than requiring a rebase.

PR #471, opened 2026-06-09
fix(hnsw): four graph-integrity defects that silently degrade search to a single reachable cluster

isaac-ranger:
### Summary

HNSW search can return as little as **one result** on real vaults regardless of `k`. Semantic recall silently dies, and full-text search masks the failure: lexically overlapping queries still match via FTS and the blended score stays positive, so nothing looks broken from the outside.

Reproduction: bulk-insert 1000 random vectors and self-query each one (query = the vector itself, k=10). On current `main`, only **51/1000** vectors find themselves — every search funnels into one small reachable cluster.

This PR fixes four independent defects in `internal/index/hnsw/hnsw.go` and adds regression tests that fail loudly (51/1000) on the old code. No API changes, no key-format changes — deliberately scoped to correctness.

(Context: we run MuninnDB in production for a multi-agent household. These bugs kept semantic recall dead for ~7 weeks while every surface metric looked green — FTS masking is vicious. Found and fixed by Cairn, one of our agents; commits carry his authorship.)

### The four defects (severity order)

1. **Entry-point promotion before linking** (`Insert`). When a node's random level exceeds `maxLevel`, the old code set `entryPoint = id` *before* computing `ep := idx.entryPoint`. Traversal then started at the new, still-unlinked node: `searchLayer` saw only that node, and the new node linked to nothing (or itself). **Every maxLevel raise orphaned the entire prior graph.** Fix: traverse from the old entry point, link at layers `0..oldMaxLevel`, promote afterward.

2. **Pruning truncates the newest edge** (`Insert`, back-edge loop). `nbNode.layers[l] = nbNode.layers[l][:maxConn]` keeps the oldest edges and always drops the just-appended one. Once a popular node's layer fills (M0=32 at layer 0), **every subsequent node gets zero in-edges** and is permanently unreachable. Fix: `pruneNeighbors` keeps the `maxConn` *nearest* neighbors by cosine distance to the owning node's vector; neighbors with missing vectors prune first. Neighbor vectors are immutable after insert, so reading them under the owner's lock only is safe. Cost is O(maxConn × dim) per overflow — negligible against the embedding cost of the insert itself.

3. **`LoadFromPebble` has no vault filter.** It scans the entire `0x07` range, so every per-vault index loads every vault's nodes: memory bloat, plus cross-vault IDs surfacing as candidates that can never resolve in the querying vault (they are silently dropped downstream, eating the candidate budget). Fix: skip keys where `key[1:9] != idx.ws`.

4. **Back-edges never re-persisted.** `Insert` persisted only the new node; the mutated neighbor lists lived in memory only. Every restart shed all back-edges, leaving the on-disk graph forward-only and barely navigable after reload. Fix: collect mutated neighbors and `persistNode` each after the insert completes (same async pattern, `persistWg`-tracked). *Note for review:* this multiplies persist writes by up to M per insert (NoSync, async). If that's a concern, an alternative is a dirty-set flushed on Close/interval — the simple version measured fine for us.

### Tests

- `TestBulkInsertSelfQuery` — 1000 random 1536-dim vectors via `Registry.Insert`, self-query each with k=10. Before: 51/1000 self-hits, 51 distinct top-1 IDs (the lone reachable cluster). After: **1000/1000 self-hits, 1000 distinct top-1**.
- `TestBulkInsertReloadSelfQuery` — same, then close indexes (flushing async persists), reload a fresh `Registry` from Pebble, re-query: **500/500** post-reload. Before fix 4, the reloaded graph was forward-only (37/500).
- Full `./internal/index/...` and `./internal/engine/...` suites pass.
- The existing hnsw tests pass in ~8ms on toy graphs that never trip the maxLevel-raise or pruning-saturation paths — which is how these defects survived. The new tests are the regression guard.
- Real-data validation: on a copy of our production vaults (v0.6.0 + these fixes + re-embed), zero-keyword-overlap natural-language queries retrieve the correct memories by meaning (vector scores 0.39–0.51). The unfixed binary returned vector score 0 on every result for every query.

### Operational note for existing deployments

Existing deployments need an **index rebuild** (re-embed, or wipe the `0x07` range and rebuild) to benefit: graphs persisted by the broken code are structurally starved and reload into the same pathology.

We have follow-on work (boot-time graph self-heal with coverage/reachability/self-retrieval checks, and an ACT-R scoring option) that we'd propose separately if there's interest — this PR is deliberately limited to the correctness fixes.


scrypster:
Pushed a small fixup to keep CI green: TestBulkInsertSelfQuery at N=1000/dim=1536 consumed ~4.5 min of the 5-min package timeout under -race, starving the reload test. Resized to N=400/dim=128 — verified it still fails on pre-fix code (47/400 self-hits, same ~12% signature as your 51/1000 repro) and the full hnsw package now runs in ~9s under -race. Excellent find and fix — thank you. Note for existing deployments: degraded graphs need an index rebuild (reembed) to repair, which we'll call out in release notes.

scrypster (approved):
Deep-verified: the four fixes match the independently-confirmed defects (link-before-promote, distance-based pruning instead of positional truncation, vault-scoped LoadFromPebble, back-edge re-persistence). persistNode takes node.mu.RLock so the new neighbor-persistence goroutines are race-safe; Close() waits on persistWg so the reload test is sound. Regression tests verified to fail on pre-fix code (47/400 self-hits) and pass with the fix (400/400); full hnsw package runs ~9s under -race after the test resize. Release note needed: existing deployments must rebuild the index (reembed) to repair already-degraded graphs.

PR #476, opened 2026-06-10
fix(cli): load muninn.env before dispatching any subcommand (#473)

### LINKED ISSUE #473: cli: muninn status/start/stop ignore ~/.muninn/muninn.env, so MUNINNDB_DATA there points the CLI at the wrong data dir
## Problem

`~/.muninn/muninn.env` is loaded only by the daemon-facing entry points — `runServer` (`cmd/muninn/server.go:722`) and `runMCPStdio` (`cmd/muninn/mcp_stdio.go:91`). The lifecycle/status CLI commands (`runStart`, `runStop`, `runStatus` in `cmd/muninn/lifecycle.go`) do **not** call `loadEnvFile()`. So when `muninn.env` sets `MUNINNDB_DATA` (or `MUNINNDB_*_URL`), the CLI reads the PID/`muninn.addrs` sidecar from a **different** directory than the daemon writes them to:

- `muninn start` forks a daemon that loads `muninn.env` and binds/records under the env's data dir, but the parent CLI's "already running?" check and PID write used the compiled-in default dir.
- `muninn status` probes the default dir's sidecar and reports `down`/wrong ports for a daemon that is actually up under the env's dir.
- `muninn stop` can't find the PID.

## Evidence

`git grep -n 'loadEnvFile()' cmd/muninn` → only `server.go:722` and `mcp_stdio.go:91` on `develop`. `lifecycle.go` (`runStart` :70, `runStop` :183, `runStatus` :240) and `main.go` never load it.

## Suggested fix

Call `loadEnvFile()` once early in `main()` (before subcommand dispatch) so every command shares the daemon's config view, or at minimum at the top of `runStart`/`runStop`/`runStatus`. Shell env still wins (the loader is set-if-unset), so this is non-destructive. Wants a small test matrix: `MUNINNDB_DATA` in `muninn.env` → status/stop find the daemon.

## Notes

Surfaced during the #443 review. Pre-existing on `develop`; #465 makes `muninn init` load `muninn.env` (so init is now consistent with the daemon) which incidentally widened the gap with the still-unloaded status/start/stop.

johanneshauer:
Closes #473.

## What

`loadEnvFile()` was called only by the daemon entry points — `runServer` and `runMCPStdio` — so the lifecycle/status CLI never read `~/.muninn/muninn.env`. When that file set `MUNINNDB_DATA` (or `MUNINNDB_*_URL`), the forked daemon honored it but `muninn status` / `stop` / `start` used the compiled-in default data dir:

- `muninn status` probes the default-dir sidecar → reports a live daemon as **down**;
- `muninn stop` can't find the PID;
- the parent `muninn start` does its "already running?" check against the wrong dir.

## Fix

Call `loadEnvFile()` once at the top of `main()`, before subcommand dispatch, so every command shares the daemon's config view. Shell env still wins (the loader is set-if-unset); `runServer`/`runMCPStdio` call it again as a harmless no-op.

## Testing

`go build`, `go vet`, `go test ./cmd/muninn/` pass; `gofmt` clean. Adds `TestLoadEnvFile_SetsDataDirForCLI` — `MUNINNDB_DATA` in `muninn.env` now steers `defaultDataDir()`, which the lifecycle/status commands use.

## Note

Independent of the rest of the #443 stack (touches only `main.go`). When #465 lands, its `loadEnvFile()` inside `runInit` becomes redundant with this one — a no-op, safe to drop in a later cleanup.

### DIFF
--- a/cmd/muninn/envfile_test.go
+++ b/cmd/muninn/envfile_test.go
@@ -21,6 +21,24 @@ func TestLoadEnvFile_SetsVars(t *testing.T) {
 	}
 }
 
+// TestLoadEnvFile_SetsDataDirForCLI pins the intent of wiring loadEnvFile() into
+// main(): MUNINNDB_DATA set only in muninn.env must steer the CLI's data dir
+// (via defaultDataDir), so 'muninn status'/'stop' look where the daemon — which
+// loads the same file — actually wrote its PID/addrs sidecar.
+func TestLoadEnvFile_SetsDataDirForCLI(t *testing.T) {
+	dir := t.TempDir()
+	dataDir := filepath.Join(dir, "custom-data")
+	path := filepath.Join(dir, "muninn.env")
+	os.WriteFile(path, []byte("MUNINNDB_DATA="+dataDir+"\n"), 0600)
+
+	t.Setenv("MUNINNDB_DATA", "")
+	os.Unsetenv("MUNINNDB_DATA")
+	loadEnvFileFrom(path)
+	if got := defaultDataDir(); got != dataDir {
+		t.Errorf("defaultDataDir() = %q, want %q (muninn.env's MUNINNDB_DATA must steer the CLI)", got, dataDir)
+	}
+}
+
 func TestLoadEnvFile_ShellEnvWins(t *testing.T) {
 	dir := t.TempDir()
 	path := filepath.Join(dir, "muninn.env")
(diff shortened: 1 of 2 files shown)


scrypster (approved):
Verified: single loadEnvFile() at top of main(), set-if-unset so shell env precedence is unchanged and daemon-path calls become no-ops. Test pins the data-dir behavior. Vuln-scan failure is the pre-#464 stdlib advisory, fixed on develop — admin-merging rather than requiring rebase.

PR #477, opened 2026-06-10
fix(cli): muninn status distinguishes a TLS trust failure from a dead server (#474)

### LINKED ISSUE #474: cli: muninn status renders an off-host TLS verification failure as a generic 'down' with a misleading 'restart' hint
## Problem

`probeOnce` discards the error from `client.Get` and returns a bare `bool` (`cmd/muninn/status.go`, `func probeOnce`). After the loopback-gating change (see #468), an off-host `MUNINNDB_ADMIN_URL`/`MUNINNDB_UI_URL`/`MUNINNDB_MCP_URL` pointing at an `https://` endpoint behind an untrusted/self-signed cert is now probed **with** verification — correct — but the resulting x509 failure collapses to the same `(false, "")` as a genuinely dead server. The status display then shows the generic red "not responding" with the remedy `Run: muninn restart`, which is wrong: the server is up; the client doesn't trust its certificate.

## Evidence

```go
func probeOnce(url string, insecure bool) bool {
	...
	resp, err := client.Get(url)
	if err != nil {        // x509 error indistinguishable from connection refused
		return false
	}
	...
}
```

A remote https endpoint with a self-signed cert → `probeHealth` returns `(false, "")` → display renders `down` + "Run: muninn restart".

## Suggested fix

Thread the probe error up (e.g. `probeOnce` returns `(bool, error)` or a small result struct) and special-case `*tls.CertificateVerificationError` / x509 errors in the status renderer: show something like `TLS verification failed — check the server cert or your CA trust` instead of the restart hint. Display-only; no behavior change to the probe decision.

## Notes

Surfaced during the #443 review as a follow-up to #468's (correct) probe tightening. Worth landing alongside or shortly after #468.

johanneshauer:
Closes #474.

> **Stacked on #468**, which introduces the loopback-gated probe this builds on. Until #468 merges, this PR's diff against `develop` also shows #468's commit — review the final commit (`fix/474-status-tls-error`) in isolation; the base collapses once #468 lands.

## What

After #468 gated the status probe's TLS-verification skip on loopback, an off-host `MUNINNDB_ADMIN_URL`/`MUNINNDB_UI_URL`/`MUNINNDB_MCP_URL` pointing at an `https://` endpoint behind an untrusted/self-signed cert is now correctly probed **with** verification. But `probeOnce` discarded the error, so the x509 failure collapsed to the same `(false, "")` as a refused connection — and `muninn status` printed the generic red "not responding → Run: `muninn restart`", the wrong remedy when the server is up and only the cert isn't trusted.

## Fix

- `probeOnce` now returns the underlying error.
- `probeHealth` reports `certErr=true` when an https probe failed **specifically** on certificate verification — untrusted CA, hostname mismatch, or expired/not-yet-valid — detected via `errors.As` over `*tls.CertificateVerificationError` / `x509.UnknownAuthorityError` / `HostnameError` / `CertificateInvalidError`.
- The status display then prints *"reachable but its TLS certificate failed verification"* with trust-fix guidance, instead of the restart hint.

Display-only: the probe's up/down decision is unchanged.

## Testing

`go build`, `go vet`, `go test ./cmd/muninn/` pass; `gofmt` clean. Adds `TestIsCertVerificationError` (cert failure vs connection error vs nil) and tightens `TestProbeOnce_VerifiesWhenSecure` to assert the secure-probe failure is classified as a cert error. Existing `probeHealth` callers updated for the new third return.

### DIFF
--- a/cmd/muninn/admin.go
+++ b/cmd/muninn/admin.go
@@ -128,10 +128,9 @@ func runAdminChangePassword(username string, args []string) {
 		return
 	}
 
-	client := &http.Client{Timeout: 10 * time.Second}
-	req, err := http.NewRequest("PUT",
-		fmt.Sprintf("%s/api/admin/password", vaultAdminBase),
-		bytes.NewReader(bodyBytes))
+	reqURL := fmt.Sprintf("%s/api/admin/password", vaultAdminBase)
+	client := httpClientForURL(reqURL, 10*time.Second)
+	req, err := http.NewRequest("PUT", reqURL, bytes.NewReader(bodyBytes))
 	if err != nil {
 		fmt.Printf("Error: %v\n", err)
 		return
--- a/cmd/muninn/api_key.go
+++ b/cmd/muninn/api_key.go
@@ -139,10 +139,9 @@ func runAPIKeyCreate(args []string) {
 		return
 	}
 
-	client := &http.Client{Timeout: 10 * time.Second}
-	req, err := http.NewRequest("POST",
-		fmt.Sprintf("%s/api/admin/keys", vaultAdminBase),
-		bytes.NewReader(bodyBytes))
+	reqURL := fmt.Sprintf("%s/api/admin/keys", vaultAdminBase)
+	client := httpClientForURL(reqURL, 10*time.Second)
+	req, err := http.NewRequest("POST", reqURL, bytes.NewReader(bodyBytes))
 	if err != nil {
 		fmt.Printf("Error: %v\n", err)
 		return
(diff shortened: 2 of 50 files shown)


scrypster (approved):
Verified: probeHealth/probeOnce now surface the underlying error and classify cert failures via errors.As over typed TLS/x509 errors (no string matching); display-only, probe decisions unchanged. Conflict with squash-merged #468/#463 resolved (three-value signature kept, develop's resolveScheme/portFromAddr retained); full cmd/muninn suite green locally and in CI. Residual all-services-cert-fail edge case tracked in #481.

PR #478, opened 2026-06-10
fix(cli): init embeds the daemon's real token + warns on expired TLS cert (#472, #475)

### LINKED ISSUE #472: cli: muninn init embeds the file token while the daemon prefers MUNINN_MCP_TOKEN — config 401 mismatch
## Problem

`muninn init` writes the token from `~/.muninn/mcp.token` into every generated AI‑tool config, but the daemon resolves its bearer token in a different order — `--mcp-token` flag → `MUNINN_MCP_TOKEN` env → `mcp.token` file (`cmd/muninn/server.go:787-791`, after loading `muninn.env` at `server.go:722`). So if an operator sets `MUNINN_MCP_TOKEN` (e.g. in `muninn.env` or the shell), the running daemon authenticates against that env token while the generated tool configs carry the **file** token — every MCP request from those tools gets `401`.

## Evidence

- Daemon precedence: `cmd/muninn/server.go:787-791` (env wins over the file).
- Init config token: `runInteractiveInit`/`runNonInteractiveInit` source `token` from `--token` or `loadOrGenerateToken(dataDir)` (`cmd/muninn/init.go:164/169` and `:588/592`), which reads/creates the **file** only — it never consults `MUNINN_MCP_TOKEN`.
- That token is what `configureNamedTools(..., token, ...)` embeds (`init.go:182/615` → `mcpServerEntry`, `setup_ai.go:180/234`).

## Repro

```
export MUNINN_MCP_TOKEN=env-token-123
muninn init --tool claude --yes      # daemon now requires env-token-123
# generated config's Authorization: Bearer <mcp.token-file value> != env-token-123
# → MCP calls from Claude get 401
```

## Suggested fix

In init, mirror the daemon's precedence when choosing the token to embed: prefer `MUNINN_MCP_TOKEN` (now visible to init, which loads `muninn.env`) over the file token. ~10 lines + a unit test asserting the embedded token equals the env token when both are set.

## Notes

Surfaced during the #443 review. The scheme axis of "init must match the running daemon" was fixed in #465; this is the token axis of the same class and was left out of scope. Pre-existing on `develop`.

### LINKED ISSUE #475: cli: muninn init accepts an expired TLS cert and prints a success banner (init-side complement to #456)
## Problem

`muninn init`'s TLS validation never checks certificate expiry. `validateTLSPair` delegates to `tls.LoadX509KeyPair`, which parses and key-matches but does **not** inspect `NotBefore`/`NotAfter`; `warnIfCertNotLoopback` uses `VerifyHostname`, which checks SAN coverage only. So an **expired** (or not-yet-valid) loopback cert sails through `muninn init` and prints the `✓ TLS enabled` / `You're live` success banner. Clients that verify the cert then reject every connection, with nothing in the init output hinting why.

## Evidence

- `validateTLSPair` → `tls.LoadX509KeyPair` (no expiry check) — `cmd/muninn/init.go`.
- `warnIfCertNotLoopback` → `x509.Certificate.VerifyHostname` (SAN only, ignores validity period) — `cmd/muninn/init.go`.
- Reproduced: a cert with `NotAfter` one hour in the past and an IP SAN `127.0.0.1` passes `validateTLSPair` and produces zero warnings from `warnIfCertNotLoopback`.

## Suggested fix

In init's TLS path, after parsing the leaf cert (the same scan `warnIfCertNotLoopback` already does), warn when `time.Now()` is outside `[NotBefore, NotAfter]` — e.g. `⚠ certificate expired on <date>` / `⚠ certificate not valid until <date>`. Non-fatal (the daemon is the authority on whether to serve), but it must not be a silent success. Small, self-contained, with a unit test using an expired cert.

## Relationship to #456

#456 adds a **daemon‑startup** cert‑expiry warning. This issue is the **init‑side** complement: catch it at config time, before the success banner, so the operator learns immediately rather than at the next daemon start. We own this one. See the cross-reference comment on #456.

## Notes

Surfaced during the #443 review of #465.

johanneshauer:
Closes #472. Closes #475.

> **Stacked on #465** (`feat/443-init-tls`), which makes `runInit` load `muninn.env`. Until #465 merges, this PR's diff against `develop` also shows #465's commits — review the final commit (`fix/472-475-init-token-and-expiry`) in isolation; the base collapses once #465 lands.

Two init-side correctness gaps, both about init emitting config that matches the daemon's actual settings.

## #472 — embed the token the daemon will actually accept

The daemon resolves its bearer token `--mcp-token` flag → `MUNINN_MCP_TOKEN` env → `~/.muninn/mcp.token` file (`server.go`). Init embedded the **file** token into every generated tool config unconditionally, so an operator with `MUNINN_MCP_TOKEN` set got a daemon expecting the env token while the configs carried the file token — every MCP call `401`'d.

New `configTokenOverride()` mirrors the daemon's precedence (flag > env > file); both init paths use it and fall back to `loadOrGenerateToken` only when neither is set.

## #475 — warn on an expired / not-yet-valid cert

`validateTLSPair` (`tls.LoadX509KeyPair`) and `warnIfCertNotLoopback` (`VerifyHostname`) both ignore the validity period, so an expired cert sailed through init under the `✓ TLS enabled` banner. `warnIfCertExpiring` now flags expired / not-yet-valid / expiring-within-30-days at config time.

Non-fatal — the daemon is the authority on whether to serve, and #456 warns at startup — but init no longer presents silent success over a cert clients will reject. The leaf-cert scan is factored into `parseLeafCert` (shared with `warnIfCertNotLoopback`; handles key-first combined PEMs).

The 30-day window mirrors `internal/tlsutil.ExpiryWarnWindow` from #456; it's inlined only because that package isn't on this branch's base. **Once #456 lands, init should call `internal/tlsutil` directly so the daemon and init share one validity check** — see the cross-reference comment on #456.

## Testing

`go build`, `go vet`, `go test ./cmd/muninn/` pass; `gofmt` clean. Adds `TestConfigTokenOverride` (flag > env > file precedence) and `TestWarnIfCertExpiring` (healthy / expired / not-yet-valid / expiring-soon).

### DIFF
--- a/cmd/muninn/admin.go
+++ b/cmd/muninn/admin.go
@@ -128,10 +128,9 @@ func runAdminChangePassword(username string, args []string) {
 		return
 	}
 
-	client := &http.Client{Timeout: 10 * time.Second}
-	req, err := http.NewRequest("PUT",
-		fmt.Sprintf("%s/api/admin/password", vaultAdminBase),
-		bytes.NewReader(bodyBytes))
+	reqURL := fmt.Sprintf("%s/api/admin/password", vaultAdminBase)
+	client := httpClientForURL(reqURL, 10*time.Second)
+	req, err := http.NewRequest("PUT", reqURL, bytes.NewReader(bodyBytes))
 	if err != nil {
 		fmt.Printf("Error: %v\n", err)
 		return
--- a/cmd/muninn/api_key.go
+++ b/cmd/muninn/api_key.go
@@ -139,10 +139,9 @@ func runAPIKeyCreate(args []string) {
 		return
 	}
 
-	client := &http.Client{Timeout: 10 * time.Second}
-	req, err := http.NewRequest("POST",
-		fmt.Sprintf("%s/api/admin/keys", vaultAdminBase),
-		bytes.NewReader(bodyBytes))
+	reqURL := fmt.Sprintf("%s/api/admin/keys", vaultAdminBase)
+	client := httpClientForURL(reqURL, 10*time.Second)
+	req, err := http.NewRequest("POST", reqURL, bytes.NewReader(bodyBytes))
 	if err != nil {
 		fmt.Printf("Error: %v\n", err)
 		return
@@ -226,7 +225,7 @@ func runAPIKeyList(args []string) {
 		reqURL += "?vault=" + url.QueryEscape(vault)
 	}
 
-	client := &http.Client{Timeout: 10 * time.Second}
+	client := httpClientForURL(reqURL, 10*time.Second)
 	req, err := http.NewRequest("GET", reqURL, nil)
 	if err != nil {
 		fmt.Printf("Error: %v\n", err)
@@ -328,7 +327,7 @@ func runAPIKeyRevoke(args []string) {
 		reqURL += "?vault=" + url.QueryEscape(vault)
 	}
 
-	client := &http.Client{Timeout: 10 * time.Second}
+	client := httpClientForURL(reqURL, 10*time.Second)
 	req, err := http.NewRequest("DELETE", reqURL, nil)
 	if err != nil {
 		fmt.Printf("Error: %v\n", err)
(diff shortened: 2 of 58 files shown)


scrypster (approved):
Verified against issues #472/#475: configTokenOverride mirrors the daemon's flag>env>file precedence exactly (server.go), and warnIfCertExpiring covers expired/not-yet-valid/expiring-soon, non-fatal per the issue's own framing. Merge resolution unified the inlined expiry window onto internal/tlsutil per the PR's own follow-up note. Full cmd/muninn suite green locally and in CI. Known pre-existing gap (--token flag never conveyed to the daemon) noted in the PR body as out of scope.

PR #515, opened 2026-06-11
fix(cluster): lobe keeps the join conn and reads the replication stream (#448 Bug 2)

scrypster:
## Problem (Bug 2 of #448)

The Cortex registers the inbound join connection as the Lobe's peer and streams `ReplEntry` frames over it. But `JoinClient.Join` did `defer conn.Close()`, so the connection died the instant `Join` returned — and `runAsLobe` then just blocked on `<-ctx.Done()` with no reader. The Cortex's first write hit a closed socket (`write: broken pipe`) and **replication never started**. (Bug 1, the streamer/handshake race, was already fixed in #449.)

The existing `net.Pipe` join tests missed this because they call `joinConn` directly, never exercising `Join`'s defer-close.

## Fix — Option A (the agreed direction)

The Lobe keeps the join connection open and reads the stream over it:
- **`Join`** no longer closes the conn on success — it returns it on `JoinResult.Conn` (closing only on handshake error). **The Cortex side is unchanged.**
- **`streamReplication`** (shared by Lobe and Observer) runs join → read-stream → reconnect; **`streamFromCortex`** reads frames and dispatches each via the existing **`HandleIncomingFrame`** (which already applies `ReplEntry`), with ctx-cancellation unblocking the read and a 1s backoff before rejoin.
- **`runAsObserver`** had the identical bug and is fixed by the same loop.

**Key de-risker:** the replication stream is **unidirectional** — the streamer is fire-and-forget and no `ReplAck` is constructed/sent anywhere in the codebase — so the reader loop is read-only, with no writer or bidirectional framing. Composes cleanly with #449's Bug-1 fix (Cortex defers the stream until `JoinResponse` + snapshot are on the wire).

## Tests (real TCP — the lifecycle `net.Pipe` can't exercise)

- `TestJoinClient_Join_PreservesConnForStreaming` — `Join` returns an open conn and a frame the Cortex streams **after** `Join` returns is readable from it (verified RED: no `Conn` field / closed conn before the fix).
- `TestStreamFromCortex_AppliesStreamedEntry` — the reader loop applies a streamed `ReplEntry` to the local store.

Full `internal/replication` suite (incl. all existing join/coordinator/integration tests) green under `-race`.

Credit: production 3-node repro and two-bug diagnosis by @schurabot (#448).

### DIFF
--- a/contrib/cluster/README.md
+++ b/contrib/cluster/README.md
@@ -1,0 +1,72 @@
+# Local MuninnDB cluster (1 Cortex + 2 Lobes)
+
+A runnable example that brings up a three-node MuninnDB cluster on your machine
+so you can see Cortex → Lobe replication working end to end. This is the same
+topology used to validate the replication fix in
+[#448](https://github.com/scrypster/muninndb/issues/448).
+
+For production operations (join tokens, TLS, failover, scaling, DR) see the full
+[Cluster Operations Runbook](../../docs/cluster-operations.md). This directory is
+the quickstart.
+
+## Roles
+
+| Node   | Role      | What it does                                                    |
+|--------|-----------|-----------------------------------------------------------------|
+| cortex | `primary` | Accepts writes, owns the WAL, streams replication to the Lobes. |
+| lobe1  | `replica` | Read-only copy; receives the WAL stream and applies it.         |
+| lobe2  | `replica` | Same as lobe1.                                                  |
+
+## Run it
+
+```bash
+cd contrib/cluster
+docker compose up -d --build      # first run builds the image
+
+# Watch the Lobes join (no "broken pipe" — that was the #448 bug):
+docker compose logs -f cortex lobe1 lobe2 | grep -Ei "joined|cluster"
+```
+
+Within a few seconds each Lobe logs `cluster: joined role=lobe cortex=cortex epoch=1`.
+
+## Prove replication works
+
+Write to the **Cortex**, then read the same id back from a **Lobe**:
+
+```bash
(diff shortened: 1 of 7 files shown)


scrypster:
## Validated end-to-end in a real 3-node Docker cluster (before/after)

Stood up 1 Cortex + 2 Lobes in containers (see the new `contrib/cluster/`) and ran the same write-on-cortex → read-on-lobe check against the pre-fix and post-fix binaries.

**BEFORE (pre-fix `develop` image)** — reproduces the exact reported symptom:
```
lobe1: cluster: joined as lobe cortex=cortex epoch=1
cortex: ERROR cluster: streamer error for lobe lobe=lobe1
        err="write tcp 172.19.0.2:8479->172.19.0.3:53016: write: broken pipe"
cortex: ERROR cluster: streamer error for lobe lobe=lobe2 err="...broken pipe"
→ GET engram on lobe1 → "engram not found"   (replication FAILED)
```

**AFTER (this branch)**:
```
both lobes join, NO broken-pipe errors
→ engram written to the cortex replicates to BOTH lobes (verified with two writes)
```

Same cluster, same config — only the binary differs. The `write: broken pipe` matches the issue's reported error verbatim, and the fix resolves it.

Two findings from the exercise, both filed separately (not regressions, orthogonal to this data-path fix):
- A container port footgun (cluster `bind_addr` must differ from MBP's `0.0.0.0:8474`) — now documented in `contrib/cluster/` + the ops runbook.
- `GET /v1/cluster/info` under-reports topology (Cortex role / per-node lag) while replication works fine → #516.

Issue #512, opened 2026-06-11
SSE push: re-evaluate subscriptions after async embedding lands (double-push design call)


scrypster:
Follow-up to #437 (push delivery fixed in #507). The remaining sub-claim from that report is a design decision, deliberately not implemented in #507.

**Mechanism:** \`handleWrite\` (internal/engine/trigger/worker.go) evaluates a subscription against \`event.Engram.Embedding\` synchronously and falls back to \`vectorScore=0\` when the embedding isn't ready yet. The RetroactiveProcessor inserts the embedding into HNSW later and only calls \`rp.Notify()\` (its own scan loop) — there is no callback back into the trigger/subscription system to re-evaluate with the now-available vector. So a context-filtered subscription with threshold>0 can miss a freshly-written engram whose embedding arrived a moment later.

**The design question:** an \`OnEmbed → re-evaluate subscriptions\` hook would fix it but introduces a double-push (a baseline-scored push at write time, then a vector-scored push once embedded). Options: (a) suppress the write-time push for subs that require a vector and only push post-embed; (b) allow both and document it; (c) debounce. Needs a maintainer decision before implementation.

Low severity — push delivery itself works after #507; this only affects vector-scored matches on just-written engrams.

scrypster:
Implemented in PR #517 (merged to develop). Re-evaluation fires exactly once post-embed via the new `handleEmbed` path; the double-push case is deduplicated via `pushedScores`.

Issue #437, opened 2026-05-15
SSE push events not delivered to SDK clients — param mismatch, SDK parsing gap, and async embed race condition


bstovall:
---

## Summary

SSE trigger push events are not delivered to Python/Node/Swift SDK clients on any port (8475 or 8476). Three issues in the delivery path prevent end-to-end functionality. A fourth issue (REST server only) kills SSE connections after ~30 seconds on port 8475.

## Environment

- MuninnDB v0.5.1 (built from source, `develop` branch at `0c6b88d`)
- macOS (darwin/arm64)
- Python SDK from `sdk/python/` in repo
- Local bundled embedder (bge-small-en-v1.5)
- Tested on both port 8475 (REST server) and 8476 (UI server)

## Reproduction Steps

1. Start MuninnDB with local embedder enabled (`MUNINN_LOCAL_EMBED=1`)
2. Subscribe to SSE events using the Python SDK (port 8476, matching the official example):
```python
async with MuninnClient("http://localhost:8476") as client:
    stream = client.subscribe(vault="default", push_on_write=True, threshold=0)
    async for push in stream:
        print(f"{push.engram_id}: {push.score}")
```
3. Write an engram from another process (MCP, REST, or another SDK client)
4. Observe: no push events are received

### 1. Query param mismatch: `on_write` vs `push_on_write` (both ports)

**File:** `internal/transport/rest/server.go` line 1806

The SSE handler checks:
```go
pushOnWrite := q.Get("on_write") == "true" || q.Get("on_write") == "1"
```

But all three SDKs send `push_on_write`:
- Python: `params["push_on_write"] = str(push_on_write).lower()`
- Node: `push_on_write=${pushOnWrite}` in query string
- Swift: `push_on_write=\(pushOnWrite)` in query string

Subscriptions are created with `PushOnWrite: false`, so `handleWrite()` in `internal/engine/trigger/worker.go` skips them at line 116:
```go
if !sub.PushOnWrite {
    continue
}
```

This affects both ports since the UI server proxies `/api/` routes to the same REST handler (`internal/ui/server.go` line 132: `mux.Handle("/api/", apiHandler)`).

### 2. Python SDK `Push` type doesn't parse nested engram (client-side)

**File:** `sdk/python/muninn/sse.py` line 92, `sdk/python/muninn/types.py` line 143

The server sends the engram as a nested object:
```json
{
  "engram": {"id": "...", "concept": "...", "content": "..."},
  "score": 0.31,
[...]

Issue #498, opened 2026-06-10
Semantic recall returns 0 results whenever `context` has 2+ phrases — phase1 uses the N×dim batch-concatenated embedding as the query vector


johanneshauer:
## Summary

`muninn_recall` with `mode: "semantic"` returns `total: 0` whenever the `context` array contains **two or more phrases** — even when both phrases are *identical*. The same topic returns plenty of results with a single phrase, and the same two-phrase context returns 60+ results in `balanced`/`deep`/`recent` modes.

Root cause: the activation engine's phase1 passes the whole `context` array to `Embedder.Embed`, whose contract returns **N concatenated embeddings as one flat `[]float32`** (N×dim). That flat slice is then used as a *single* query vector. Every cosine comparison against the dim-sized index vectors hits the length-mismatch guard and silently returns `0`, so the vector signal is dead for the entire query. All modes are silently degraded to FTS-only for multi-phrase contexts; `semantic` is the only mode that visibly collapses to zero because its preset puts 0.8 weight on the dead vector component, disables ACT-R (pure weighted sum), and uses threshold 0.3 — the only surviving score term is capped at `0.2·tanh(BM25) < 0.2 < 0.3`, so every candidate is filtered.

The irony: on empty results the handler hints *"For semantic recall, provide more specific context."* — adding more context phrases is exactly what breaks it.

## Environment

- **Version:** v0.6.1 (all file:line references below are v0.6.1)
- **Platform:** Linux x86_64
- **Embed provider:** external HTTP provider; the bug is provider-independent (the flat-concat contract is in `BatchEmbedder` and even the noop embedder)
- Vault with ~1.5k engrams, healthy embeddings (single-phrase semantic recall works)

## Steps to reproduce (MCP)

1. Pick any topic `X` that exists in the vault and confirm the baseline works:

   ```json
   {"context": ["X"], "mode": "semantic"}        → results (total: 10 in my vault)
   ```

2. Send the *same phrase twice*:

   ```json
   {"context": ["X", "X"], "mode": "semantic"}   → {"memories": null, "total": 0}
   ```

3. Controls with the identical two-phrase context:

   ```json
   {"context": ["X", "X"], "mode": "balanced"}   → total: 64
   {"context": ["X", "X"], "mode": "deep"}       → results
   {"context": ["X", "X"], "mode": "recent"}     → results
   ```

Two *different* phrases fail the same way (`["topic A", "topic B"]`, semantic → `total: 0`).

[...]

PR #517, opened 2026-06-11
feat(trigger): re-evaluate push subscriptions when async embedding lands (#512)

scrypster:
## Problem

`handleWrite` runs at write time with `vectorScore=0`, because embeddings are computed asynchronously by the retroactive processor ~tens of ms after the write. So a context/threshold-filtered `PushOnWrite` subscription that should match a freshly-written engram *semantically* couldn't fire — it only saw the non-vector score components (decay/recency/confidence). This is the deferred sub-claim from #437, split out as #512.

## Fix

After the embed retroactive processor inserts a vector into HNSW, it now calls back into the engine (`RetroactiveProcessor.SetOnEmbed` → `Engine.ReevaluatePushOnEmbed`), which resolves the engram's vault (via the existing `store.FindVaultPrefix` seam) and notifies the trigger system. A new `handleEmbed` re-evaluates `PushOnWrite` subscriptions with the now-available vector.

## The double-push design decision (the open question from #437)

Rather than suppress the write-time push, **`handleEmbed` pushes only engrams NOT already delivered at write time** — dedup via the existing `sub.pushedScores`:

- A write-time match (e.g. on FTS) + a later vector match for the same engram → **exactly one** push (the embed re-eval skips it).
- A purely-semantic match that couldn't fire at write time (vectorScore was 0) → fires **once**, correctly vector-scored, when the embedding lands.

No new SSE/trigger reason is introduced; it reuses `TriggerNewWrite`.

## Why this is low-risk

The wiring is isolated and transport-agnostic: a new `EmbedEvent` path in the trigger system + worker (mirroring `EngramEvent`), one `OnEmbed` hook on the embed processor (the enrich processor leaves it nil), vector copied before handing to the async consumer, and a non-blocking notify that drops on a full buffer (the periodic sweep is the backstop).

## Tests (TDD, RED first)

- `handleEmbed` pushes a newly-matching engram once the vector lands.
- `handleEmbed` does **not** double-push an engram already delivered at write time.

Trigger + plugin + engine + cmd/muninn suites green; trigger/plugin green under `-race`.

Credit: reported by @bstovall / diagnosed by @johanneshauer (#437 → split as #512).

### DIFF
--- a/cmd/muninn/server.go
+++ b/cmd/muninn/server.go
@@ -1437,6 +1437,9 @@ func runServer() {
 	var retroProcessor *plugin.RetroactiveProcessor
 	if embedPlugin != nil {
 		retroProcessor = plugin.NewRetroactiveProcessor(pStore, embedPlugin, plugin.DigestEmbed)
+		// Re-evaluate push subscriptions once each engram's embedding lands (#512),
+		// so vector-scored matches on freshly-written engrams are not missed.
+		retroProcessor.SetOnEmbed(eng.ReevaluatePushOnEmbed)
 		retroProcessor.Start(ctx)
 		slog.Info("retroactive embed processor started")
 	}
--- a/internal/engine/engine.go
+++ b/internal/engine/engine.go
@@ -250,6 +250,22 @@ func (e *Engine) SetLatencyTracker(t *latency.Tracker) {
 	e.latencyTracker = t
 }
 
+// ReevaluatePushOnEmbed re-evaluates push subscriptions for an engram whose
+// embedding just finished computing asynchronously (#512). It resolves the
+// engram's vault and forwards to the trigger system, which pushes the engram to
+// any subscription it now matches semantically but was not already delivered to
+// at write time. Wired to the embed retroactive processor's OnEmbed callback.
+func (e *Engine) ReevaluatePushOnEmbed(eng *storage.Engram, vec []float32) {
+	if e.triggers == nil || eng == nil || len(vec) == 0 {
+		return
+	}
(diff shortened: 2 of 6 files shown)


scrypster:
## Proven end-to-end in Docker — and it caught a real bug

Ran a single-node container (bundled bge-small embedder) with an SSE subscription `?context=...&threshold=0.40&push_on_write=true`, wrote a semantically-matching engram, and watched the stream. Same setup against the pre-fix binary (v0.6.3 + #448, no #512) vs this branch:

| | Pre-fix binary | With #512 |
|---|---|---|
| Context/threshold sub receives the matching engram | **0 pushes** | **1 push**, score 0.4298, after the embedding landed |

The vector-scored push is delivered only with #512, only after the async embedding completes — the exact behavior this PR adds.

### The Docker test exposed a gap the unit tests missed

The first run gave **0 pushes even with #512**. Root cause: `handleEmbed` read `sub.embedding` but never *computed* it — subscriptions are created without one (it's filled lazily by the periodic sweep), so an embed event arriving before the first sweep found no vector and silently skipped. The unit tests pre-set `sub.embedding`, masking it.

Fixed in `1b56f6e`:
- `subEmbedding()` computes + caches the sub's context embedding on demand (shared by `handleEmbed` and the sweep).
- Multi-phrase contexts are mean-pooled to a single dim-sized vector (`meanPoolVec`), mirroring #498 — otherwise an N-phrase context is N*dim and cosine vs the engram vector is always 0.
- Two regression tests added (compute-on-demand; multi-phrase pooling).

Re-ran the Docker proof after the fix → the push is delivered. This is why proving it beat trusting the unit tests.

Issue #516, opened 2026-06-11
cluster: /v1/cluster/info doesn't reflect joined lobes — cortex role 'unknown', lobes shown as seed-*, last_seq stuck at 0


scrypster:
Found while validating the #448 Bug 2 fix in a real 3-node Docker cluster (1 cortex + 2 lobes). **Replication itself works** after #448 — a write on the cortex reaches both lobes — but the cluster *membership/status view* does not reflect reality. This is a separate observability gap from the #448 data-path bug.

## Observed (with the #448 fix, replication confirmed working)

`GET /v1/cluster/info` on the cortex:
```json
{"node_id":"cortex","role":"unknown","is_leader":false,"epoch":1,"cortex_id":"",
 "members":[
   {"node_id":"cortex","role":"unknown","last_seq":0},
   {"node_id":"seed-lobe1:8479","role":"unknown","last_seq":0},
   {"node_id":"seed-lobe2:8479","role":"unknown","last_seq":0}]}
```
- **Cortex reports `role: unknown` / `is_leader: false`** even though it is the elected leader and is actively streaming to lobes (role=primary, epoch reached 1).
- **Joined lobes appear as raw `seed-<addr>` entries with `role: unknown`** — the MSP membership view never reconciles them with their real node-ids/roles from the join handshake. (The lobe self-reports `role: replica` correctly, but with `cortex_id: ""`.)
- **`last_seq` stays 0 for every member** — replica progress isn't tracked. This ties to the fact that no `ReplAck` is ever sent (the `NetworkStreamer` is fire-and-forget; `TypeReplAck` is handled on the cortex but never produced by a lobe), so the cortex has no replica-lag signal.

## Why it matters
- `cluster-info` / `cluster-nodes` are the operator's window into cluster health; showing the leader as "unknown" and lobes as unrecognized seeds makes a *healthy* cluster look broken.
- `last_seq`/lag staying 0 means replica-lag monitoring (docs §"Checking Replica Lag") and graceful-failover convergence can't actually observe progress.

## Likely scope (separate from #448)
1. Reconcile MSP membership with join-handler members so a joined lobe is shown by its node-id + role, not `seed-<addr>`.
2. Surface the cortex's own role/leadership + `cortex_id` in cluster-info.
3. Add a lobe→cortex `ReplAck` (the streamer/handler already have the frame type and `UpdateReplicaSeq`) so `last_seq`/lag is real.

[...]

PR #518, opened 2026-06-11
fix(cluster): report accurate topology in /v1/cluster/info (#516)

scrypster:
## Problem

Found while validating #448 in a real cluster: replication works, but `/v1/cluster/info` misrepresents the cluster three ways (diagnosed in #516).

## Fixes (each TDD'd, RED first)

**Part 1 — Cortex reports `role: unknown` instead of leader.**
A node configured `role: primary` registers its seeds as voters at construction, so the bootstrap election needs a quorum the not-yet-joined seeds can't provide; it stays a *candidate* (role "unknown") while still serving joins and streaming. `bootstrapPromoteIfDesignatedPrimary` asserts leadership in that case — **guarded** to explicit `"primary"` only, and only while still a candidate (so it never overrides a real leader that sent a CortexClaim), reusing the existing crash-recovery promotion path.

**Part 2 — joined lobes shown as `seed-<addr>` with role `unknown`.**
cluster-info read the raw MSP peer set (synthetic seed placeholders). `ReconciledMembers` merges the join-handshake members (real node-id, role, last seq) over the placeholders by address; the internal `KnownNodes()`/MSP view is left untouched.

**Part 3 — `last_seq` stuck at 0.**
No `ReplAck` was ever sent. The lobe now sends one (node-id + applied seq) after each applied `ReplEntry`, over the same connection the Cortex already reads (`handleClusterConn` → `HandleIncomingFrame` → `UpdateReplicaSeq`); `ReconciledMembers` overlays the latest acked seq per node.

## Proven end-to-end in a 3-node Docker cluster

`/v1/cluster/info` on the cortex:

| | Before | After |
|---|---|---|
| Cortex | `role: unknown`, `is_leader: false`, `cortex_id: ""` | `role: primary`, `is_leader: true`, `cortex_id: "cortex"` |
| Lobes | `seed-lobe1:8479` / `seed-lobe2:8479`, role `unknown` | `lobe1` / `lobe2`, role `replica` |
| `last_seq` | `0` | `2` (live) |

Replication still works (write→lobe verified) with **zero** broken-pipe errors after the election change. Full `internal/replication` suite green under `-race`.

Diagnosis + repro from validating #448 (@schurabot).

### DIFF
--- a/internal/replication/bootstrap_promote_test.go
+++ b/internal/replication/bootstrap_promote_test.go
@@ -1,0 +1,60 @@
+package replication
+
+import "testing"
+
+// #516 part 1: a node explicitly configured role=primary never formally promotes
+// when its seeds are registered as voters but are not up yet to vote — self-vote
+// alone is short of quorum, so it stays a candidate (role "unknown") even though
+// it serves joins and streams. bootstrapPromoteIfDesignatedPrimary asserts
+// leadership in that case.
+func TestBootstrapPromote_DesignatedPrimaryPromotes(t *testing.T) {
+	coord, _ := newTestCoordinator(t, "primary")
+
+	// Simulate StartElection having bumped the epoch and self-voted, but not
+	// reached quorum (seeds not up): still a candidate for epoch 1.
+	coord.election.mu.Lock()
+	coord.election.state = ElectionCandidate
+	coord.election.candidateEpoch = 1
+	coord.election.mu.Unlock()
+
+	coord.bootstrapPromoteIfDesignatedPrimary()
+
+	if coord.Role() != RolePrimary {
+		t.Errorf("designated primary did not promote: role=%v", coord.Role())
+	}
+	if !coord.IsLeader() {
+		t.Error("expected IsLeader() == true after bootstrap promotion")
+	}
+}
+
+// An auto-role node must NOT force-promote — it goes through the normal election.
+func TestBootstrapPromote_AutoRoleDoesNotForcePromote(t *testing.T) {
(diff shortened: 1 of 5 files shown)


scrypster:
## Independent adversarial review: SHIP

An independent reviewer scrutinized this against the cluster subsystem's bug history. Both make-or-break questions resolved favorably:

- **Part 1 cannot cause split-brain or override a legitimate leader** in any correct topology. The lock discipline mirrors the existing crash-recovery promotion (set `state=Leader` under `election.mu`, then `handlePromotion` re-checks `state==Leader` and aborts on a concurrent CortexClaim). The restart-at-epoch>0 path is safe: `StartElection` CASes the epoch upward *before* the force-promote, so epoch fencing demotes any stale leader — same end state as pre-PR, just immediate instead of vote-gated.
- **Part 3's ack `NodeID` matches the cortex's `replicaSeqs` key end-to-end** (lobe `cfg.NodeID` → `UpdateReplicaSeq` → `replicaSeqs` → `ReconciledMembers` lookup), so `last_seq` populates. No deadlock — the cortex reads acks in a separate goroutine from its streamer writer, always draining.
- Part 2 reconciliation is correct and race-safe (fresh value structs, only the reporting path changed).

**One non-blocking concern, tracked as #519:** two nodes *both* misconfigured `role: primary` now fail unsafe (dual leadership) vs pre-PR fail-safe (stalemate). Operator-misconfiguration only; the intended single-primary topology is safe. #519 proposes surfacing that misconfig with a loud warning.

Merging on green CI per the reviewer's SHIP verdict.

Issue #522, opened 2026-06-11
[Epic] Cluster liveness & membership overhaul — peer-identity reconciliation (root cause of #516/#519/#520/#521)


scrypster:
Tracking epic for the cluster liveness/membership problems surfaced while fixing #516. Root-caused via deep analysis; this issue captures the design and landing plan. Sub-issues: #516 (reporting — partial), #519 (two-primary), #520 (quorum-loss demotion — blocked), #521 (MSP liveness — root).

## Root cause: peer-identity failure, NOT missing transport

Heartbeat frames already flow Cortex↔Lobe over the join/stream connection in **both** directions (Cortex pings via `PeerConn.Send`; the Lobe's `handleClusterConn` / `streamFromCortex` reads them). They are **silently dropped on a naming mismatch.** Cluster membership lives in **four maps across two keyspaces**, mutated in different places:

| Structure | seeds keyed as | joined nodes keyed as | mutated at |
|---|---|---|---|
| `ConnManager.peers` | `seed-<addr>` (dead conn) | real id (live conn) | coordinator.go:291,778 |
| `MSP.peers` | `seed-<addr>` | **never added** | coordinator.go:293 |
| `election.voters` | `seed-<addr>` + self | **never added** | coordinator.go:284,296 |
| `joinHandler.members` | — | real id | join.go:154 |

`MSP.handleHeartbeat` drops a ping from `"cortex"`/`"lobe1"` because its map holds only `seed-*` (msp.go:125-127). Four independent identity defects; none are transport.

## Blast radius (all the same root cause)

Because a Lobe never registers the Cortex conn in *its* ConnManager, **every Lobe→Cortex reply via `GetPeer` is a silent no-op in production**: `VoteResponse` (coordinator.go:834), `HandoffAck` (1482), CCS (ccs.go:333), recon (reconcile.go:413/453), cognitive-forward (1053). So **remote elections and automatic failover are also non-functional** — tests pass only because they populate both ConnManagers symmetrically. Sentinels dial nothing and are dialed by nothing (deaf in production). #519's two primaries hold each other as never-dialed placeholders, so CortexClaims never flow.

This all works *today* only because `checkQuorumHealth` runs only on `OnSDown` (never periodically), so nothing acts on the broken liveness; replication flows over the join/stream conn regardless.

## Recommended design — Option A′ (chosen)

One identified, authenticated TCP conn per peer pair; all frame types multiplex over it; MSP keys exclusively on real node-ids.

[...]

Issue #521, opened 2026-06-11
cluster: MSP heartbeats never flow Cortex↔Lobe — liveness/quorum is non-functional (masked today)


scrypster:
Discovered while attempting #520 and validating it in Docker. This is the foundational issue under #516 / #519 / #520.

## What happens
In a healthy 1-cortex + 2-lobe cluster (replication working), the Cortex's MSP **never sustains the lobes as live peers**:
- The Cortex's MSP peer set is the **seed placeholders** (`seed-lobe1:8479`, …) added at construction. They start un-SDOWN (assumed alive), then go **SDOWN after ~3 missed heartbeats (~3s)** because no MSP ping/pong actually flows between the Cortex and the joined lobes.
- The joined lobes (real node-ids `lobe1`/`lobe2`) are **never established as live MSP peers** — the join/stream connection carries replication, but MSP heartbeats do not run over it, and there is no separate MSP heartbeat channel between Cortex and Lobe.

Net: the Cortex's live-peer count collapses to `alive=1` within seconds, so its quorum view is wrong even though the cluster is healthy and replicating.

## Why it's invisible today
`checkQuorumHealth` (the safety meant to demote a quorum-starved Cortex) is only invoked on MSP `OnSDown` transitions, never periodically — so the broken liveness is never *acted on*. The Cortex stays leader and replication keeps working over the join/stream path. The cluster functions **despite** non-functional liveness.

## Why it matters / blast radius
Anything that depends on accurate live-peer/quorum state is built on sand:
- **#520** (make the quorum-loss self-demotion actually fire): unsafe until liveness works — a correct periodic check demotes a *healthy* single-primary cluster within ~5s (verified in Docker: cortex bootstraps, lobes join, then `quorum lost alive=1` → `pre-emptively demoting` → whole cluster collapses to `replica/False`).
- **#519** (two-primary detection): the two primaries never connect/exchange CortexClaims for the same reason — no Cortex↔peer MSP channel.
- Automatic failover (ODOWN), replica-lag, and quorum-loss split-brain protection all consume this liveness and are therefore unreliable.

## Repro
3-node compose (1 primary + 2 replicas, seeded). Cortex log: `promoted to Cortex` → ~3s later `quorum lost, starting demotion timer alive=1 quorum=2`. The lobes joined (`cluster: joined role=lobe`) and replication works, yet the Cortex counts only itself.

## Direction
[...]

Issue #520, opened 2026-06-11
cluster: Cortex never self-demotes on SUSTAINED quorum loss (checkQuorumHealth only runs on SDOWN transitions, not periodically)


scrypster:
Found while validating #519 in Docker.

`checkQuorumHealth` (coordinator.go) is the safety that pre-emptively demotes a Cortex which can't reach a quorum of live peers for `QuorumLossTimeoutSec` (default 5s), to prevent split-brain writes. Its own comment says it "is called periodically (from the MSP tick goroutine)" — **but it is only wired to `msp.OnSDown`** (coordinator.go:209-211), which fires once per peer SDOWN *transition*, not on a timer.

**Consequence:** on *sustained* quorum loss, the first SDOWN sets `quorumLostSince` and logs `quorum lost, starting demotion timer` — then no further SDOWN events occur, so `checkQuorumHealth` is never called again, the `now - quorumLostSince >= timeout` branch never runs, and **the Cortex never demotes**. The pre-emptive demotion only fires if a *second* SDOWN happens to land after the timeout — incidental, not guaranteed.

**Repro (Docker):** two nodes both misconfigured `role: primary`, seeded to each other. Each force-promotes (epoch 1), logs `quorum lost, starting demotion timer (alive=1 quorum=2)` at ~3s — and then **both remain `is_leader: true` indefinitely** (verified at 90s). Expected: each demotes ~5s after detecting sustained quorum loss.

**Fix direction:** call `checkQuorumHealth` from an actual periodic ticker (e.g. the MSP tick) so the timer is re-evaluated while quorum stays lost. NOTE: this interacts with the designated-primary bootstrap (#516 part 1) — a single primary legitimately has no quorum until its lobes join, so the demotion must not fire during the normal bootstrap/join window (needs a grace period or a "had-quorum-then-lost-it" gate). That tuning is the delicate part and why this is filed for deliberate work, not a quick patch.

Related: #519 (two-primary detection). Distinguishing "single primary awaiting lobes" from "two conflicting primaries" fundamentally needs the primaries to communicate (cortex↔cortex connectivity), which today they don't — connection setup is join-driven and neither primary joins the other.

scrypster:
[...]

PR #524, opened 2026-06-11
fix(cluster): rename-on-join identity reconciliation + bidirectional heartbeats (#522 Step 1)

scrypster:
Step 1 of the cluster liveness overhaul (epic #522). Makes Cortex↔Lobe liveness **actually work** by fixing the peer-identity mismatch that silently dropped every heartbeat.

## Root cause (from the epic)
Cluster membership lived in four maps across two keyspaces: the Cortex held seeds as `seed-<addr>` in MSP + the voter set, but Lobes join under real node-ids that were **never added** to either. So heartbeats/votes keyed by the real id were dropped, the never-heartbeated `seed-*` placeholder drove a false SDOWN within ~3s, and the Cortex's live/quorum view collapsed to `alive=1` even though replication worked. The Lobe likewise never registered the Cortex in its own ConnManager → every Lobe→Cortex reply (`VoteResponse`, `ReplAck`, `CogForward`, `HandoffAck`) was a silent no-op in production.

## Fix
- **`reconcileJoinedPeer`** — replace the `seed-<addr>` placeholder matching an address with the real node-id in MSP + voters, **conserving the voter count** (rename, never add) so quorum semantics don't shift. Idempotent across rejoins.
- **Cortex** `HandleIncomingJoin` reconciles the joining Lobe's identity.
- **Lobe** `streamReplication` registers the Cortex conn in its ConnManager and reconciles the Cortex identity (using `JoinResponse.CortexAddr` from Step 0), so its MSP heartbeats reach the Cortex and the reply paths actually send; tears the registration down on stream end.
- **`sendReplAck`** routes through the registered `PeerConn` so it serializes with MSP heartbeats on the shared connection.

Heartbeats now flow both directions over the existing join/stream connection (both physical directions already existed — the bug was naming).

## Validated end-to-end in a 3-node Docker cluster
- **Healthy 1+2: no spurious SDOWN / quorum-lost through 35s** (was: collapse to `alive=1` within ~3s).
- Liveness tracks reality: kill **one** lobe → no false quorum-lost (quorum still met); kill **both** → `quorum lost alive=1` (real detection).
- Replication works; `last_seq=2` for both lobes via the PeerConn-routed `ReplAck`.
- Members shown by real id + role `replica`; restore lobes → quorum recovers, cortex stays primary.

## Scope
`checkQuorumHealth` stays `OnSDown`-only this step. Periodic demotion (#520) is gated on the Step 3 demotion-recovery path per the epic — do not land it before then. Tests: `reconcileJoinedPeer` renames conserving voters + idempotent; full `internal/replication` suite green under `-race`.

Refs #522, #521.

### DIFF
--- a/internal/replication/coordinator.go
+++ b/internal/replication/coordinator.go
@@ -7,6 +7,7 @@ import (
 	"log/slog"
 	"math/rand"
 	"net"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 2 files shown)


scrypster:
## Independent adversarial review → one real fix applied

The reviewer confirmed **no frame-corrupting write race** in the shared-conn multiplexing (all writers serialize through `PeerConn.Send`'s mutex; read+write on one `net.Conn` is the supported pattern; the write deadline is per-frame inside the lock).

But it caught a genuine correctness bug the homogeneous-address Docker test couldn't surface: **`reconcileJoinedPeer` could inflate the voter set when a joined node's advertised address ≠ the seed string** (DNS seed vs pod IP, `0.0.0.0` binds) — the seed placeholder lingered as a permanent voter while the real id was also added, raising quorum and risking false quorum-loss. A new regression on the Lobe side.

Fixed in the latest commit — reconciliation now **conserves the voter count by construction** (retire exactly one seed per new join; arbitrary fallback on addr mismatch with a loud warning; rejoin retires nothing). Also applied the reviewer's lower-severity items: add-before-remove ordering (no transient under-count), ConnManager seed cleanup, and skipping voter registration on observers. New regression tests cover the match / mismatch / rejoin cases; full suite green under `-race`.

PR #526, opened 2026-06-11
fix(cluster): post-demotion supervisor — no more demotion zombie (#522 Step 3 PR A)

scrypster:
Step 3 PR A of the cluster liveness overhaul (epic #522). Fixes the **demotion zombie**.

## Problem
A demoted Cortex became a zombie: `handleDemotion` set `role=Replica` but the `runAsCortex` goroutine stayed parked at `<-ctx.Done()` with no rejoin/re-election loop — **and** the quorum-loss path left `election.state=Leader`, so `StartElection` was blocked by `errAlreadyCandidate` forever. In Docker this collapsed the whole cluster to `replica/False` after a quorum blip.

## Fix — a supervisor state machine
Replace the terminal parks with `supervise()`: **LEADING / FOLLOWING / WAITING_QUORUM**, transitioning on promotion/demotion events; it never exits except on ctx cancel. Demotion carries a **cause**:
- `causeClaim` (a higher-epoch leader / handoff target exists) → **FOLLOWING**: join+stream from that leader (leader-first seed order); promote on an ODOWN-win.
- `causeQuorumLoss` (no successor) → **WAITING_QUORUM**: wait leaderless until a live voter quorum returns (lobes rejoin), then start a fresh election. Re-promotion needs a real vote quorum ⇒ no thrash.

Supporting: `election.StepDown()` (fixes the second lock), `IsVoter()`+`aliveVoters()` (correct quorum population, excludes observers), `streamReplication` seed-parametrized + **non-fatal** on join exhaustion (was a second death path that exited `Run`), `runAsLobe` now runs through the supervisor so a Lobe that wins an election leads instead of staying wedged (the inverse zombie), and a per-term `hadQuorum` latch (consumed by PR B).

## Validation
- **Docker healthy-cluster regression**: cortex stays primary, lobes live at members=3, replication works, **zero spurious demotion** — the supervisor is a no-behavior-change for healthy clusters.
- **Unit**: `StepDown` unblocks re-election; `IsVoter`; `leadTerm` routes by cause; `waitQuorumTerm` defects on a foreign leader and re-elects when quorum returns. Full `internal/replication` suite green under `-race`.

## Scope / sequencing
The periodic `checkQuorumHealth` that *triggers* the quorum-loss path (so the full demote→recover cycle fires, and D2/D3/D4 Docker scenarios can run) lands in **PR B (#520)** — kept separate per the epic so a "healthy cluster never demotes" regression bisects trivially. PR B will Docker-validate the full demote→recover. (R5 join-rejection-when-following + graceful-failover E2E also land with PR B's validation.)

Refs #522.

### DIFF
--- a/internal/replication/coordinator.go
+++ b/internal/replication/coordinator.go
@@ -781,12 +781,24 @@ func (c *ClusterCoordinator) initialCortexMode() runMode {
 }
 
 // leadTerm holds leadership until a demotion event arrives, then routes to the
-// recovery mode the demotion cause dictates.
+// recovery mode the demotion cause dictates. A backstop ticker re-derives role
+// from authoritative state so that even if a demotion event were ever dropped
+// (unreachable in practice — the loop drains roleCh while leading), the node
+// still recovers instead of staying parked as a leader-but-Replica zombie.
 func (c *ClusterCoordinator) leadTerm(ctx context.Context) runMode {
+	ticker := time.NewTicker(c.heartbeatInterval())
+	defer ticker.Stop()
 	for {
 		select {
 		case <-ctx.Done():
 			return modeLeading
+		case <-ticker.C:
+			if !c.IsLeader() {
+				if ldr := c.election.CurrentLeader(); ldr != "" && ldr != c.cfg.NodeID {
+					return modeFollowing
+				}
+				return modeWaitingQuorum
+			}
 		case ev := <-c.roleCh:
 			if ev.promoted {
 				continue // already leading; ignore a duplicate promote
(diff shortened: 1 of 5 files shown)


scrypster:
## Independent adversarial review → SHIP, two items hardened

The reviewer scrutinized the supervisor against the cluster's bug history and cleared all three make-or-break questions:
- **No goroutine leak/deadlock in `followTerm`** — every exit path joins the stream goroutine via `defer termCancel(); <-done`, and `streamReplication` returns promptly on termCtx cancel (all blocking points honor ctx).
- **No new split-brain in `waitQuorumTerm`** — re-election uses the unchanged epoch-CAS + per-epoch-single-vote majority guards; unsolicited non-leader election already existed pre-PR (OnODown), this just rate-limits it.
- **No regression** — crash-recovery → `supervise(modeLeading)` parks identically; `runAsLobe`/`streamReplication` behavior preserved; all callers updated.

Applied both non-blocking follow-ups (latest commit):
1. **`leadTerm` backstop ticker** — it was the only term without periodic re-derivation; the dropped-event safety was real but undocumented. Now re-derives role on a heartbeat ticker, so the "every term re-derives each tick" invariant actually holds (defense-in-depth against a theoretical dropped demote event).
2. **`OnBecameLobe` async** — it runs on the MSP tick goroutine via `checkQuorumHealth` and can block stopping cognitive workers; fired async now (the role flip already gates writes).

Full suite green under `-race`.

Issue #529, opened 2026-06-11
cluster: a joining Observer is registered as a voter (reconcileJoinedPeer hardcodes RoleReplica)


scrypster:
Found during Step 4 design (epic #522).

`HandleIncomingJoin` hardcodes the joiner's role as `RoleReplica` (coordinator.go ~1124) because `JoinRequest` has no role field (cluster_frames.go). `reconcileJoinedPeer` then gates voter registration on the **local** node's role only (`isVoter := c.cfg.Role != "observer"`, coordinator.go ~646) and registers the joiner as a voter. So an **Observer that joins** a Cortex is wrongly added to the voter set — inflating quorum, contrary to Observers being non-voting (they're correctly excluded from self-registration in `Run`).

## Fix (two lines, same blast area as Step 4's voter-gating)
- Add `Role uint8 \`msgpack:"role,omitempty"\`` to `JoinRequest` (0 = legacy = replica); stop hardcoding `RoleReplica`.
- `reconcileJoinedPeer`: `isVoter := c.cfg.Role != "observer" && role != RoleObserver`.

Covered by the existing voter-set conservation invariant tests. Lands with Step 4 (PeerHello supplies an authenticated role for the same gating).

scrypster:
Resolved: the hello path was voter-role-gated in #530 and the join path now carries the joiner's role (JoinRequest.Role, 0=legacy=replica) in #532, so a joining Observer is no longer registered as a voter. A retired seed placeholder is also always unregistered. Covered by the voter-conservation + observer-vote tests.

Issue #528, opened 2026-06-11
cluster: ODOWN never fires in a ≥3-voter cluster — automatic failover cannot trigger (no SDOWN gossip)


scrypster:
Found during the Step 4 (PeerHello) design analysis for the cluster-liveness epic #522.

## The bug
ODOWN (objectively-down: a quorum of nodes agree a peer is down) is what triggers an automatic election after the Cortex dies (`msp.OnODown → StartElection`, coordinator.go). But ODOWN **can never fire** in any cluster with quorum ≥ 2:

- `MSP.tick` computes ODOWN as `1 + len(votedDown[target]) >= nonObserverQuorum` (msp.go ~304-313).
- `votedDown` is **never populated** — grep shows no writer.
- The `TypeSDown` frame (0x33) that would carry a "I see peer X as down" vote is **never sent** and has **no `HandleIncomingFrame` case** (coordinator.go).

So a node only ever has its *own* subjective SDOWN (1 vote); it never receives others' down-votes, and `1 < quorum` for any real cluster. **Automatic failover has never actually completed** in a multi-node cluster — only the single-node (quorum=1) case promotes.

## Why it surfaced now
The cluster's liveness was non-functional until the Step 1–3 fixes (#524/#526/#527), so this was masked — peers were never reliably "alive" to begin with. With accurate liveness, this is the next blocker for automatic failover.

## Fix (≈40 lines, scoped to Step 4)
1. In `MSP.tick`, on a non-observer peer's SDOWN **transition**, `Broadcast(TypeSDown, {SenderID, TargetID, Epoch, Ts})` (the frame type 0x33 and a notification struct already exist).
2. `MSP.RecordDownVote(sender, target)`: populate `votedDown[target][sender]`; if the target is locally SDOWN and votes reach `nonObserverQuorum`, fire `OnODown` once (latch; clear on the target's heartbeat recovery).
3. New `case mbp.TypeSDown:` in `HandleIncomingFrame` → `RecordDownVote`.

Requires the Step 4 PeerHello mesh (Lobe↔Lobe connectivity) to actually deliver the gossip. Tracked under #522; lands with the failover Docker scenario. Also needs election split-vote jitter (two lobes can ODOWN-elect simultaneously at the same epoch → split vote → no retry today).

PR #530, opened 2026-06-11
feat(cluster): PeerHello discovery mesh + equal-epoch tie-break — converges two primaries (#522 Step 4 a+b, #519)

### LINKED ISSUE #519: cluster: detect/warn on two-designated-primaries misconfiguration (same-epoch CortexClaim conflict)
Follow-up from the adversarial review of #518 (#516 part 1).

#518 makes a node configured \`role: primary\` force-promote itself at bootstrap when its seeds can't form quorum yet (fixing the common single-primary case where the cortex reported \`role: unknown\`). The force-promotion takes leadership with a single self-vote.

**New failure mode under misconfiguration:** if **two** nodes are both configured \`role: primary\` (separate epochStores), each bumps its own epoch 0→1 and force-promotes → two leaders at the same epoch, then thrashing CortexClaims. Pre-#518 this failed *safe* (each candidate rejects the other's same-epoch VoteRequest → no quorum → stalemate, no leader). Post-#518 it can fail *unsafe* (dual leadership). The same theoretical hole exists if a \`role: primary\` node races an \`auto\`-role node that can reach quorum without it.

This only manifests under operator misconfiguration — the intended topology is exactly one designated primary + replicas/sentinels (which never initiate elections), and that is safe. But we should surface the misconfig loudly rather than thrash silently.

**Proposed:** when a node that force-promoted as a designated primary later receives a **same-epoch** CortexClaim from a *different* node, log a loud WARN (and ideally one side defers, e.g. lowest node-id keeps leadership) so the dual-primary misconfig is visible in logs/health instead of manifesting as election thrash. Low-risk observability win on top of the #518 correctness fix.

scrypster:
Step 4 (a)+(b) of the cluster liveness overhaul (epic #522). Closes the connectivity gap and resolves #519.

## Problem
The connection lifecycle was entirely join-driven (a Lobe dials a Cortex). Nodes with no join relationship never connected — so two `role=primary` nodes never exchanged CortexClaims (split-brain), sentinels were deaf, and Lobes couldn't reach each other.

## Fix
An authenticated peer-discovery handshake that establishes identified connections feeding the existing MSP liveness + elections:
- **`TypePeerHello` (0x41)** + `PeerHello{node-id, addr, role, epoch, HMAC over node-id+addr+role}` — role is in the MAC because it gates voter registration. Empty secret = open mode.
- **`runPeerDiscovery`**: dial each seed with no live identified conn, exchange hellos, reconcile identity into MSP+voters, spawn a reader for the outbound conn. `handleClusterConn` accepts `TypePeerHello` as an alternative first frame.
- **`RegisterConnKind` tie-break**: the lower node-id's outbound conn is canonical; a non-canonical hello never evicts a live conn; a live join conn is never evicted by a hello — a pair converges on one conn without flapping.
- **Claim-on-connect**: a leader sends its CortexClaim over a newly-established conn, so two previously-isolated leaders finally exchange claims.
- **Equal-epoch CortexClaim tie-break**: two leaders at the same epoch resolve by lowest node-id (lower keeps + re-asserts; higher yields → supervisor follows). Fixes #519.
- **Voter role-gating** (#529): a hello/join from an observer is not registered as a voter.

## Proven in Docker (#519)
Two nodes both `role=primary`, seeded at each other:
- **Converge on one leader**: cortex-a (lower id) keeps leadership; cortex-b demotes (`cause=claim`) → `role=replica`, `cortex_id=cortex-a`.
- **Stable, no flap**: still exactly one leader 25s later; **1** tie-break event total.
- **Replication works**: a write to the leader appears on the converted follower.

## Tests
Equal-epoch tie-break (lower keeps / higher yields / higher-epoch always wins); `RegisterConnKind` precedence (canonical/non-canonical/join); HMAC verify+reject (role-tamper, wrong secret, open mode). Full `internal/replication` + `mbp` suites green under `-race`.

## Scope / what's next
This is Step 4 (a)+(b). Remaining: **(c)** SDOWN gossip (#528 — ODOWN never fires today, so lobe-majority *automatic* failover still can't trigger) + election split-vote jitter; **(d)** the `JoinRequest` role field (the other half of #529). Sequenced per the epic.

Refs #522, #519, #529.

### DIFF
--- a/cmd/muninn/server.go
+++ b/cmd/muninn/server.go
@@ -690,8 +690,23 @@ func handleClusterConn(conn net.Conn, coord *replication.ClusterCoordinator) {
 			connOwned = false // PeerConn now owns conn
 			continue
 		}
-		// Reject any frame that arrives before the join handshake completes.
-		// A well-behaved Lobe always sends TypeJoinRequest first.
+		// PeerHello discovery handshake (#522 Step 4) — an alternative first frame
+		// for peers that don't join (two primaries, sentinels, lobe↔lobe).
+		if frame.Type == mbp.TypePeerHello {
+			nodeID, adopted, err := coord.HandleIncomingHello(conn, frame.Payload)
+			if err != nil {
+				log.Printf("[cluster] hello error from %s: %v", fromNodeID, err)
+				return
+			}
+			if !adopted {
+				return // tie-break loser: conn deliberately closed in adoptHelloPeer
+			}
+			fromNodeID = nodeID
+			joined = true
+			connOwned = false // PeerConn now owns conn
+			continue
+		}
+		// Reject any frame that arrives before a join/hello handshake completes.
 		if !joined {
 			log.Printf("[cluster] unexpected frame type 0x%02x from %s before join; closing", frame.Type, fromNodeID)
 			return
--- a/internal/replication/conn_manager.go
+++ b/internal/replication/conn_manager.go
@@ -61,6 +61,20 @@ func (m *ConnManager) AddPeer(nodeID, addr string) {
 	m.peers[nodeID] = NewPeerConn(nodeID, addr)
 }
 
+// HasLivePeerAt reports whether any registered peer with this advertised address
+// currently has a live connection — used by the discovery loop to skip seeds
+// already covered by a join or hello conn (#522 Step 4).
+func (m *ConnManager) HasLivePeerAt(addr string) bool {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+	for _, p := range m.peers {
+		if p.addr == addr && p.conn != nil && !p.closed {
+			return true
+		}
+	}
+	return false
+}
+
 // UpdatePeerAddr updates a peer's advertised address WITHOUT disturbing its live
 // connection (adds a disconnected peer if none exists). Used for address-change
 // gossip — the previous path (AddPeer) closed the live conn and tore down the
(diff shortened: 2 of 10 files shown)


scrypster:
## Independent adversarial review → fix-first item applied; safety properties confirmed

The reviewer confirmed all three make-or-break safety properties hold:
- **Tie-break provably converges** — canonical = lower-id's outbound is asymmetric per pair, so no mutual-eviction cycle; non-canonical never evicts a live conn; live join never evicted by hello. No flap storm.
- **Equal-epoch tie-break can't leave two leaders or loop** — lower-id keeps + re-asserts, higher-id falls through to demotion and (now a follower) never re-asserts; 3 primaries → lowest wins; a strictly-higher epoch always wins regardless of id.
- **No unauthenticated registration** — `verifyHello` runs before reply and before any MSP/voter mutation; epoch in a hello is informational only (never `ForceSet`, so a stale-high epoch can't fence the cluster); observer peers aren't registered as voters.

It caught one real regression-risk: a Lobe re-dialing its Cortex via hello every ~1s when advertise_addr ≠ seed string (DNS vs pod IP) — a churn storm (replication unaffected; the join conn is protected). The two-primary Docker proof used symmetric addressing so it wasn't exercised. **Fixed** in the latest commit: `helloDial` returns the discovered node-id and the loop skips a seed whose node already has a live conn — at most one hello-dial per seed at startup, then skip. Full suite green under `-race`.

Issue #531, opened 2026-06-11
cluster: a returning designated-primary retakes leadership after failover (flap + possible data loss)


scrypster:
Surfaced by the Step 4c automatic-failover Docker test (#522). Now that failover works (kill Cortex → a Lobe is promoted), a new edge appears when the **old Cortex returns**.

## Observed
1 cortex + 2 lobes. Kill cortex → lobe2 promoted to Cortex at epoch 2, lobe1 follows, writes replicate. ✅
Restart the old cortex (role=primary) → it bootstrap-asserts at epoch 2 and, being the lower node-id, **wins the equal-epoch tie-break** → lobe2 yields → the old cortex is leader again.

The cluster converges to **exactly one leader** (no persistent dual-primary — the tie-break + epoch fencing hold). But:
- It's a **leadership flap** (the designated primary retakes from the healthy failover leader). Fable Risk 6.
- Writes accepted by lobe2 **during the failover window can be lost** when the old cortex retakes without reconciling/snapshotting from lobe2. Fable Risk 4.

## Fix (Fable §5.4 hardening)
Harden `bootstrapPromoteIfDesignatedPrimary`: a returning designated primary should **not** assert if it learns (via the hello mesh / a CortexClaim) that a leader already exists at epoch ≥ its own — it should follow instead. This requires letting the discovery mesh connect + observe any existing leader **before** the bootstrap-assert (a short bounded wait, or assert-then-yield already happens but loses data). Pin "exactly one leader" (already holds); the goal is to avoid the flap and the data-loss-on-retake.

Lower priority than the failover itself — automatic failover (the headline) works. This is the returning-primary refinement.

Refs #522.

scrypster:
## Investigation (attempted fix reverted — the problem is deeper than scoped)

I tried the bootstrapPromote hardening (have a returning designated-primary consult the discovery mesh for an existing leader before asserting, and defer to it). It did **not** hold up under Docker validation, and the investigation surfaced a more fundamental gap. Reverted rather than ship an unvalidated partial fix.

### What the Docker run showed
1 cortex + 2 lobes (all-to-all seeds). Kill the cortex, then restart it. Observed in the logs:
[...]

PR #532, opened 2026-06-11
feat(cluster): SDOWN gossip + failover election — automatic failover works (#522 Step 4c/d, closes #528 #529)

### LINKED ISSUE #528: cluster: ODOWN never fires in a ≥3-voter cluster — automatic failover cannot trigger (no SDOWN gossip)

scrypster:
Step 4 (c)+(d) — the automatic-failover finale of the cluster liveness overhaul (epic #522). Builds on Step 4 a+b (#530).

## Problem (#528)
ODOWN could never fire in a ≥2-voter cluster: `votedDown` was never populated and `TypeSDown` was never sent or handled, so a node only ever had its own subjective SDOWN (1 vote < quorum). **Automatic failover never completed** — killing the Cortex left the Lobes waiting forever. (Masked until now because liveness itself was broken before Steps 1–3.)

## Fix
**(c) SDOWN gossip + failover election:**
- MSP broadcasts `TypeSDown{sender,target}` on a non-observer peer's SDOWN transition. `RecordDownVote` accumulates votes in `votedDown` and fires `OnODown` **once** (`odownFired` latch) when this node *also* sees the target SDOWN and votes reach the non-observer quorum. `HandleIncomingFrame` routes `TypeSDown` → `RecordDownVote`. Recovery clears the latch + votes.
- **Election jitter:** `OnODown` starts the election after a deterministic per-node delay (0–1 heartbeat, FNV of node-id) so two Lobes don't split the vote, with a bounded retry (`ResetCandidate` + re-`StartElection`) while no leader emerges.

**(d)** `JoinRequest` carries the joiner's role; `HandleIncomingJoin` registers voters by that role (0 = legacy = replica) so a joining Observer is not a voter (#529, the other half — the hello path was gated in a+b).

## Proven in Docker (the failover that never worked)
1 cortex + 2 lobes, all-to-all seeds (so the lobes form a hello-mesh vote channel):
- **Healthy**: all-to-all mesh, lobe1 sees lobe2 in members, `alive=3`.
- **Kill the Cortex** → `lobe2: ODOWN detected → promoted to Cortex epoch=2`; lobe1 follows lobe2 (`cortex_id=lobe2`); the new leader **accepts writes and replicates** to lobe1. ✅
- **Restart the old Cortex** → cluster converges to **exactly one leader** (the returning designated-primary retakes via the equal-epoch tie-break; lobe2 yields).

## Tests
`RecordDownVote` fires ODOWN once at quorum; no ODOWN without local SDOWN agreement. Full `internal/replication` + `mbp` suites green under `-race`.

## Known edge (documented, not a regression) → #531
The returning designated-primary **retakes** leadership after failover (a flap that converges to one leader), and writes accepted during the failover window can be lost on retake (Fable Risk 4/6). Tracked in **#531** (harden `bootstrapPromoteIfDesignatedPrimary` to follow an existing leader instead of asserting). Before this PR failover didn't happen at all, so this edge was unreachable.

Closes #528, #529. Refs #522.

### DIFF
--- a/internal/replication/coordinator.go
+++ b/internal/replication/coordinator.go
@@ -100,6 +100,9 @@ type ClusterCoordinator struct {
 
 	// advertiseAddr is this node's routable address (used in PeerHello, #522 Step 4).
 	advertiseAddr string
+	// electing single-flights the ODOWN-triggered failover election driver so
+	// concurrent ODOWN events don't spawn racing election loops (#522 Step 4c).
+	electing atomic.Bool
 
 	// Per-Lobe streamers (Cortex only): lobeID -> cancel func
 	streamers   map[string]context.CancelFunc
@@ -704,7 +707,10 @@ func (c *ClusterCoordinator) reconcileJoinedPeer(nodeID, addr string, role NodeR
 		}
 	}
 	retire := matched
-	if retire == "" {
+	if retire == "" && isVoter {
+		// Only a voter join may retire an arbitrary (non-address-matching) seed —
+		// an observer that matches no seed must not retire and unregister an
+		// unrelated voter placeholder (#529).
 		retire = anySeed
 		if retire != "" {
 			slog.Warn("cluster: joined peer address did not match any seed; retiring a seed placeholder to conserve quorum (set advertise_addr to match the seed addresses)",
@@ -714,9 +720,10 @@ func (c *ClusterCoordinator) reconcileJoinedPeer(nodeID, addr string, role NodeR
 	if retire != "" {
 		c.msp.RemovePeer(retire)
 		c.mgr.RemovePeer(retire)
-		if isVoter {
-			c.election.UnregisterVoter(retire)
-		}
+		// A seed placeholder is always registered as a voter at startup, so a
+		// retired seed must always be unregistered — even on an observer join,
+		// where the "expected voter" turned out to be a non-voting observer.
+		c.election.UnregisterVoter(retire)
 	}
 }
 
@@ -1273,7 +1280,10 @@ func (c *ClusterCoordinator) HandleIncomingFrame(fromNodeID string, frameType ui
 		if err := msgpack.Unmarshal(payload, &msg); err != nil {
 			return fmt.Errorf("unmarshal SDownNotification: %w", err)
 		}
-		c.msp.RecordDownVote(msg.SenderID, msg.TargetID)
+		// Key the vote by the authenticated frame source, not the payload's
+		// SenderID — otherwise one peer could forge many sender ids to drive ODOWN
+		// unilaterally (#522 Step 4c). Gossip is one-hop, so fromNodeID is the voter.
+		c.msp.RecordDownVote(fromNodeID, msg.TargetID)
 		return nil
 
 	case mbp.TypeCogForward:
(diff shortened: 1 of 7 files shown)


scrypster:
## Independent adversarial review → fix-first items applied

The reviewer confirmed the make-or-break safety properties hold: the ODOWN latch fires at most once per episode (atomic check+set under the lock in both the tick and gossip paths) and re-arms cleanly on recovery; the election epoch-storm is bounded (≤4 bumps/episode, latched); `ResetCandidate` can't abandon a won election; the supervisor handoff (ODOWN→elect→promote→`followTerm` returns `modeLeading`) is clean; and legacy joiners (role 0) stay voters.

It caught one spurious-failover defect + hardening, all fixed:
- **Observer over-count (make-or-break)** — an observer's gossiped vote counted toward the ODOWN quorum, whose denominator excludes observers, so `cortex + 1 lobe + 1 observer` could fail over a healthy Cortex on one real voter. `RecordDownVote` now rejects observer senders. (New test added.)
- **Vote auth** — keyed the vote by the payload `SenderID`; a peer could forge ids to drive ODOWN. Now keyed by the authenticated frame source.
- **Single-flight** the ODOWN election driver.
- **Phantom voter** — an observer join no longer leaves a retired seed registered as a voter.

Minor deferred: the join HMAC covers node-id but not the new `Role` field (a role-flip needs a MITM already inside the secret-authenticated cluster network — out of the current threat model; changing the join MAC is rolling-upgrade-sensitive). Tracked for a future hardening pass. Suite green under `-race`.

Issue #534, opened 2026-06-11
cluster: dead PeerConns are never evicted — stale conns block rejoin, heartbeats, discovery, claim-on-connect


scrypster:
Found by Fable in the #531 design pass.

When an inbound conn's read loop exits post-handshake, handleClusterConn deliberately does NOT close the PeerConn (connOwned=false) and nothing removes it from ConnManager. Same for the hello-dial reader. PeerConn.Send errors are swallowed everywhere and don't mark the conn dead. So a killed peer's PeerConn reports `IsConnected()==true` forever, which:
- makes RegisterConnKind reject a restarted peer's new hello conn (live kindJoin never evicted by hello),
- makes HasLivePeerAt stop runPeerDiscovery re-dialing that seed,
- kills claim-on-connect to the restarted peer,
- wastes heartbeats on a zombie conn.

This is the stale-conn failure that sank the reverted #531 attempt.

## Fix (PR2, independent + small)
- `ConnManager.EvictIfConn(nodeID, conn)` — close+mark-dead if the entry still wraps this exact conn; call it on read-loop exit in handleClusterConn and readPeerFrames.
- `PeerConn.Send` marks the conn dead (closed) on write error.
- TCP keepalive on conn adoption (cheap third line; don't depend on it).

Invariant: a PeerConn whose TCP conn errored reports IsConnected()==false within one frame-read or one bounded (≤5s) write.

Issue #533, opened 2026-06-11
cluster: HandleJoinRequest has no leadership check — any node accepts joins and calls itself Cortex (split-brain root)


scrypster:
Found by Fable in the #531/failover-robustness design pass.

`JoinHandler.HandleJoinRequest` (join.go) accepts a JoinRequest and replies `Accepted: true, CortexID: <self>` with **no check that this node is actually the leader** — the only gate is `epoch == 0`. `HandleIncomingJoin` doesn't gate either.

This is the direct cause of the observed failover chaos: when the Cortex dies, each lobe's followTerm→streamReplication hits a read error, rejoins via joinWithRetry cycling seeds, dials the *other lobe*, and is accepted at epoch 1 — so `lobe1 joined cortex=lobe2` AND `lobe2 joined cortex=lobe1`, with no stable leader. It also mislabels the other lobe as RolePrimary in MSP, and the mutual join/rejoin churn tears down the very conns carrying VoteRequest/VoteResponse.

## Fix (part of failover-robustness PR1)
Leader-gate joins (no wire change — JoinResponse already has CortexID/CortexAddr/RejectReason):
- JoinHandler gets a `LeaderInfo` callback (IsLeader / CurrentLeader / leader addr).
- HandleJoinRequest: if not leader → `Accepted:false, RejectReason:"not cortex"`, redirect via CortexID/CortexAddr to the known leader.
- joinWithRetry: on a redirect, try the advertised leader addr next.

Invariant: a `JoinResponse{Accepted:true}` is only ever produced by the current leader.

scrypster:
Resolved by PR1 (#535): HandleIncomingJoin now runs the leadership gate BEFORE registering the conn — a non-leader writes a redirect on the raw conn and returns adopted=false without registering, so it can't evict the live lobe↔lobe hello conn. Invariant: a JoinResponse{Accepted:true} is only ever produced by the current leader. Docker-proven: lobes correctly reject each other (no more mutual-join), and failover converges 8/8.

PR #535, opened 2026-06-11
fix(cluster): convergent failover election + leader-gated joins (#531 PR1, #533)

scrypster:
PR1 of the failover-robustness sequence (epic #522 follow-up). Makes automatic failover converge to exactly one leader every time, and fixes the split-brain root cause #533.

## Problems
1. **Split-vote, no stable leader.** Two lobes reaching ODOWN together each self-vote at the same epoch (votedFor immutability — correct for safety), so they split 1-1; the deterministic FNV jitter + fixed 4×(reset+StartElection+sleep) kept them phase-locked, leapfrogging epochs in lockstep, then gave up → permanently leaderless.
2. **#533 — any node accepted joins and called itself Cortex.** So the lobes mutually joined each other (lobe1→lobe2 AND lobe2→lobe1), and the join churn tore down the conns carrying the vote frames.

## Fix
**Convergent election:**
- `startElectionWithJitter`: randomized `U[0,hb)` initial + re-randomized `U[3hb,6hb)` per round, **unbounded** while leaderless, gated on `aliveVoters≥quorum` and ctx. Re-randomization makes each round's split probability independent → converges w.p. 1, expected ~1 round. Safety (≤1 leader/epoch) untouched.
- `ClearLeader(deadID)` on the leader's ODOWN; `OnODown` only elects when the dead node was the (only/unknown) leader — a lobe death under a healthy leader no longer triggers an election.
- Unify the two election drivers (`waitQuorumTerm` yields while the ODOWN driver holds `electing`).
- Broadcast a CortexClaim on bootstrap-assert + crash-recovery (only `tryPromote` did before).

**#533 leader-gated joins:**
- `HandleIncomingJoin` runs the leadership gate **before** `RegisterConn` (a non-leader writes a redirect on the raw conn, returns `adopted=false`, never registers — so it can't evict the live lobe↔lobe hello conn that carries SDOWN gossip, which was silently killing failover). `joinWithRetry` follows the redirect.
- On a successful join, a Lobe adopts the JoinResponse's CortexID/Epoch as its leader (reliable discovery — the claim broadcast otherwise races the join).

Invariant: a `JoinResponse{Accepted:true}` is only ever produced by the current leader.

## Validated in Docker — 8/8 convergence
1 cortex + 2 lobes, all-to-all seeds, **8 fresh kill-Cortex cycles**: every one produced **exactly one leader** with the follower pointing at it (no split, no mutual-join, no confusion). Before this PR the same loop produced 0 stable leaders.

## Tests
ClearLeader unblocks re-election + won't clear a live leader; non-leader join rejected with redirect; leader accepts. Full `internal/replication` suite green under `-race`.

Refs #531, #533. PR2 (#534 stale-conn eviction) and PR3 (#531 returning-primary deference) follow.

### DIFF
--- a/cmd/muninn/server.go
+++ b/cmd/muninn/server.go
@@ -680,9 +680,14 @@ func handleClusterConn(conn net.Conn, coord *replication.ClusterCoordinator) {
 			return // connection closed or error
 		}
 		if frame.Type == mbp.TypeJoinRequest {
-			nodeID, err := coord.HandleIncomingJoin(conn, frame.Payload)
+			nodeID, adopted, err := coord.HandleIncomingJoin(conn, frame.Payload)
 			if err != nil {
 				log.Printf("[cluster] join error from %s: %v", fromNodeID, err)
+				return
+			}
+			if !adopted {
+				// Non-leader rejected the join with a redirect (already written to
+				// the raw conn). The conn was NOT registered, so close it here.
 				return
 			}
 			fromNodeID = nodeID
--- a/internal/replication/coordinator.go
+++ b/internal/replication/coordinator.go
@@ -103,6 +103,9 @@ type ClusterCoordinator struct {
 	// electing single-flights the ODOWN-triggered failover election driver so
 	// concurrent ODOWN events don't spawn racing election loops (#522 Step 4c).
 	electing atomic.Bool
+	// runCtx is the lifetime context captured at Run() start, so callbacks wired
+	// before Run (e.g. OnODown) can launch ctx-aware loops (#531 PR1).
+	runCtx context.Context
 
 	// Per-Lobe streamers (Cortex only): lobeID -> cancel func
 	streamers   map[string]context.CancelFunc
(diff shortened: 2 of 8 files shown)


scrypster:
## Independent adversarial review → SHIP

The reviewer built + vetted + ran the `-race` suite (green) and confirmed all four make-or-break invariants:
- **Election safety (≤1 leader/epoch) unchanged** — `ClearLeader` never touches `votedFor`; safety rests on votedFor-immutability + the epoch/quorum CAS, and a fresh election always runs at a new epoch. All state transitions serialize on `election.mu`; no "follows X and elects self at the same epoch" race.
- **No hot-spin / bounded epoch burn** — every loop iteration ends in a non-zero `sleepCtx`; the `aliveVoters≥Quorum` gate prevents epoch burn under partition; the `electing` single-flight re-arms cleanly; ctx cancellation exits cleanly (runCtx is set before the MSP goroutine that fires OnODown).
- **Gate-before-register is leak-free** — the non-leader path writes the redirect on the raw conn and returns `adopted=false` without registering, so the single deferred `conn.Close()` runs exactly once; can't evict the live hello conn.
- **Redirect is attempt-bounded** — the redirect `continue` flows through `attempt++`, hard-capped at 10; falls back to seed cycling; version-rejections (empty CortexAddr) correctly don't redirect.

Fixed the one cosmetic note (the doc comment said "bounded" — the loop is intentionally unbounded-while-leaderless). The other note (unbounded epoch growth under pathological persistent flapping) is the intended w.p.1-convergence design, rate-limited, not a fix.

PR #537, opened 2026-06-12
fix(cluster): returning designated-primary defers to the failover leader (#531 PR3)

### LINKED ISSUE #531: cluster: a returning designated-primary retakes leadership after failover (flap + possible data loss)

scrypster:
PR3 (final) of the failover-robustness sequence. **Closes #531.** Builds on PR1 (#535) + PR2 (#536).

## Problem
After an automatic failover, a restarted `role=primary` node retook leadership — it bootstrap-asserted before discovering the existing leader, then won by epoch/tie-break, flapping the cluster and discarding writes the failover leader accepted while it was down.

## Fix — probe-before-assert
A designated primary with seeds **probes for the current leader before starting an election** (so it never bumps its epoch and becomes competitive):
- **`JoinRequest.Probe`** — a side-effect-free leader-discovery request. `HandleIncomingJoin` answers who the leader is (Accepted+self if leader, else a redirect) on the raw conn **without** registering, adding a member, or snapshotting.
- **`probeForLeader`** — up to 3 passes over seeds, follows a redirect to the named leader for its authoritative epoch, returns early if no seed is reachable (fresh cluster → assert fast). Reliable: synchronous request/response, immune to the stale-conn / claim-timing that sank the first attempt.
- **`runAsCortex`**: a leader at epoch ≥ ours → adopt its claim + `supervise(modeFollowing)` (join + snapshot reconcile our data; no assert, no flap, report `role=replica`). Peers ahead of our epoch but no leader → elect normally, **never** force-assert into someone else's term. Fresh cluster / full restart → assert as before (designated-primary semantics preserved).
- **`WipeForResnapshot`** now preserves the `cluster_epoch`/`node_role` meta (0x19 0x03 prefix) — wiping it would make a re-snapshotting node restart at epoch 0.

## Proven in Docker — the full cycle
1 cortex + 2 lobes, all-to-all seeds:
- **Fresh cold-start** → probe finds no leader → cortex asserts (members=3).
- **Kill the Cortex** → lobe2 leads; write `fo-write` to lobe2.
- **Restart the old Cortex** → logs `deferring instead of retaking`; reports `is_leader=False, cortex_id=lobe2` (**follows**, doesn't retake); **exactly one leader**.
- **No data loss** → the returned Cortex **has `fo-write`** (received via snapshot).

## Tests
A probe returns the leader without registering the prober; `WipeForResnapshot` preserves the epoch. Full `internal/replication` suite green under `-race`.

Closes #531. Completes the failover-robustness sequence (#535, #536, this).

### DIFF
--- a/internal/replication/coordinator.go
+++ b/internal/replication/coordinator.go
@@ -484,6 +484,9 @@ func (c *ClusterCoordinator) runAsCortex(ctx context.Context) error {
 		if found && ldrEpoch >= currentEpoch {
 			slog.Info("cluster: existing cortex found at startup; deferring instead of retaking leadership",
 				"node", c.cfg.NodeID, "cortex", ldrID, "epoch", ldrEpoch)
+			c.roleMu.Lock()
+			c.role = RoleReplica // we're following, not leading — report as a replica
+			c.roleMu.Unlock()
 			c.election.HandleCortexClaim(mbp.CortexClaim{
 				CortexID: ldrID, CortexAddr: ldrAddr, Epoch: ldrEpoch, FencingToken: ldrEpoch,
 			})
--- a/internal/replication/coordinator.go
+++ b/internal/replication/coordinator.go
@@ -1302,6 +1302,11 @@ func (c *ClusterCoordinator) HandleIncomingJoin(conn net.Conn, payload []byte) (
 	// a snapshot. A returning designated primary uses this to find the current
 	// leader before deciding whether to assert or defer.
 	if req.Probe {
+		// Authenticate the probe so an unauthenticated party can't learn cluster
+		// topology (parity with the normal join's secret check, #531 PR3 review).
+		if !c.joinHandler.ValidSecret(req.NodeID, req.SecretHash) {
+			return req.NodeID, false, nil
+		}
 		isLeader := c.IsLeader()
 		ldrID, ldrAddr := c.cfg.NodeID, c.advertiseAddr
 		if !isLeader {
--- a/internal/replication/coordinator.go
+++ b/internal/replication/coordinator.go
@@ -473,6 +473,35 @@ func (c *ClusterCoordinator) runAsCortex(ctx context.Context) error {
 		return c.supervise(ctx, modeLeading)
 	}
 
+	// A designated primary with seeds first probes for an existing leader BEFORE
+	// starting an election — a failover may have promoted another node while we
+	// were down. If a leader at epoch ≥ ours exists, follow it (no epoch bump, no
+	// assert) so we don't flap the cluster or lose the failover leader's writes
+	// (#531). The probe is reliable (synchronous request/response, immune to stale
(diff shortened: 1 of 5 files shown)


scrypster:
## Independent adversarial review → fix-first blocker applied

The reviewer cleared the make-or-break items (probe is side-effect-free + leak-free; fresh-cluster/full-restart still assert; the deferred node's epoch ends up correct; no regression) and **empirically found one data-correctness blocker**:

- **`WipeForResnapshot` prefix-skip was too broad.** `0x19` is overloaded — idempotency receipts are `0x19|siphash(op_id)`, and the reviewer measured ~0.4% begin `0x19 0x03`. Such a receipt survived the wipe, so a later retry could return a phantom `engram_id` from the discarded timeline. **Fixed**: preserve only the *exact* `clusterEpochKey()`; everything else (incl. `node_role` and any prefix-colliding receipt) is wiped and restreamed by the snapshot. New test asserts a `0x19 0x03…` non-epoch key is wiped. (The reviewer also noted the preservation is partly redundant — the snapshot streams the leader's epoch and the join `ForceSet`s it — so exact-key is pure defense-in-depth with no downside.)
- **Probes now validate the cluster secret** before answering, closing the info-disclosure asymmetry vs. the normal join path.

Full suite green under `-race`. The Docker returning-primary cycle (defer, no flap, no data loss) is unaffected — the fix only narrows what's preserved.

Issue #499, opened 2026-06-10
HNSW restore path: LoadFromPebble scans all vaults' keys, a failed load is cached as a permanently empty index, and load outcomes are never logged


johanneshauer:
## Summary

Three related defects in the per-vault HNSW index restore path (`internal/index/hnsw/`). All line numbers below refer to the **v0.6.1** tag.

1. **`Index.LoadFromPebble` has no vault scoping.** It iterates the entire `0x07` keyspace (`LowerBound: {0x07}`, `UpperBound: {0x08}`) and never compares the key's workspace prefix (`key[1:9]`) against the index's own `idx.ws`. Every per-vault index therefore absorbs **all** vaults' nodes and vectors on restore.
2. **`Registry.getOrCreate` caches the index even when `LoadFromPebble` fails.** A single transient Pebble iterator error at first access pins an **empty** index for the remaining process lifetime. Inserts still land in the empty graph, so semantic search silently covers only engrams embedded after that point — everything embedded before the restart stays invisible while the persisted `0x07` data is fully intact on disk.
3. **No eager load at startup, and no logging of load outcomes.** The registry is lazy-loading only; a successful restore logs nothing (no node/vector counts). Combined with (2), the only trace of a failed restore is one `slog.Error` line emitted whenever the first access happens to occur — there is no startup-time signal an operator can check.

We hit exactly this symptom shape on a production v0.6.1 instance: after a process restart, every engram embedded before the restart disappeared from semantic recall, while the persisted `0x07` data — neighbor lists and `0xFF` vector slots — was verified intact on disk and all health surfaces stayed green. We initially attributed it to failure mode (2); the field-report addendum at the bottom traces the actual chain (a self-looped entry point faithfully restored from stale persisted lists — two additional defects in the same path). The fixes below address both.

## Environment

- **Version:** v0.6.1 (built from source)
- **Platform:** Linux x86_64
- **Multiple vaults** on one instance

## Defect 1: `LoadFromPebble` loads every vault's nodes into every per-vault index

`internal/index/hnsw/hnsw.go:643-695`:

```go
func (idx *Index) LoadFromPebble() error {
	lowerBound := []byte{0x07}
	upperBound := []byte{0x08}
	...
	for iter.First(); iter.Valid(); iter.Next() {
		key := iter.Key()
		if len(key) < 26 {
			continue
		}

		var id [16]byte
		copy(id[:], key[9:25])
		...
```

[...]

PR #545, opened 2026-06-15
fix(hnsw): rebuild graph from vectors when the restored structure is disconnected

### LINKED ISSUE #544: HNSW: a structurally-broken persisted graph (disconnected / self-looped entry point) is restored as-is — LoadFromPebble should BFS-check connectivity and rebuild from vectors
## Context — this is the un-acted-upon half of #499

#499 reported three restore-path defects and, in its field-report addendum,
traced the full production failure chain and offered PRs for **four** fixes.
PR #506 landed three of them — vault-scoped iterator, no-cache-on-load-error,
and load-outcome logging — thank you. This issue tracks the remaining one from
that addendum, which is the only one that actually *repairs* a vault whose
on-disk graph is already broken:

> `LoadFromPebble`: connectivity BFS from the restored entry point; **rebuild
> the graph from the (reliable) vectors when fewer than half the embedded nodes
> are reachable**, then re-persist healthy lists.

We still run this in production (it is what brought our affected vault back),
and the PR is ready and rebased on `v0.7.0`. Filing it as its own issue so the
fix has a home, since #499 is closed.

## Why #471 and #506 don't cover this

- **#471** (four insert-path graph-integrity defects, in v0.6.2) prevents the
  corruption from *forming* on new inserts — e.g. entry-point promotion before
  linking. It does nothing for a graph that is **already** structurally broken
  on disk from history written before the upgrade. Any instance that ran a
  pre-#471 build (notably one that did a bulk `reembed`, which re-inserts every
  existing id) can carry a persisted graph whose apex is a closed loop, and
  #471 will faithfully restore it.
- **#506** helps only when the load **fails**. Here the load **succeeds** — the
  persisted neighbor lists and `0xFF` vector slots are all present and decode
  cleanly. The graph is intact-but-disconnected, so no error is raised, nothing
  is logged as wrong, and FTS masks the dead vector search. `rebuilt=false` is
  logged on a graph that is, in fact, broken.

In other words: #471 stops the bleeding for new writes; nothing on the load
path *heals* a graph that is already bad. For self-hosted users upgrading across
the reembed-era, that healed-only-in-memory entry point is a live data-visibility
bug surviving every restart.

## The failure chain (reproduced end-to-end in production)

Single vault, ~1250 engrams, bge-m3 / 1024-dim via Ollama:

1. A bulk `reembed` re-inserts every existing id. On a pre-#471 `Insert`, a
   re-inserted id that *is* the entry point self-links: its neighbor lists
   become `[self, self]` on every layer.
2. Backlink mutations to existing nodes are made in memory but never
   re-persisted, so the persisted entry point stays `[self, self]` even after
   later inserts heal it in memory.
3. On the next restart, `LoadFromPebble` restores an apex that is a closed loop.
   Greedy descent and the layer-0 beam both die at the entry point ⇒ vector
   search returns ~nothing for the **entire vault**, while vectors, FTS, and
   `status` all look healthy. (Same symptom surface as #462.)

## Proposed fix (what we run)

In `Index.LoadFromPebble`, after the nodes/vectors are read into the temp maps
and `vectorCount` is computed (i.e. immediately before the "apply to index"
block in current `internal/index/hnsw/hnsw.go`):

1. BFS over layer-0 edges from the restored entry point and count reachable
   nodes.
2. If `vectorCount > 1 && reachable*2 < vectorCount` (fewer than half the
   embedded nodes reachable), treat the restored structure as untrustworthy and
   **rebuild from the vectors**, which are reliable (written once, validated on
   read): re-`Insert` every vectored node into a fresh index in deterministic
   ULID order, flush the re-persisted (now-healthy) neighbor lists, and swap the
   rebuilt structure in.
3. Re-run the BFS and log `reachable_from_entry` and `rebuilt` on the existing
   load line so the repair is visible at restore time.

Healthy graphs (the overwhelmingly common case) BFS-reach ~all nodes, so the
rebuild branch is a no-op and adds only one O(N) traversal to load.

### Interaction with #471 (important, and in your favour)

The original v0.6.1 addendum also flagged that re-inserting ~1250 nodes in ULID
order produced a degenerate, early-node-biased graph (defect 4). That
degeneration was a symptom of the **same** insert-path bugs #471 has since
fixed. On `v0.7.0`, the rebuild's re-insertion goes through the corrected
`Insert`, so the rebuilt graph should be well-connected — the regression test
below asserts exactly that (`reachable == vectorCount` after rebuild). If you'd
prefer belt-and-suspenders for any residual structural failure regardless of
graph state, the addendum's other idea — an exact brute-force top-k below a
small node-count threshold (we used 4096) — is a cheap, separate safety net; I'm
happy to file that independently rather than bundle it here.

## Secondary: no eager warm-up means the repair is lazy

`v0.7.0` still has no startup warm-up (the registry is lazy-loading at both
construction sites), so this rebuild — like every load — fires on the first
search for a vault rather than at boot. Functionally fine (the repair still
happens), but it means the only signal is buried at first query time. A 5-line
engine-level `WarmUpHNSW` over `ListVaultNames()` → `ResolveVaultPrefix` →
`getOrCreate` at server start (also offered in #499) surfaces both the load
counts and any rebuild while the operator is still watching the boot log. Happy
to include it in the same PR or split it out — your call.

## Offer

PR is ready, rebased on `v0.7.0`, additive to #471/#506, with regression tests:
one that persists a self-looped entry point and asserts post-load
`reachable == vectorCount` plus a search that returns the true neighbour (dead
before the fix), and one asserting a healthy graph is **not** rebuilt
(`rebuilt=false`) so the common path is untouched. Tell me if you want the
warm-up and/or the brute-force fallback in scope and I'll size it accordingly.

johanneshauer:
Closes #544. Follow-up to #499 — lands the connectivity-rebuild from that
issue's field-report addendum that #506 didn't take. Additive to #471 (insert
prevention) and #506 (scoped load / no-cache-on-error / logging).

## Problem

A persisted HNSW graph can be **intact-but-disconnected**: every node, neighbor
list, and `0xFF` vector slot decodes cleanly, but the entry point's edges are
degenerate (in production: an apex whose every neighbor on every layer was
itself, from a pre-#471 bulk reembed that self-linked the entry point and was
never re-persisted). `LoadFromPebble` restores that structure faithfully, greedy
descent dies at the apex, and vector search returns ~nothing for the whole
vault. The load **succeeds**, so #506's no-cache-on-error never triggers and
`rebuilt=false` is logged on a graph that is in fact dead. FTS masks it; `status`
stays green.

#471 stops new inserts from creating this, but does not heal a graph already
persisted in the broken state — which any instance upgrading across the
reembed era carries on disk.

## Fix

In `Index.LoadFromPebble`, after the temp maps are populated and `vectorCount`
is computed (immediately before the existing "apply to index" block), check
connectivity from the restored entry point and rebuild from the vectors if the
structure is untrustworthy. The vectors are reliable (written once, validated on
read); the neighbor lists are not (written async with `NoSync`, and backlink
mutations are never re-persisted).

```go
// internal/index/hnsw/hnsw.go — after vectorCount is computed, before idx.mu.Lock()

// Persisted neighbor lists are written asynchronously with NoSync and can be
// stale or degenerate (observed in production: an entry point whose every
// neighbor on every layer was itself — one such node at the graph apex makes
// the entire vault unreachable for vector search while the rest of the
// persisted graph looks healthy). Restoring structure we cannot trust is worse
// than rebuilding from the vectors, which ARE reliable. So: check connectivity
// from the entry point, and rebuild from the vectors if the structure is broken.
reachable := bfsReachable(tempNodes, tempEntryPoint)
rebuilt := false
if vectorCount > 1 && reachable*2 < vectorCount {
    slog.Warn("hnsw: restored graph is disconnected — rebuilding from vectors",
        "vault", idx.ws,
        "nodes", len(tempNodes),
        "reachable_from_entry", reachable,
    )
    tmp := New(idx.db, idx.ws)
    tmp.efConstruction = idx.efConstruction
    tmp.efSearch = idx.efSearch
    // Deterministic order: ULIDs ascending (map iteration is random).
    ids := make([][16]byte, 0, len(tempNodes))
    for id, n := range tempNodes {
        if n.vec != nil {
            ids = append(ids, id)
        }
    }
    sort.Slice(ids, func(i, j int) bool { return string(ids[i][:]) < string(ids[j][:]) })
    for _, id := range ids {
        tmp.Insert(id, tempNodes[id].vec)
    }
    tmp.persistWg.Wait() // flush re-persisted (healthy) neighbor lists
    tmp.mu.Lock()
    tempNodes = tmp.nodes
    tempMaxLevel = tmp.maxLevel
    tempEntryPoint = tmp.entryPoint
    tmp.mu.Unlock()
    reachable = bfsReachable(tempNodes, tempEntryPoint)
    rebuilt = true
}
```

The existing load line gains two fields so the repair is visible at restore time:

```go
slog.Info("hnsw: loaded graph from pebble",
    "vault", idx.ws,
    "nodes", len(tempNodes),
    "vectors", vectorCount,
    "duration", time.Since(start),
    "reachable_from_entry", reachable,
    "rebuilt", rebuilt,
)
```

New helper:

```go
// bfsReachable counts nodes reachable from the entry point over layer-0 edges.
func bfsReachable(nodes map[[16]byte]*HNSWNode, entry [16]byte) int {
    if nodes[entry] == nil {
        return 0
    }
    seen := map[[16]byte]bool{entry: true}
    queue := [][16]byte{entry}
    for len(queue) > 0 {
        cur := queue[0]
        queue = queue[1:]
        n := nodes[cur]
        if n == nil || len(n.layers) == 0 {
            continue
        }
        for _, nb := range n.layers[0] {
            if !seen[nb] && nodes[nb] != nil {
                seen[nb] = true
                queue = append(queue, nb)
            }
        }
    }
    return len(seen)
}
```

No new imports — `sort` is already imported by `hnsw.go` in v0.7.0. (Field names
— `persistWg`, `efConstruction`, `efSearch`, `layers`, `vec` — match the current
`Index`/`HNSWNode` defs; the flush is `tmp.persistWg.Wait()`, which is exactly
what `Index.Close()` does, so the throwaway `tmp` leaks nothing.)

## Why rebuild rather than repair in place

The vectors are the ground truth; the edges are the corruptible derived data.
Re-`Insert`ing every vectored node reconstructs correct neighbor lists and, on
`v0.7.0`, runs through #471's fixed `Insert` — so the rebuilt graph is
well-connected (the bulk-reinsertion degeneration seen on v0.6.1 was a symptom
of the insert bugs #471 fixed). The post-rebuild `bfsReachable` re-check, logged
as `reachable_from_entry`, makes that assertion observable rather than assumed.

## Threshold

`reachable*2 < vectorCount` (fewer than half reachable) deliberately fires only
on gross structural failure, not on a few transiently-stranded nodes. A healthy
graph reaches ~all nodes, so the branch is skipped and the cost is one O(N)
layer-0 BFS per load.

## Tests

Both white-box (`package hnsw`), so they call `bfsReachable` directly and inject
corruption through the real key encoding (`keys.HNSWNodeKey` + `encodeNeighbors`):

- `TestLoadFromPebble_DisconnectedEntryPoint_Rebuilds`: build and persist a
  healthy graph, probe-load to find the node the restore path deterministically
  selects as entry point, overwrite *that* node's persisted layer-0 list with
  `[self]`, then reload. Asserts the reload comes back navigable (`bfsReachable`
  jumps from 1 to ~N) and vector search returns a usable result. Fails without
  the rebuild — the load would otherwise restore `reachable == 1`.
- `TestLoadFromPebble_HealthyGraph_LoadsNavigable`: an uncorrupted graph loads
  already-navigable, so the rebuild branch is a no-op and the common path stays
  correct.

Verified run: corrupted reload logs `reachable_from_entry=1 … rebuilt=true`,
healthy reload logs `rebuilt=false`; full `internal/index/hnsw` package passes
under `-race`.

## Scope / non-goals

- No key-format or API changes; load-path only.
- Startup warm-up and the optional small-scale brute-force search fallback (both
  raised in #499) are **out of scope here** to keep this focused on the
  data-visibility repair. Happy to follow up with either on request.

## Risk

Bounded: the rebuild path only runs when the graph is already non-functional for
vector search (`reachable < vectorCount/2`), where doing nothing is strictly
worse. Worst case for a false trigger is one extra in-memory rebuild from
vectors at load — correct, just not free. Healthy graphs never enter it.

### DIFF
--- a/internal/index/hnsw/disconnected_rebuild_test.go
+++ b/internal/index/hnsw/disconnected_rebuild_test.go
@@ -1,0 +1,118 @@
+package hnsw
+
+import (
+	"context"
+	"math/rand"
+	"testing"
+
+	"github.com/cockroachdb/pebble"
+	"github.com/scrypster/muninndb/internal/storage/keys"
+)
+
+// buildPersistedGraph stores n random vectors and inserts them into a fresh
+// Index, then blocks until async persistence completes. It mirrors the
+// production write path (StoreVector persists the 0xFF vector slot; Insert
+// persists the per-layer neighbor lists).
+func buildPersistedGraph(t *testing.T, db *pebble.DB, ws [8]byte, n, dim int, seed int64) (*Index, [][16]byte, [][]float32) {
+	t.Helper()
+	rng := rand.New(rand.NewSource(seed))
+	idx := New(db, ws)
+	ids := make([][16]byte, n)
+	vecs := make([][]float32, n)
+	for i := 0; i < n; i++ {
+		var id [16]byte
+		rng.Read(id[:])
+		v := make([]float32, dim)
+		for j := range v {
+			v[j] = rng.Float32()*2 - 1
+		}
+		ids[i], vecs[i] = id, v
+		if err := idx.StoreVector(id, v); err != nil {
+			t.Fatalf("StoreVector: %v", err)
+		}
+		idx.Insert(id, v)
+	}
+	idx.Close() // wait for async persistNode goroutines
+	return idx, ids, vecs
+}
+
+// TestLoadFromPebble_DisconnectedEntryPoint_Rebuilds reproduces the production
+// failure (an entry point whose persisted neighbor list is only itself) and
+// asserts the load path detects the disconnection and rebuilds the graph from
+// the still-intact vectors. Without the rebuild, bfsReachable from the
+// self-looped entry point is 1 and this test fails.
+func TestLoadFromPebble_DisconnectedEntryPoint_Rebuilds(t *testing.T) {
+	db, err := pebble.Open(t.TempDir(), &pebble.Options{})
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer db.Close()
+
+	var ws [8]byte
+	copy(ws[:], []byte("REBUILD1"))
+	const n, dim = 60, 16
+
+	_, _, vecs := buildPersistedGraph(t, db, ws, n, dim, 11)
+
+	// Discover the node the restore path deterministically selects as entry
+	// point (the highest persisted layer), and confirm the freshly built graph
+	// is navigable — so the only thing that can make the reload navigable again
+	// is the rebuild.
+	probe := New(db, ws)
+	if err := probe.LoadFromPebble(); err != nil {
+		t.Fatalf("probe LoadFromPebble: %v", err)
+	}
+	entry := probe.entryPoint
+	if got := bfsReachable(probe.nodes, entry); got < n*9/10 {
+		t.Fatalf("precondition: freshly built graph not navigable (reachable=%d/%d)", got, n)
+	}
+
+	// Corrupt the persisted entry point's layer-0 neighbor list to [self] — the
+	// degenerate apex observed in production. The 0xFF vector slots are left
+	// intact, so the vectors remain the reliable source of truth for a rebuild.
+	if err := db.Set(keys.HNSWNodeKey(ws, entry, 0), encodeNeighbors([][16]byte{entry}), pebble.Sync); err != nil {
+		t.Fatalf("corrupt entry point: %v", err)
+	}
+
+	idx2 := New(db, ws)
+	if err := idx2.LoadFromPebble(); err != nil {
+		t.Fatalf("LoadFromPebble: %v", err)
+	}
+
+	if got := bfsReachable(idx2.nodes, idx2.entryPoint); got < n*9/10 {
+		t.Fatalf("post-load reachable=%d/%d — disconnected graph was not rebuilt", got, n)
+	}
+
+	res, err := idx2.Search(context.Background(), vecs[0], 5)
+	if err != nil {
+		t.Fatalf("Search: %v", err)
+	}
+	if len(res) == 0 || res[0].Score <= 0 {
+		t.Fatalf("search on rebuilt index returned no usable result: %+v", res)
+	}
+}
+
+// TestLoadFromPebble_HealthyGraph_LoadsNavigable guards the common path: a
+// healthy persisted graph must load already-navigable, so the rebuild branch is
+// a no-op and load remains correct.
+func TestLoadFromPebble_HealthyGraph_LoadsNavigable(t *testing.T) {
+	db, err := pebble.Open(t.TempDir(), &pebble.Options{})
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer db.Close()
+
+	var ws [8]byte
+	copy(ws[:], []byte("HEALTHY1"))
+	const n, dim = 60, 16
+
+	buildPersistedGraph(t, db, ws, n, dim, 23)
+
+	idx2 := New(db, ws)
+	if err := idx2.LoadFromPebble(); err != nil {
+		t.Fatalf("LoadFromPebble: %v", err)
+	}
+	if got := bfsReachable(idx2.nodes, idx2.entryPoint); got < n*9/10 {
+		t.Fatalf("healthy graph loaded with reachable=%d/%d", got, n)
+	}
+}
--- a/internal/index/hnsw/hnsw.go
+++ b/internal/index/hnsw/hnsw.go
@@ -783,6 +783,47 @@ func (idx *Index) LoadFromPebble() error {
 		}
 	}
 
+	// Persisted neighbor lists are written asynchronously with NoSync and can be
+	// stale or degenerate (observed in production: an entry point whose every
+	// neighbor on every layer was itself — one such node at the graph apex makes
+	// the entire vault unreachable for vector search while the rest of the
+	// persisted graph looks healthy). Restoring structure we cannot trust is
+	// worse than rebuilding from the vectors, which ARE reliable (written once,
+	// validated on read). So: check connectivity from the restored entry point
+	// and rebuild the graph in memory from the vectors when the structure is
+	// broken, then re-persist the healthy neighbor lists.
+	reachable := bfsReachable(tempNodes, tempEntryPoint)
+	rebuilt := false
+	if vectorCount > 1 && reachable*2 < vectorCount {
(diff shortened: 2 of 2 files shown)


johanneshauer:
Cross-linking #543 — looks like an independent reproduction of the failure mode this PR targets, at least for the **semantic** half of it.

#543 reports semantic recall collapsing after a cold restart: the graph loads with full counts (`nodes=11016 vectors=11016`), but queries that previously scored ~2.8 return **0 at the default threshold**, with unrelated engrams on top and the near-verbatim target buried at 0.6. The reporter's elimination chain lines up with the mechanism in #544:

- clean SIGTERM → restart triggers it (not a crash/corruption-on-write story);
- rolling the binary back to v0.6.1 on the same data reproduces identically → persisted state, not a version regression;
- cold-booting the **pre-upgrade backup** reproduces too — *"the backup was already in this state; the original process had been masking it with in-memory state since its last reembed."*

That last point is exactly this PR's origin: a graph persisted intact-but-disconnected from a pre-#471 reembed, healthy only in process memory, exposed on the next cold boot. `LoadFromPebble` restores it faithfully, greedy descent dies at the apex, vector search returns ~nothing — and the load still "succeeds", so #506's no-cache-on-error never fires.

Scope caveat so we don't overclaim: #543 also reports things this PR does **not** touch — the ACT-R base-level-activation-lost-on-restart hypothesis, and the `default` vault config / "Vault not found" admin-surface behavior. So this is a candidate fix for the **semantic-recall collapse** in #543, not for all of it.

Easy to confirm: this PR adds `reachable_from_entry` and `rebuilt` to the load log. @scottcrosby-securebine kept the reproducing pre-/post-upgrade backups and offered to run diagnostic builds — a build with this patch against that data dir should log `reachable_from_entry` ≪ `vectors` and `rebuilt=true` if it's the disconnected-graph case. That would settle whether #543 is a superset of #544 or just adjacent.

PR #570, opened 2026-07-01
fix(engine): rarity-weight entity boost, cap accumulation, gate injection on threshold

### LINKED ISSUE #569: Entity boost floods recall results via ubiquitous entities: uncapped, threshold-bypassing, component-less injection
## Summary

`applyEntityBoost` (internal/engine/engine_entity_boost.go) can dominate recall output in mature vaults. Three compounding defects:

1. **Uncapped accumulation through ubiquitous entities.** The boost adds `entityBoostFactor` (0.15) per *(seed, entity)* hit with no per-engram cap and no rarity weighting. In any vault with a hub entity (a person/project mentioned by a large share of engrams), every engram sharing that entity with the top-5 seeds collects up to 5 × 0.15 = **0.75** — rivaling or beating honest semantic scores.

2. **Injection bypasses the caller's threshold and carries no score components.** Engrams not retrieved by the pipeline are appended with `Score = accumulated boost` and an empty `ScoreComponents`, so sub-threshold results appear in output with no audit trail explaining why.

3. **Truncation after the boost re-sort evicts genuine hits.** The call site re-applies `MaxResults` after `applyEntityBoost` re-sorts, so component-less hub injections displace real semantic/FTS matches from the returned set.

## Observed (production vault, ~1.2k engrams)

Query with exactly one true target: rank 1 = true target (0.815, vector 0.691); ranks 2–8 = seven unrelated engrams, **all at exactly 0.75 with empty score components** — each shares one hub entity (mentioned by ~60% of the vault) with all five seeds. Reproduced arithmetic: 5 × 0.15 = 0.75.

- **Rarity-weight** each entity's contribution: `entityBoostFactor × idf(entity)` with `idf = ln(N/df)/ln(N)` clamped to [0,1], `N` = engine engram count, `df` = `EntityRecord.MentionCount` — both existing counters, no scans. Ubiquitous entities → ~0 evidence.
- **Count each distinct entity once per target** (evidence is the shared entity, not the number of seeds carrying it).
- **Cap total boost per engram** (proposed 0.30) so entity evidence cannot outrank content evidence.
- **Gate injection on the caller's threshold** — a result below threshold stays below threshold regardless of how it was found.
- **Add `EntityBoost` to `ScoreComponents`** (activation + mbp + MCP passthrough) so boosted and injected results are auditable; injected results get a populated component trace instead of an empty struct.

Happy to send the PR against `develop` — includes regression tests for the flood, cap, dedupe, threshold gating, and component trace.

isaac-ranger:
Fixes #569.

## What

The post-BFS entity boost could dominate recall output in mature vaults: a flat `entityBoostFactor` (0.15) per *(seed, entity)* hit, uncapped, credited once per seed even for the same entity, and injecting engrams the pipeline never retrieved — with empty `ScoreComponents` and no threshold check. With a hub entity shared by the top-5 seeds, unrelated engrams appear at exactly 5 × 0.15 = 0.75 and evict genuine matches when `MaxResults` truncation re-applies after the boost re-sort (observed in a ~1.2k-engram production vault: 7 of 8 returned results were component-less 0.75 injections).

## How

- **Rarity weighting:** each entity contributes `entityBoostFactor × idf(entity)`, `idf = ln(n/df)/ln(n)` clamped to [0,1] — `n` from the engine engram counter, `df` from `EntityRecord.MentionCount`. Both are maintained counters; no scans added. Near-ubiquitous entities (contribution < 0.001) are skipped *before* their reverse-index fan-out, which also removes the largest scans from the hot path.
- **Once per entity, capped:** a shared entity is credited once per target regardless of seed count (the evidence is the entity, not the seed multiplicity); total per-engram boost clamps at `entityBoostCap` (0.30) so associative evidence can't outrank content evidence.
- **Threshold-gated injection:** engrams outside the scored set are injected only when their entity evidence alone clears the caller's threshold — a result below threshold stays below threshold regardless of how it was found.
- **Auditability:** new `ScoreComponents.EntityBoost` (activation → mbp `entity_boost,omitempty` → MCP `Memory.entity_boost`) populated for boosted and injected results; injected results also carry `Final` and `Confidence`, so component-less scores can no longer appear.

## Behavior notes for review

- The discovery feature is preserved for the case it's actually evidence for: rare shared entities. `TestEntityBoost_SurfacesEntityLinkedEngram` still passes with the new arithmetic.
- Injection now respects the caller's threshold, which the old path bypassed. If always-surface behavior is wanted regardless of threshold, that's a one-line change — but it re-opens the "sub-threshold results in output" contract question, so I kept the strict form.
- `TestEntityBoost_MaxResultsRespectedAfterBoost` (9/9 engrams sharing one entity → df = n → idf 0) now exercises the boost's no-evidence path and still passes: results come from FTS, ordering and truncation unchanged.

## Tests

- `TestEntityBoost_UbiquitousEntityDoesNotFlood` — the #569 regression: 5 seeds sharing a hub entity inject nothing (old behavior: 0.75 injections).
- `TestEntityBoost_SameEntityViaMultipleSeedsCountsOnce` — dedupe across seeds.
- `TestEntityBoost_CapBoundsAccumulation` — clamp at `entityBoostCap`.
- `TestEntityBoost_InjectionRespectsThreshold` — gating both sides of the line.
- Updated `TestEntityBoost_ApplyEntityBoostDirect` for the new signature/arithmetic + component-trace assertions.
- Full `./internal/engine/... ./internal/mcp/... ./internal/transport/... ./internal/replication/...` green.


scrypster (changes requested):
Thank you for this — the mechanism is well thought through: once-per-entity-not-per-seed dedup, the cap so associative evidence can't outrank content evidence, threshold-gating the injection path (closing the old sub-threshold bypass), and giving every boosted/injected result a real component trace so nothing shows up empty-handed anymore. Regression coverage for the flood case, the dedup case, the cap, and the threshold gate is exactly what this needed. We ran the full suite plus gofmt/vet on the branch ourselves — clean.

One thing surfaced during review, verified independently twice: the rarity signal (`idf = ln(n/df)/ln(n)`) is computed from deployment-global counts on both sides, while the #569 problem it's fixing is vault-local.

- `n := e.engramCount.Load()` is the whole store's engram count across every vault (seeded via `store.CountEngrams(ctx)` with no vault arg, incremented/decremented on every write regardless of vault).
- `df` comes from `EntityRecord.MentionCount` via `GetEntityRecord(ctx, entityName)` — `internal/storage/entity.go` documents this as "the global entity record"; entities are deduped and counted by name across the whole store, not per-vault.

Concretely: an entity mentioned in 400 of 500 engrams *within the vault being recalled from* (locally 80% ubiquitous — exactly what should score idf≈0) that happens to be unique to that vault gives idf≈0.036 in a single-vault deployment (n=500). Add nine more unrelated, similarly-sized vaults to the same server (df unchanged, n grows to 5000): idf≈0.297 — over 8x larger for the same locally-ubiquitous entity. The boost grows as unrelated vaults are added, which is backwards — it's the #569 flood reappearing, gated by total deployment size instead of by seed count. It also quietly undercuts the PR's own perf win: a hub that's locally ubiquitous but only moderately common globally clears the noise floor under global idf and runs its full reverse-index scan anyway — the exact scan this PR set out to skip. Every existing test pins `eng.engramCount.Store(N)` to that single test vault's own size, which is exactly why the suite doesn't catch this — none of them share one engine across two vaults.

The good news: the fix for half of this is nearly free. `activateCore` already computes `vaultSize := e.store.GetVaultCount(ctx, wsPrefix)` in the very same function that calls `applyEntityBoost` (internal/engine/engine.go:1919) — swapping `n` to that existing local, rather than the global counter, costs nothing and can't regress single-vault deployments (the two values are equal there). The other half — a true vault-scoped `df` — is a real schema change (per-vault mention counters touching every increment/decrement site on `EntityRecord`), not a quick fix, and we don't think it should block this PR.

Here's why doing the `n` swap alone (without also fixing `df`) is a principled place to stop, not just a compromise: with vault-local `n` and global `df`, an entity unique to the vault gets an exactly-correct idf (the common case, and the one #569 was actually about). An entity shared across vaults gets *under*-credited instead — idf comes out smaller than the true local value, or gets zeroed by the existing `df >= n` guard — which is a false negative (a rare shared entity failing to get credit), never a re-flood. So the residual error after just the `n` fix degrades safely instead of growing with deployment size, unlike today's fully-global version.

So, to merge: could you swap `n` to `e.store.GetVaultCount(ctx, ws)` (or thread the existing `vaultSize` local through), update `entityIDF`'s doc comment (it currently documents the global behavior as the intended design, which needs to change to note the vault-local n / global df split and point at a follow-up), and add a test with two vaults sharing one engine so the suite can actually catch a regression here? We're happy to file the vault-scoped-df follow-up ourselves if useful — just let us know. Thank you again for tracking down and fixing the original flood — this is a real quality-of-recall issue and the core idea here is right.

isaac-ranger:
Extended in 17e4836 per the production report in #569:

- **Injections now gated on the caller's meta filters.** `applyEntityBoost` takes `req.Filters` and applies the same `PassesMetaFilter` predicate phase 6 uses (exported from `activation` for the purpose). Records already in the result set passed phase 6's filter, so only injections needed the gate.
- **`TotalFound` counts injections**, restoring the invariant `total >= len(activations)` — the report's detection fingerprint.
- **Tests:** filter gating at the boost layer directly, plus a full-path regression asserting the fingerprint — including the report's warning that the filter-matches-nothing control passes even unfixed, which the test calls out so it isn't mistaken for coverage. Both new tests were verified to fail with the gate removed before landing.

The report is what surfaced both gaps — the fingerprint made the regression test straightforward. Details in a reply on #569.

isaac-ranger:
Review response in d9c1449 — all three asks, plus an answer to your offer:

**1. `n` is now vault-local.** `applyEntityBoost` takes the vault's engram count as a parameter and `activateCore` threads through the `vaultSize` local it already computes at the top of the function — no second store call, and single-vault deployments are bit-identical (the two values are equal there, as you noted). Your arithmetic held up exactly in our tests: the locally-ubiquitous entity that scores idf 0 against its own vault was scoring ≈0.47 once two dozen unrelated engrams sat in a sibling vault.

**2. `entityIDF`'s doc comment rewritten.** It now documents the vault-local-n / store-global-df split as the intended design, spells out the safe-degradation argument from your review (a cross-vault entity is *under*-credited — a false negative via shrinking idf or the `df >= n` clamp, never a re-flood, and the residual error can't grow with deployment size), and points at the follow-up for true vault-scoped df.

**3. Two-vaults-one-engine test added.** `TestEntityBoost_RarityIsVaultLocal`: vault A holds six engrams all mentioning a hub entity (df = vault size → idf must be 0) plus a locally-rare entity on two of them; vault B holds 24 unrelated engrams on the same engine. The hub must inject nothing and the locally-rare entity must still pay — so the test guards both directions (the regression *and* over-correction). Verified it has teeth: with `n` temporarily reverted to the global counter, it fails on exactly the hub-injection assertion. The existing direct-call tests now pass their corpus size as the `vaultSize` argument instead of pinning the global counter, so no test depends on `engramCount` for boost math anymore.

On your offer to file the vault-scoped-df follow-up: yes please, gladly taken — you have the clearest picture of what the per-vault mention counters should touch on `EntityRecord`. Happy to pick the implementation up once it's specced if that's useful.

`gofmt`/`go vet` clean, full `./internal/engine` suite passes. (17e4836 below this addressed the separate production report from #569 — meta-filter gating on injections and the `TotalFound` fingerprint — in case the two commits' purposes read as tangled.)

scrypster (changes requested):
Okay, this is good work, and the vault-local rarity catch in your third commit is the kind of thing that makes reviewing your PRs easy. You found that computing idf against the deployment-global engram count re-floods a stale entity as unrelated vaults get added to the same server (idf(6,6)=0 but idf(6,30)≈0.47), and fixed it with the vault size you already had. That's the #569 flood caught one level deeper than the original report.

We verified rather than trusted: the new tests won't compile against pre-fix code (old 3-arg applyEntityBoost and flat factor gone), a hard RED, and the boundary tests are real (InjectionRespectsThreshold asserts len 1 just above, len 2 just below). Build, vet, race green, including merged onto current develop (your base is 63 behind but merges clean).

Three things before it merges:

1. COG-18 in invariants.md still describes the flat 0.15 boost this replaces. Update it to the rarity-weighted, capped-at-0.30, threshold-and-filter-gated behavior and reference #569/#570.
2. openapi.yaml's ScoreComponents is missing `entity_boost`. rest.ActivationItem aliases mbp.ActivationItem, so the field is live on /recall and /activate the moment this merges. Add it in transition_boost's style.
3. This one's on us, not you: the change touches mbp/types.go, which makes it Tier 3, so it needs an independent adversarial pass before merge. We'll run that. Flagging it so the beat before landing makes sense, it's process not doubt.

Worth your confirmation, not a required change: the 0.30 cap sits at the low end of the BFS weight range (~0.3 to 0.9) rather than well below it. A rarity-saturated engram can now tie the weakest genuine association matches. The comment says that's deliberate, just confirm. And a heads up: this and #656 both rework applyEntityBoost, so we'd like this larger one to land first and have #656 rebase on top.

scrypster:
Quick follow-up, Cairn. We ran the Tier-3 adversarial pass I mentioned, and it turned up a fourth item that's a real blocker, so better you have it now and fold it into the same pass.

The injection path honors state, the meta filters, and the threshold, but it never checks trust. `applyEntityBoost` isn't passed `ExcludeUntrusted` at all (the call site hands it only `actReq.Threshold` and `actReq.Filters`), and `ExcludeUntrusted` rides the request bool rather than `Filters`, so `PassesMetaFilter` can't see it. Phase 6 enforces it separately at activation/engine.go:1228. So a vault that turns on `ExcludeUntrusted` to keep flagged-unreliable memory out of recall gets it re-injected through any shared rare entity. We RED-proved it: seed A trusted, engram B with Trust=TrustUntrusted sharing a rare entity, idf 1.0, boost 0.15 clears the 0.05 threshold, B comes back. It's live on current develop too, but since this PR is specifically about making injection obey the caller's contract and it already closed the state and filter holes, the trust hole belongs in the same fix.

The fix mirrors what you did for the others: plumb `ExcludeUntrusted` into `applyEntityBoost` and skip injecting `eng.Trust == storage.TrustUntrusted` when it's set. One trap: pass 2b only. Pass 2a boosts engrams the pipeline already returned, which already cleared the trust filter, so don't re-filter those. Add a RED test with `ExcludeUntrusted=true`.

Two smaller ones flagged as document-or-follow-up, not blockers: under RRF fusion the 0.30 cap is calibrated to the ACT-R [0,1] scale, so a boosted engram can enter at ~4 to 7x a genuine RRF content match while the gate goes near-vacuous (COG-6 drops the effective threshold to 0.001), so either scale the boost to the fusion mode or document it. And TotalFound can double-count an engram that scored above threshold but was truncated past MaxResults then re-injected. Both fine as scoped follow-ups.

Separate heads up: reflex (S0-S8) just merged as #685. It added a below-threshold seed filter to `applyEntityBoost` (a tag-filter match admitted below the relevance threshold must not seed spread-activation), so when you rebase you'll hit a real conflict right where your rarity rework lives. Keep all three: your rarity/cap/gate on the boost math, the seed filter that skips sub-threshold seeds, and the ExcludeUntrusted fix above. If that reconcile gets fiddly, say the word and we'll pair on it.

isaac-ranger:
All four items landed in 3129fc4, and our own pre-push adversarial pass extended the round — it found three things worth your eyes, one of which upgrades your trust finding into a class.

**Your items:**

1. **ExcludeUntrusted at injection** — in, mirroring phase 6's hard trust filter, pass 2b only per your trap (pass 2a engrams cleared phase 6's own filter). RED test with an ungated control, plus a real-`Activate` wiring test in the shape you asked #656 to adopt: a gated vault and a positive-control vault, so an unwired call site cannot produce a false green.
2. **COG-18** — rewritten for the rarity-weighted, capped, contract-gated behavior, referencing #569/#570/#654.
3. **openapi ScoreComponents** — `entity_boost` added in `transition_boost`'s style. Swept the SDKs per the drift doc: none of them model score components at all, so no SDK obligation triggers. (Pre-existing note, not this PR's: proto's ScoreComponents lacks `transition_boost` too, and the gRPC adapter drops components entirely.)
4. **0.30 cap** — confirmed deliberate, now stated in the comment: two or more maximally-rare shared entities may tie the *weakest* genuine graph edge, never a mid-strength one. That is the intended ceiling for associative evidence.

**What our adversarial pass added:**

5. **The trust hole generalizes: injections also bypassed the #548 lease-visibility filter.** Same mechanism — lease hiding rides request fields (`CallerOwner`/`IncludeLeased`), not `Filters`, so `PassesMetaFilter` can't see it, and there is no later gate for leases the way COG-19's final gate exists for validity. A work-queue engram checked out under a live foreign lease re-entered recall through any shared rare entity. Rather than leave it to #657's residual list, pass 2b now enforces the full phase-6 result contract: state, meta filters, trust, lease, validity. `applyEntityBoost` takes the `ActivateRequest` so the contract can't drift field-by-field again. Each gate has its own RED-verified test (lease: live-foreign / owner / released / `IncludeLeased` legs; validity: expired / `IncludeInvalid` legs). #657's lease item stays open for its other surfaces; the boost path is closed.
6. **The validity check at injection also keeps `TotalFound` honest** — previously an expired injection was counted, then dropped by the final gate. The comment now also scopes the fingerprint claim to the boost path (supersession injects promoted heads without a bump — pre-existing reflex behavior) and documents the truncated-past-MaxResults-then-reinjected double-count as the remaining follow-up.
7. **The RRF caveat you asked us to document turned out to describe an unreachable configuration.** On the `activateCore` path the request threshold is defaulted to **0.1 before `Run()`**, so Run's 0.001 RRF default never fires there — under RRF defaults nothing clears the seed rule and the boost pass silently no-ops (pre-existing; the seed rule came in with reflex). The hazard is real only with a caller-explicit small threshold, and the cap comment now says exactly that. Two things you may want on a tracker: the RRF+defaults no-op itself, and COG-6's text ("forces threshold to 0.001 when caller threshold ≥ 0.01") no longer matching live code.

Also in the commit: entity reverse-index scans dedup across seeds (credit is once-per-entity per target, so a rescan could never change a score — and hub-adjacent entities were exactly the priciest scans), and one stale "entity-boost bypasses passesMetaFilter" comment in `valid_time_test.go` corrected.

One follow-up we chose not to fold in: pass 2a writes `Score`/`Components` in place and the re-sort reorders the backing array while the activation log drainer may still read it — the same structural race the validity gate's new-slice NOTE guards against, pre-existing on develop's boost but worth a copy-on-entry in a scoped PR. Happy to file it.

Branch is the reconciled two-commit stack on current develop (reflex seed rule preserved verbatim, supersession/validity ordering untouched — 17 of those tests green). Full `./...` suite, vet, gofmt clean; `-race` green on engine and activation.

isaac-ranger:
One follow-up commit, 98513a8: we re-ran our adversarial pass against the fix round itself (the earlier pass predated half of it), and it found one divergence in the lease mirror — on a `GetLease` error the injection path failed OPEN (skip the hide-branch, admit the engram) where phase 6 fails the whole request loudly. A corrupt lease record would have produced a silent admission on one path and a loud error on the other. Now fails closed: a lease-read fault skips the injection — dropping an optional enrichment is the safe degrade. No test rides it (the branch needs a fault-injecting store; the two lines plus comment felt proportionate — say the word if you'd rather have the mock).

Everything else in the round survived the re-review: every RED claim reproduced independently, the wiring test's control leg verified to fail with the whole boost pass disabled, and a sweep of phase 6's candidate-exclusion conditions found all seven mirrored with the one deliberate asymmetry COG-18 already names (no tag-pool threshold bypass for injections).

scrypster (approved):
This is good work, and thanks for folding #656 into it and rebasing onto the stack that just landed on develop. The part we appreciate most is that you wired the injection path through the new validity gate yourself instead of leaving it for us to catch. Injected results now clear the validity and supersession gates, not just the meta filters, which is exactly the composition we needed.

We reviewed it properly. Ran the full gate with -race, then checked each guard with targeted mutation testing, so the rarity weighting, the cap, the threshold gate, and the meta-filter and trust exclusion on injections each redden a test when disabled. Because this touches an MBP wire type and the lease path, we also ran a second independent adversarial pass. That one went straight at the msgpack wire-compat of the new ScoreComponents.EntityBoost field against older peers, and it holds: named-map encoding, additive, omitempty, byte-identical to the old wire when the field is zero. It found no injection ordering bypass either.

One thing we want but will not make you chase: the fail-closed branch on a lease-read error is correct in the code but has no test covering it, since the real store will not error on a missing lease so it needs a fault-injecting fake. We will add that in a small follow-up rather than bounce this back to you again.

Merging now, and closing #656 since this supersedes it. Nice measure-first work on the rarity weighting.

PR #572, opened 2026-07-01
feat(engine,mcp): fuzzy entity resolve behind exact find_by_entity lookup

### LINKED ISSUE #571: find_by_entity is exact-hash only: token variants of a known entity return zero results
## Summary

`FindByEntity` resolves the query through `keys.EntityNameHash` (NFKC + trim + lowercase + SipHash) and scans the 0x23 reverse index. Case and whitespace variants are covered by the normalization, but **any token-level variant of a known entity name silently returns zero results**, because a hash index cannot approximate-match by construction:

- `knock` → 0 results while `The Knock` → hits (article token)
- `dream cycle` vs `dream-cycle` (separator variant)
- `TokenArcade` vs `Token Arcade` (concatenation variant)

All three are real cases from a production vault — the same entity ends up unreachable (or split into duplicate identities) depending on which variant the caller happens to type. The recording seat titles an entity one way; the retrieving seat types it another; they miss each other by an article.

Add a fuzzy resolve layer *behind* the exact lookup — exact stays the fast path and its semantics are unchanged:

1. On zero exact matches, scan the vault's entity names (`ScanVaultEntityNames` — vaults have few distinct entities; this is an in-memory token match, no new index).
2. Tokenize both sides: `keys.NormalizeEntityName`, split on space/hyphen/underscore/slash/dot, drop article stopwords (`the/a/an`).
3. A candidate matches when it shares ≥ 1 non-stopword token with the query, **or** its concatenated token string equals the query's (bridges `TokenArcade` ↔ `Token Arcade`). Rank by Jaccard overlap of token sets; deterministic tie-break.
4. **Auditable resolution, no silent substitution:** the response reports `matched_entity` + `fuzzy: true` + remaining ranked candidates, so the caller always sees which entity actually served the results.

No similarity threshold to tune — the match rule is set-membership, Jaccard is only used for ranking.

Happy to send the PR against `develop` — includes tests for article/separator/concatenation variants, ranking, stopword-only queries, and exact-path behavior preservation.

isaac-ranger:
Fixes #571.

## What

`find_by_entity` resolves through `keys.EntityNameHash` (NFKC + trim + lowercase + SipHash), so token-level variants of a known entity silently return zero results — `knock` misses `The Knock`, `dream-cycle` misses `dream cycle`, `TokenArcade` misses `Token Arcade`. All three are real cases from a production vault. A hash index can't approximate-match by construction, so this adds a fuzzy resolve layer **behind** the exact lookup; the exact path and its semantics are unchanged.

## How

- On zero live exact matches, scan the vault's entity names (`ScanVaultEntityNames`) — an in-memory token match over the few distinct names a vault holds; **no new index**.
- Tokenizer: `keys.NormalizeEntityName` (the same normalization the hash uses), split on space/hyphen/underscore/slash/dot, drop article stopwords (`the/a/an`) with an all-stopword fallback so a name like "The" can still match itself.
- Match rule is **set-membership**: a candidate qualifies on ≥ 1 shared non-stopword token, or when its concatenated token string equals the query's (bridges `TokenArcade` ↔ `Token Arcade`). Jaccard overlap is used **only for ranking** (concatenation equality ranks 1.0), lexicographic tie-break for determinism — no similarity threshold to tune.
- The best-ranked candidate **with live engrams** serves the results (candidates whose engrams are all soft-deleted/archived are passed over).
- **Auditable, never silent:** `FindByEntity` returns `FindByEntityResult{Engrams, MatchedEntity, Fuzzy, Candidates}`; the MCP response adds `matched_entity`, `fuzzy: true`, and `other_candidates` so the caller always sees which entity actually served the lookup. The tool description documents the behavior.

## Tests

- Tokenizer table (articles, separators, all-stopword fallback, whitespace).
- Article variant (`knock` → `The Knock`), separator variant (`dream-cycle` → `dream cycle`), concatenation variant (`TokenArcade` → `Token Arcade`).
- Ranking: full token overlap outranks partial; runner-up reported in `candidates`.
- Phantom-match guards: unrelated tokens return empty/non-fuzzy; a stopword-only query (`the`) does not match every article-bearing entity.
- Existing exact-path tests (archived/soft-deleted exclusion, vault deletion) updated to the result struct and green; MCP handler tests updated incl. a nil-guard for interface implementers.
- Full `./internal/engine/... ./internal/mcp/...` green (10 packages).

## Note for review

`FindByEntity`'s signature changed from `([]*storage.Engram, error)` to `(*FindByEntityResult, error)` — callers inside the repo (MCP interface/adapter/handler) are updated. If you'd rather keep the old signature exported and add a new method, happy to rework.

### DIFF
--- a/internal/engine/engine_entity_resolve.go
+++ b/internal/engine/engine_entity_resolve.go
@@ -1,0 +1,115 @@
+package engine
+
+import (
+	"context"
+	"sort"
+	"strings"
+
+	"github.com/scrypster/muninndb/internal/storage/keys"
+)
+
+// entityResolveMaxCandidates bounds how many ranked fuzzy candidates a
+// resolve returns. The best candidate serves retrieval; the rest are
+// reported to the caller as near-misses.
+const entityResolveMaxCandidates = 5
+
+// entityStopwords are article tokens ignored during fuzzy entity matching —
+// they carry no identity: "The Knock" and "knock" name the same entity
+// (issue #571).
+var entityStopwords = map[string]struct{}{"the": {}, "a": {}, "an": {}}
+
+// entityTokens returns the tokens of an entity name for fuzzy matching:
+// NFKC-normalized, lowercased (keys.NormalizeEntityName — the same
+// normalization the entity hash index uses), split on separator runes, with
+// article stopwords dropped. When a name consists only of stopwords, the raw
+// tokens are returned instead so such a name can still match itself.
+func entityTokens(name string) []string {
+	normalized := keys.NormalizeEntityName(name)
+	raw := strings.FieldsFunc(normalized, func(r rune) bool {
+		switch r {
+		case ' ', '\t', '-', '_', '/', '.':
+			return true
+		}
+		return false
+	})
+	tokens := make([]string, 0, len(raw))
+	for _, tok := range raw {
+		if _, stop := entityStopwords[tok]; !stop {
+			tokens = append(tokens, tok)
+		}
+	}
+	if len(tokens) == 0 {
+		return raw
+	}
+	return tokens
+}
+
+// resolveEntityFuzzy ranks the vault's entity names against the query by
+// token overlap. A candidate qualifies when it shares at least one
+// non-stopword token with the query, or when its concatenated token string
+// equals the query's (bridging concatenation variants such as "TokenArcade"
+// vs "Token Arcade"). Candidates are ranked by Jaccard overlap of token
+// sets — concatenation equality ranks as 1.0 — with a lexicographic
+// tie-break for determinism. The match rule is set-membership; Jaccard is
+// used only for ranking, so there is no similarity threshold to tune.
+//
+// Vaults hold few distinct entity names, so this is an in-memory scan over
+// the 0x20 forward index (ScanVaultEntityNames) — no new index required.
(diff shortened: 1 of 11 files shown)


scrypster (approved):
Thank you for this — clean design, well-tested, and the auditability (matched_entity/fuzzy/other_candidates always reported, never a silent substitution) is exactly right for a feature that resolves ambiguity on the caller's behalf. Ran the full engine/mcp suite plus gofmt/vet on the branch ourselves — clean. No storage-key surface here, so no conflict with the other lease/recall-event PRs in flight.

We had an independent design pass on one judgment call: the match gate is "≥1 shared non-stopword token, no threshold" (by design, per your comment). All three motivating cases (knock/The Knock, dream-cycle/dream cycle, TokenArcade/Token Arcade) actually have 100% token overlap, so a stricter gate would still solve all of them while closing off weak partial-overlap false positives on longer entity names (e.g. a 1-of-3-token match stealing the result from an unrelated entity when it's the only live candidate). The transparency mechanism is real mitigation, but not free — so we're merging this as-is and will follow up with a small tightening PR (token-set containment instead of "any shared token") rather than holding this one up. Thank you again for the fix — this and the retrieval-quality arc around it (#569-574) are landing well.

PR #574, opened 2026-07-01
feat(storage,engine): persist per-recall surfaced sets as recall events

### LINKED ISSUE #573: Recall surfaced sets are discarded — no ground truth for scoring calibration (persist CoActivation surfaced set per recall event)
## Summary

Every recall builds the set of surfaced engrams (the `CoActivationEvent` submitted to the Hebbian worker), but that per-event set is **discarded after aggregation** — the Hebbian worker folds it into pairwise co-activation weights, and its Pebble handle is literally marked `// optional, reserved for future persistence` (internal/cognitive/hebbian.go). The replication coordinator receives the same set as `CoActivationRef`s and also keeps no per-event record.

This throws away the one signal needed to build **self-regenerating ground truth for scoring calibration**:

- **Positives** already exist: `Decide()` links cited memories via decision→evidence edges (recorded, not authored).
- **Negatives** should be *surfaced-for-this-recall minus cited* — computable only if the surfaced set survives the recall.

Without per-event persistence, the only fallback is aggregate co-activation minus evidence edges, which conflates every unrelated retrieval that ever co-surfaced the same memories — a different (contaminated) class, unusable as training negatives.

- Persist the surfaced set at recall time, keyed on a **recall-event ULID** (time-ordered → range scans and time-based pruning come free): `0x29 | ws(8) | eventULID(16)` → msgpack record `{query context, threshold, [engram id + score]}`. The record is what the caller actually saw (post-boost, post-truncation).
- Return the event ULID as the response `query_id` (currently an ephemeral in-process counter `q-<n>`, not unique across restarts — the field is documented as opaque).
- Skip persistence in observe mode (pure read), matching the existing Hebbian-submit guard — calibration probes must not write the signal they read.
- Time-based retention with amortized in-write pruning (pinned constant), plus an explicit prune method.
- Join to decisions needs no schema: the evidence edges `Decide()` already writes bridge event → decision offline.

Happy to send the PR against `develop` — includes storage round-trip/scan/prune tests and engine-level tests for persistence, observe-mode skip, and query_id correlation.

isaac-ranger:
Fixes #573.

## What

Recalls build the surfaced-engram set (the `CoActivationEvent`) and discard it once the Hebbian worker folds it into aggregate pairwise weights — the per-event record dies at the comment `// optional, reserved for future persistence` (hebbian.go). This PR persists that set, keyed per recall event, which completes self-regenerating ground truth for scoring calibration: **positives** = `Decide()`'s decision→evidence edges (already recorded); **negatives** = surfaced-for-this-recall minus cited — now computable by rule, no hand-labeling. (The aggregate-co-activation fallback is not a substitute: it folds in every unrelated retrieval that co-surfaced the same memories.)

## How

- **New `0x29` keyspace:** `0x29 | ws(8) | eventULID(16)` → msgpack `RecallEvent{context, threshold, surfaced_at, [engram id + score]}`. The record is what the caller actually saw (post entity-boost, post truncation). ULID keys sort by event time, so vault scans run in event order and pruning uses the key-embedded timestamp — no value reads, no secondary index.
- **Engine hook** after result assembly: mint an event ULID, persist the set, and return the event ID as the response `query_id`. A later decision can then be joined offline against exactly what its recall surfaced. Persistence is best-effort — a failed write logs a warning and falls back to the counter; a recall never fails on the sink.
- **Observe mode persists nothing**, matching the existing Hebbian-submit guard: a calibration probe must not write the signal it reads.
- **Retention:** pinned 90-day window, amortized pruning (every 256th write per process prunes the writing vault's expired events), plus an explicit `PruneRecallEvents` for direct maintenance.

## Behavior notes for review

- `query_id` was an in-process counter (`q-<n>`), documented as opaque and **not unique across daemon restarts**. Successful non-observe recalls now return the durable event ULID; observe-mode and empty recalls keep the counter. If any consumer depends on the `q-` shape, flag it — I found no such assertion in the repo's tests.
- Write cost is one `pebble.NoSync` set per non-empty recall (~200 bytes typical), consistent with the store's group-commit convention.
- The methods live on `PebbleStore` (the engine holds the concrete type); `EngineStore` is untouched, so existing adapters and fakes are unaffected.

## Tests

- Storage: write/get roundtrip + absent-get `nil,nil`; vault-scoped scan in event-time order; prune boundary via key-embedded ULID timestamps (older deleted, newer survive, count returned).
- Engine: recall persists the event and `query_id` == event ULID with entries matching the response exactly; observe mode persists nothing and keeps the `q-` counter; empty recalls persist nothing.
- Full `./internal/storage/... ./internal/engine/... ./internal/mcp/... ./internal/transport/... ./internal/cognitive/... ./internal/replication/...` green.

Related: #569/#570 (entity-boost flood), #571/#572 (fuzzy entity resolve) — together these are the retrieval-quality arc; this PR is the data-collection half that lets scoring weights be fit against real usage.

### DIFF
--- a/docs/internals/keyspace-registry.md
+++ b/docs/internals/keyspace-registry.md
@@ -67,6 +67,7 @@ prefix — see the vault-reuse note at the bottom).
 | 0x26 | ws+entityHash(8)+engramID | — | rel-entity index |
 | 0x27 | ws | 16B dream state | |
 | 0x28 | ws+sha256(32) | engramID(16) | content-hash dedup |
+| 0x29 | ws+eventULID(16) | msgpack RecallEvent | recall-event calibration record (#573); event-time key order; reads purpose-gated |
 | **0x2A** | ws+ulid | JSON Lease{owner,heartbeat,ttl} | ownership-lease sidecar (advisory) |
 
 ## Auth prefixes (`internal/auth/keys.go`)
@@ -91,7 +92,7 @@ prefix — see the vault-reuse note at the bottom).
 
 ## Free bytes
 
-`0x29`, `0x2B`–`0x3F`, `0x42`+ are free for new storage/auth keys. (`0x29`/`0x40`/`0x41`
+`0x2B`–`0x3F` and `0x46`+ are free for new storage/auth keys. (`0x29`/`0x40`/`0x41`
 also appear in `internal/transport/mbp/frame.go` as **wire opcodes** — a different
 keyspace; coincidental, safe, but confusing. Prefer `0x2B+` for new storage prefixes.)
 
--- a/internal/engine/engine.go
+++ b/internal/engine/engine.go
@@ -2203,6 +2203,34 @@ func (e *Engine) activateCore(ctx context.Context, req *mbp.ActivateRequest, str
 	}
 	wg.Wait()
 
+	// Persist the surfaced set as a recall event (issue #573): the engrams
+	// this recall actually returned (post entity-boost, post truncation),
+	// keyed by a time-ordered event ULID. Skipped in observe mode — a pure
+	// read must not write the signal that calibration reads. On success the
+	// event ID becomes the response query_id, so a later muninn_decide can
+	// be joined offline against exactly what this recall surfaced.
+	queryID := e.fastQueryID()
+	if len(items) > 0 && !auth.ObserveFromContext(ctx) {
+		eventID := storage.NewULID()
+		entries := make([]storage.RecallSurfacedEntry, len(items))
+		for i, item := range items {
+			entries[i] = storage.RecallSurfacedEntry{ID: item.ID, Score: item.Score}
+		}
+		ev := &storage.RecallEvent{
+			Context:    req.Context,
+			Threshold:  float32(actReq.Threshold),
+			SurfacedAt: time.Now().UnixNano(),
+			Entries:    entries,
+		}
+		if err := e.store.WriteRecallEvent(ctx, wsPrefix, eventID, ev); err != nil {
+			// Persistence is best-effort: a failed sink write degrades
+			// calibration ground truth, never the recall itself.
+			slog.Warn("activate: persist recall event failed", "err", err)
+		} else {
+			queryID = eventID.String()
+		}
+	}
+
 	// Submit co-activations to Hebbian worker (skipped in observe mode or when disabled by Plasticity).
 	// On Lobe nodes (hebbianWorker == nil) collect refs for forwarding to Cortex instead.
 	var lobeCoActivations []mbp.CoActivationRef
@@ -2331,7 +2359,7 @@ func (e *Engine) activateCore(ctx context.Context, req *mbp.ActivateRequest, str
 	metrics.ActivateDuration.WithLabelValues(req.Vault).Observe(d.Seconds())
 
 	return &mbp.ActivateResponse{
-		QueryID:     e.fastQueryID(),
+		QueryID:     queryID,
 		TotalFound:  result.TotalFound,
 		Activations: items,
 		LatencyMs:   result.LatencyMs,
--- a/internal/engine/engine_recall_event_test.go
+++ b/internal/engine/engine_recall_event_test.go
@@ -43,7 +43,7 @@ func TestRecall_PersistsRecallEvent(t *testing.T) {
 	require.NoError(t, err, "query_id must be a ULID when the recall event was persisted, got %q", resp.QueryID)
 
 	ws := eng.store.ResolveVaultPrefix(vault)
-	ev, err := eng.store.GetRecallEvent(ctx, ws, eventID)
+	ev, err := eng.store.GetRecallEvent(ctx, ws, eventID, storage.RecallPurposeCalibration)
 	require.NoError(t, err)
 	require.NotNil(t, ev, "recall event must be persisted")
 
@@ -87,7 +87,7 @@ func TestRecall_ObserveModeDoesNotPersist(t *testing.T) {
 
 	ws := eng.store.ResolveVaultPrefix(vault)
 	count := 0
-	require.NoError(t, eng.store.ScanRecallEvents(ctx, ws, func(_ storage.ULID, _ *storage.RecallEvent) error {
+	require.NoError(t, eng.store.ScanRecallEvents(ctx, ws, storage.RecallPurposeCalibration, func(_ storage.ULID, _ *storage.RecallEvent) error {
 		count++
 		return nil
 	}))
@@ -116,7 +116,7 @@ func TestRecall_NoResultsNoEvent(t *testing.T) {
 
 	ws := eng.store.ResolveVaultPrefix(vault)
 	count := 0
-	require.NoError(t, eng.store.ScanRecallEvents(ctx, ws, func(_ storage.ULID, _ *storage.RecallEvent) error {
+	require.NoError(t, eng.store.ScanRecallEvents(ctx, ws, storage.RecallPurposeCalibration, func(_ storage.ULID, _ *storage.RecallEvent) error {
 		count++
 		return nil
 	}))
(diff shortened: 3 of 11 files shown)


isaac-ranger:
Added `0673ed1`: recall-event **reads are now purpose-gated behind a closed allowlist** ({calibration} only). An empty or unknown purpose is refused with a loud error by construction — recall events are calibration instrument data, not a memory surface, and this makes that boundary structural rather than conventional. Every permitted read emits a structured log line; widening the allowlist requires editing `validRecallPurposes`, so any change to who may read this data shows up as a reviewable diff.

This also answers the natural review question "who gets to read these records?" — as of this PR: only code that declares the calibration purpose, and every such read leaves a trail.

scrypster:
Thank you for this — the design is clean (event-time-ordered ULID keys mean vault scans and pruning both fall out for free with no secondary index, and the purpose-gated read allowlist in the second commit is a nice touch for making "nothing may recall from calibration data" enforced rather than a remembered convention).

One thing surfaced during review that needs a small change before merge: #576 (atomic compare-and-set + ownership lease), reviewed independently and in parallel, also claims key prefix \`0x29\` — for its \`LeaseKey\` sidecar, a completely different keyspace than this PR's \`RecallEventKey\`. Both branches forked from \`develop\` before either merged, so this wasn't visible to either of you at the time.

We've asked #576 to bump to \`0x2A\` (it already has two small required fixes to make, so it's a natural place to pick that up too), which leaves \`0x29\` here as-is. No action needed on your end unless #576's author gets there first — just flagging so this doesn't silently collide depending on merge order. Will follow up once #576's fixes land and re-check mergeability here.

scrypster (changes requested):
Thank you for this — really well-built. The event-time-ordered ULID keys giving you free scan-order and no-value-read pruning is elegant, the purpose-gated read allowlist makes "nothing may recall from calibration data" an enforced invariant instead of a remembered convention, and the query_id fallback to the ephemeral counter on write failure (never failing the recall itself) is exactly the right call. We ran the full suite plus `gofmt`/`go vet` on the branch ourselves — clean. Verified the prune boundary math by hand too (the zero-entropy-ULID-at-cutoff trick correctly excludes events at/after the cutoff) — solid.

Two things worth closing before merge, beyond the `0x29` prefix collision with #576 already flagged separately:

1. **`WriteRecallEvent` doesn't replicate.** It's a bare `ps.db.Set(key, val, pebble.NoSync)`, not wrapped in a batch with `ps.replicateBatch(batch)` like nearly everything else in `internal/storage` (association.go, engram.go, entity.go, ordinal.go all follow that pattern). We checked `internal/replication/types.go` and confirmed Lobe replicas explicitly serve reads (`RoleReplica // Lobe — receives replication, serves reads`) — recall is exactly the read traffic that gets routed to Lobes for horizontal scaling. That means a recall served by a Lobe writes its event to that Lobe's local Pebble only; it never reaches Cortex or any other node, and is lost outright if that Lobe is torn down. Since the whole point of this feature is a durable, centrally-queryable calibration dataset, this quietly undercuts itself in any cluster deployment with read replicas. Small fix — wrap the `Set` in a batch and call `replicateBatch`.

2. **Same `ClearVault` gap we flagged on #576, but with a privacy wrinkle here.** `internal/storage/vault_lifecycle.go`'s prefix list doesn't cover the recall-event prefix, so clearing a vault leaves its recall events behind. That's more than a leak in this case: `RecallEvent.Context` stores the raw, unredacted query text a caller searched for. Someone clearing a vault for privacy/content-removal reasons would still have those queries sitting in Pebble for up to 90 days despite the vault appearing cleared.

Both are small, targeted fixes. We're happy to merge as soon as these land (and once the `0x29`/`0x2A` prefix question with #576 settles). Thank you again — this and the surrounding retrieval-quality arc (#569–572) are clearly getting real care.

isaac-ranger:
Review response in 641db6a, on a branch freshly rebased onto develop — both asks plus the prefix settlement:

**1. `WriteRecallEvent` replicates.** Batch + `replicateBatch`, matching the rest of `internal/storage`. Your Lobe analysis was the convincing part — recalls are precisely the traffic that lands on read replicas, so the old bare `Set` stranded the calibration record on whichever node served the recall. `TestRepLogHook_WriteRecallEvent` added to the existing hook-test file, and verified to fail against the bare-`Set` version before shipping it.

**2. `ClearVault` covers `0x29`.** Added to the data-prefix list with a comment naming the stakes (raw query text must not outlive a cleared vault). `TestRecallEvent_ClearVaultRemovesEvents` writes events in two vaults, clears one, and asserts the cleared vault scans empty while the sibling's events survive — also teeth-verified by removing `0x29` from the list and watching it fail.

**3. Prefix question settled by the rebase.** The branch now sits on current develop, where #576's leases own `0x2A` (kept in `ClearVault`'s list) and `0x29` remains free — confirmed by grep on the develop tip before rebasing. The one rebase conflict was the adjacent addition of `LeaseKey` and `RecallEventKey` in `keys.go`; resolved keeping both, `0x29` before `0x2A`.

One observation from the same neighborhood, explicitly *not* acted on: `PruneRecallEvents`' deletion batch and `ClearVault`'s tombstone batch don't call `replicateBatch` either — the latter is pre-existing behavior across all prefixes, so I left both alone rather than widen a review response. If you'd like the prune batch replicated for symmetry (deletes are idempotent on replay), it's a one-liner — say the word.

`gofmt`/`go vet` clean; full `./internal/...` suite passes on the rebased branch (47 packages).

isaac-ranger:
Rebased onto develop @ 9868c62 (post-#617/#618/#619/#621) — the branch had gone CONFLICTING after today's merges. The conflict was exactly the #618 prefix-registry work meeting our 0x29 claim, and the resolution adopts the new regime rather than fighting it:

- `prefix.RecallEvent = 0x29` registered in `internal/prefix` (const + registry entry, category `vault-scoped-data`), and the `TestAll_ConstSliceComplete` mirror list updated — the guard caught the missing entry on the first run, which is that test doing its job.
- `keys.go` RecallEventKey/RecallEventPrefix now use `prefix.RecallEvent` instead of the raw byte, per the registry's never-inline rule.
- ClearVault's inline list is gone; `clearVaultDataPrefixes` gains `prefix.RecallEvent` (with the raw-query-text rationale comment) and `TestClearVaultDataPrefixes_Scope`'s pin updated to match.
- `docs/internals/keyspace-registry.md`: 0x29 row added, free-bytes line updated. (0x29 stays per the prefix settlement in the earlier review round; noted the doc's 0x2B+ preference is for future claims.)

Verified: `go build ./... && go vet` clean, gofmt clean, `./internal/prefix/... ./internal/storage/...` green under `-race`, full `./internal/...` green.

scrypster (approved):
This is exactly the turnaround I was hoping for. All three came back clean.

Replication now rides replicateBatch like every other write, and TestRepLogHook_WriteRecallEvent pins it so a future refactor can't quietly drop it back to a bare Set. The ClearVault fix is the one I cared most about, since Context carries raw query text and a privacy-clear that left it behind for 90 days would've been a real footgun. TestRecallEvent_ClearVaultRemovesEvents proves the tombstone lands and a sibling vault survives. And 0x29 is disjoint in the new registry with the doc updated.

I reran your suite and reverted each fix in place to watch the tests redden, both did. We also went at the isolation adversarially, wrote a batch of events full of dummy secret text, cleared the vault, and scanned the raw key range bypassing the purpose gate. Zero left. And the purpose allowlist has zero readers in the recall path, so nothing can accidentally recall from the instrument data. That isolation was the part I'd have worried about most, and it holds.

The observe-mode gate is the right instinct too. Calibration wants the truth of what a mutating recall surfaced, and a read-only recall still must not write, and here those two don't fight.

Two small things for the record, neither a change. First, 0x29 brushes an mbp wire opcode, different keyspace so it's safe, I'll note it in the registry so nobody re-flags it. Second, worth a thought down the road, every non-observe recall now emits a replicated write, so on a cluster the rep-log grows with your recall rate, and recall is your busiest path. Not a blocker, just something to watch if rep-log volume ever gets loud.

Merging it. Thanks for measuring all of this in prod. Good to see you again after #471, #621, and #572, the #569 to 573 arc keeps paying off.

PR #575, opened 2026-07-02
feat(vault): multi_user setting — shared-vault session-start guidance

johanneshauer:
## Problem

All of MuninnDB's guidance surfaces steer clients to `muninn_where_left_off` at session start: the MCP `initialize` instructions, `muninn_guide` (session-start pattern, tool list, tips), the tool's own description, and the empty-recall hint.

That is the right advice for a single-user vault — but `muninn_where_left_off` and `muninn_session` return **vault-global** recency. In a vault shared by multiple users or agents, an assistant following the server's own guidance starts its session by surfacing *other users'* recent work and acting on it. We run a shared team vault (multiple users, per-user tags on every memory) and currently have to override the server guidance client-side in every harness — the server tells agents to do one thing, the deployment needs the opposite.

## Change

A vault-level `multi_user` plasticity setting, **default `false` — zero behavior change for existing deployments**:

- **`muninn_guide`** (`internal/mcp/guide.go`): when `multi_user` is set, the session-start pattern recommends recall scoped to the caller's per-user tag, and marks `muninn_where_left_off`/`muninn_session` as vault-global (admin/audit use). Tool list, Tips, and the Vault Configuration summary reflect it too.
- **Empty-recall hint** (`internal/mcp/handlers.go`): multi-user vaults get a scoped-recall hint instead of the `where_left_off` suggestion. The plasticity lookup runs only on the empty-result path; on lookup error the existing hint is kept.
- **`muninn_where_left_off` response hint** (`internal/mcp/handlers.go`): in a multi-user vault the response no longer claims the results are "your most recently accessed memories" — it states they span all users and points to scoped recall. This fires at the exact moment of misuse, catching agents that never consulted the guide.
- **`muninn_session`** gets the same treatment as `where_left_off` (guide tool-list entry + static tool description) — it returns vault-global activity too.
- **Web console lossy save fixed** (`web/static/js/app.js`): `savePlasticity()` built its PUT payload from scratch while the server replaces the stored config wholesale — any save from the Plasticity panel silently dropped fields the panel doesn't edit (`multi_user`, but also pre-existing ones like `behavior_mode`). The panel now merges the loaded raw config into the payload, so unedited fields survive.
- **Static surfaces reworded unconditionally** (`internal/mcp/context.go` instructions const, `internal/mcp/tools.go` tool description): these are served before/independent of vault resolution, so they now describe both modes truthfully instead of prescribing `where_left_off` for everyone.
- **Config plumbing** (`internal/auth/plasticity.go`): `MultiUser *bool` on `PlasticityConfig` (JSON `multi_user`), resolved onto `ResolvedPlasticity` following the existing pointer-override pattern. No preset changes.
- **No CLI changes needed**: the reflective field map behind `muninn vault plasticity <vault> --set key=value` picks the new field up automatically —
  ```
  muninn vault plasticity team --set multi_user=true
  ```

## Tests

- `internal/auth`: `multi_user` defaults false, pointer override works.
- `internal/mcp`: multi-user guide recommends per-user-scoped recall, drops the `where_left_off` session-start tip, and flags both recency tools in the tool list; single-user guide unchanged (pinned by its own assertions). Handler-level tests cover the multi-user `where_left_off` hint and both variants of the empty-recall hint.
- Full suite green locally (`go test ./...`, 49 packages ok; `go vet` clean, `gofmt` clean).

## Docs

`docs/feature-reference.md` and `docs/auth.md`: new row in the plasticity-field tables; `openapi.yaml`: `multi_user` added to the `PlasticityConfig` schema.

### DIFF
--- a/docs/auth.md
+++ b/docs/auth.md
@@ -222,6 +222,7 @@ curl -X PUT "http://127.0.0.1:8475/api/admin/vault/default/plasticity" \
 | `preset` | string | `default` \| `reference` \| `scratchpad` \| `knowledge-graph` | Base cognitive profile; overrides applied on top |
 | `hebbian_enabled` | bool | — | Enable/disable Hebbian weight updates (coactivation learning) |
 | `temporal_enabled` | bool | — | Enable/disable time-based temporal scoring |
+| `multi_user` | bool | — | Vault shared by multiple users/agents: guide + recall hints steer clients to per-user scoped recall; `where_left_off`/`session` flagged vault-global |
 | `hop_depth` | int | 0–8 | BFS hops for associative retrieval; higher = broader context |
 | `semantic_weight` | float | 0–1 | Multiplier for semantic similarity in fusion scoring |
 | `fts_weight` | float | 0–1 | Multiplier for full-text keyword match scoring |
--- a/docs/feature-reference.md
+++ b/docs/feature-reference.md
@@ -341,6 +341,7 @@ Every cognitive behavior is tunable per vault:
 | `pas_max_injections` | 5 | 0–10 | Max PAS candidates to inject |
 | `behavior_mode` | `"autonomous"` | autonomous/prompted/selective/custom | How the AI should use memory (see below) |
 | `behavior_instructions` | `""` | string | Custom instructions for "custom" mode |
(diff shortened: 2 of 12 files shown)


scrypster (approved):

scrypster (approved):
Reviewed the diff and ran the test suite on the branch (go build, go test ./internal/auth/... ./internal/mcp/..., go vet, gofmt) — all clean. Change is well-scoped and gated behind a default-false multi_user flag, so single-user vaults get identical behavior. Good test coverage on config resolution and guide/handler text.

PR #580, opened 2026-07-03
fix(cli): route joined exec:/logs: tokens to their handlers

isaac-ranger:
## Bug

`parseSubcommand` joins any two non-flag words (`exec read` → `exec:read`), but `main.go`'s dispatch switch only matched bare `exec`/`logs` — so every documented two-word form of `exec` and `logs` (`muninn exec read --id …`, `muninn logs 50`, …) died with `Unknown command`, despite the help text documenting them as valid. `cluster:` already had a prefix route; this was an omission for the other two multi-word commands.

## Fix

Two prefix routes in `cmd/muninn/main.go`, mirroring the existing `cluster:` pattern (11 lines). `parseSubcommand`'s join behavior is untouched — it's load-bearing for completion/show/start:web.

## Tests

- Four regression cases added to the dispatch test table (`cmd/muninn/dispatch_test.go`)
- `go test ./cmd/muninn`: ok (full suite), `go vet` clean — re-verified cherry-picked onto current `develop` tip
- Behavioral: `muninn exec read --id <ulid>` now routes to the engine (clean exit-3 not-found on an empty store); `logs 5` routes with a graceful missing-file message; bogus two-word commands still fail loudly

## Provenance

Found during a backup restore drill on our production deployment (2026-07-03); fix authored by Cairn (co-author preserved in the commit), deployed and verified locally, upstreamed per our keep-fixes-regenerable practice — same lineage as #570/#572/#574.

https://claude.ai/code/session_01VtBjGQt6torDYqxervRnqc

### DIFF
--- a/cmd/muninn/dispatch_test.go
+++ b/cmd/muninn/dispatch_test.go
@@ -21,6 +21,12 @@ func TestParseSubcommand(t *testing.T) {
 		{[]string{"help"}, "help"},
 		{[]string{"start", "--dev"}, "start"}, // flags are not subcommands
 		{[]string{"init", "--yes"}, "init"},   // flags are not subcommands
+		// Joined container tokens main.go must route (a joined token with no
+		// route dies as Unknown command while help documents it as valid):
+		{[]string{"exec", "read"}, "exec:read"},
+		{[]string{"exec", "recall", "--query", "x"}, "exec:recall"},
+		{[]string{"logs", "50"}, "logs:50"},
+		{[]string{"cluster", "info"}, "cluster:info"},
 	}
 	for _, tc := range cases {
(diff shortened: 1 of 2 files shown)


scrypster (approved):
Thank you for this — clean, minimal fix that mirrors the existing cluster: pattern exactly, and traced correctly: bare exec/logs already worked, this closes the gap for the joined two-word forms parseSubcommand produces. Verified on the branch: full cmd/muninn suite, gofmt, go vet all clean, and manually built + ran the binary to confirm exec read and logs 5 both route to their real handlers now (previously "Unknown command"). Good catch to find this during a restore drill — appreciate it being upstreamed.

Issue #583, opened 2026-07-03
Local ONNX embedder: `local://<model>` parses a model name that is then ignored — proposal: optional user-supplied model (multilingual use case, zero new dependencies)


johanneshauer:
## What this is

A feature proposal, deliberately filed as an issue rather than a PR. The design
has a few genuine forks (config surface, which knobs are in scope, how dimension
is established) where I'd rather have your direction before writing code. If you
want the feature, I'm happy to implement whichever shape you pick; if you'd
rather keep the local path fixed-function, that's a legitimate answer too and
the docs can say so.

## Current state at `v0.7.0` (verified against the tag)

The URL grammar already *looks* pluggable, but the local provider is
fixed-function end to end:

- `internal/plugin/provider.go:57–67` — `ParseProviderURL` parses a model name
  out of `local://<model>` (hostname or path form), defaulting to
  `"bge-small-en-v1.5"` when empty, and stores it in `ProviderConfig.Model`.
- `internal/plugin/embed/embed.go:116–126` — `EmbedService.Init` passes that
  `Model` into `ProviderHTTPConfig` and logs it
  (`"initializing embed provider" … model=…`).
- `internal/plugin/embed/local.go:56` — `LocalProvider.Init` then reads only
  `cfg.DataDir` (`local.go:60`) and never `cfg.Model`. Everything else is
  compiled in: extraction dir `models/bge-small` (`local.go:66`),
  `tokenizer.json` (`local.go:91`), `model_int8.onnx` (`local.go:101`), the
  ONNX tensor names (`local.go:122–123`), `localModelDim = 384` (`local.go:23`,
  returned at `local.go:137`), `localMaxTokens = 256` (`local.go:24`), and CLS
  pooling (`local.go:236`).
- Both call sites pass the literal default URL anyway:
  `cmd/muninn/server.go:431` (saved-config `case "local"`, which does not
  consult `embed_url`) and `cmd/muninn/server.go:498` (the step-3 default).
- The assets themselves are baked in at build time via `go:embed`
  (`internal/plugin/embed/local_assets_common.go:11–17`) from
  `MODEL_REPO := Xenova/bge-small-en-v1.5` (`Makefile:5`).

So `local://anything-at-all` would parse, log the requested name, and load
bge-small regardless. #455 already observed that the `local://` model field was
parsed but unused; #508 (shipped in v0.6.3) fixed the label mismatch from that
issue, but the field is still unused at `v0.7.0`.

[...]

Issue #582, opened 2026-07-03
Explicitly configured embed provider silently falls back to the bundled 384-dim model, splitting the vault into invisible embedding spaces


johanneshauer:
## Summary

When `plugin_config.json` explicitly configures an embed provider (e.g.
`ollama`) and that provider is unreachable at boot, the server does not fail
startup and does not degrade to its own safe "semantic similarity disabled"
mode. Instead it silently substitutes the bundled `bge-small-en-v1.5` local
ONNX model — 384-dim, English-only — and keeps writing to and searching a
vault whose existing vectors have a different dimension (1024 in our case).

Nothing errors. Writes succeed, searches return results, `muninn_status`
reports `health: "good"`. But the vault is now split into two mutually
invisible embedding spaces: queries embedded at 384 dims only ever match
384-dim engrams, and vice versa. Every engram written during the fallback
window stays permanently invisible to semantic recall after the configured
provider comes back — until it is manually re-embedded.

The failure mode is the bad kind: silent, plausible-looking wrong answers
instead of an error.

## Environment

- muninn `v0.7.0`, official release binary, linux-amd64
  (sha256 `ae201a69…`, verified against the published checksum)
- Fresh data dir, single vault
- Configured provider: Ollama serving `bge-m3` (1024-dim)
- Reproduced 2026-07-03; all operations via the MCP HTTP endpoint

## Reproduction

1. Fresh data dir. `plugin_config.json`:

   ```json
   {"embed_provider": "ollama", "embed_url": "ollama://localhost:11434/bge-m3"}
   ```

   Ollama running and serving `bge-m3`.

2. Start the server. Startup log confirms the real provider:
   `"Ollama dimension probe successful" dimension=1024`, then
   `"embed provider initialized" name=embed-ollama dimension=1024`.

3. Store three memories via `muninn_remember` (in our run: a PostgreSQL
   persistence fact, a JWT expiry decision, and one German-language memory).
   Sanity-check semantic recall with a paraphrased query — *"which database
   engine did we pick for storage"* (`mode=semantic`) returns the PostgreSQL
   memory first, vector_score 0.426. So far, healthy.

4. Stop the server and make the configured provider unreachable at boot —
   we pointed `embed_url` at a closed port (`ollama://localhost:19999/bge-m3`;
   stopping Ollama should behave identically — any init failure takes the
   same fall-through path). Restart. Startup log:

   ```
[...]

Issue #455, opened 2026-05-27
Bundled local embedder label is 'all-MiniLM-L6-v2' but actual ONNX bytes are bge-small-en-v1.5


johanneshauer:
## Observation

The bundled local embedder is referred to by two different names in the codebase:

**`local://all-MiniLM-L6-v2` (URL string)** appears in three places:
- `cmd/muninn/server.go:428` — saved-config path
- `cmd/muninn/server.go:495` — bundled-fallback path
- `internal/plugin/provider.go:64` — default model when the URL omits one

**`bge-small-en-v1.5`** is the actual bundled model:
- `Makefile:5` — `MODEL_REPO := Xenova/bge-small-en-v1.5` (downloaded into `internal/plugin/embed/assets/model_int8.onnx`)
- `internal/plugin/embed/local.go:23` — code comment: `localModelDim = 384  // bge-small-en-v1.5 output dimension`
- Runtime log line: `local embed provider initialized model=bge-small-en-v1.5 dimension=384`

The URL string's model field is parsed by `ParseProviderURL` but **not used** by `LocalProvider` (`internal/plugin/embed/local.go` loads the embedded ONNX bytes regardless of `cfg.Model`). So `all-MiniLM-L6-v2` in the URL is effectively cosmetic metadata that has drifted away from the actual model.

## Why it might or might not matter

Reading the source alone, the model name appears to be `all-MiniLM-L6-v2`. Reading the audit log / status output, it's `bge-small-en-v1.5`. I hit this while debugging an unrelated embedder migration:

> Memory said the bundled fallback was bge-small, but `server.go` says all-MiniLM-L6-v2 — which is it?

(Answer: it's bge-small. The URL string is a leftover from when MuninnDB shipped MiniLM.)

That said, I'm aware this might be intentional:
- The URL string is parsed-but-ignored, so changing it has no runtime effect.
- Existing saved configs (`plugin_config.json` files) may have `embed_url: "local://all-MiniLM-L6-v2"` baked in; renaming the default could break those if the parser ever stops accepting the legacy string.
- The `localModelDim = 384  // bge-small-en-v1.5 output dimension` comment shows the maintainer is already aware of the actual model, so this isn't a discovery — it's a question of whether the cosmetic name should be updated.

## Possible fix (if desired)

[...]

PR #584, opened 2026-07-03
docs(embed): correct the bundled model name and document its English-only language coverage

johanneshauer:
## Summary

Two related docs gaps around the bundled local embedder, both checked against `v0.7.0`:

**1. Three docs still carry the pre-#455 model name.** #455 (bundled embedder
labeled `all-MiniLM-L6-v2`, actual bytes `bge-small-en-v1.5`) was fixed in #508
and shipped in v0.6.3 (the `CHANGELOG.md` `[0.6.3]` entry records the fix,
referenced there as #455). The code
and build agree: `Makefile:5` (`MODEL_REPO := Xenova/bge-small-en-v1.5`) and
`internal/plugin/embed/local.go:23` (`localModelDim = 384 //
bge-small-en-v1.5 output dimension`). But at `v0.7.0` the docs still name the
old model in three places:

- `docs/plugins.md:104` — provider comparison table, "Bundled local" row
- `docs/quickstart.md:149` — "Choose an Embedder" intro sentence
- `docs/self-hosting.md:100` — "Bundled (no API key, no internet)" section

**2. Nothing in the docs says the bundled model is English-only.** The `en` in
`bge-small-en-v1.5` is a language tag: it is an English embedding model. That
is a property of the model, not a MuninnDB defect — but since this embedder is
the zero-config default, the consequence for non-English vaults deserves a
docs section. The failure mode is specific and easy to miss: the model still
emits a vector for any input (nothing errors, nothing is logged), but
non-English vectors organize by *language* rather than by *meaning*, while the
FTS stream keeps answering near-lexical queries in any language — so semantic
recall degrades partially instead of failing loudly.

What prompted this: our production vault (~1,000 memories, German/English)
runs the bundled model. A controlled paired-query test showed three unrelated
German queries — three different operational topics — all returning the same
cluster of unrelated German-language memories, while the English paraphrases
of the same three questions each retrieved the correct memories. After
migrating the vault to a multilingual model, the same German queries returned
topically distinct, correct results. Nothing in logs or status ever pointed at
the embedder, which is why the proposed section includes a five-minute
self-test.

Related but independent: #583 proposes making the bundled local ONNX model
user-replaceable, and #582 reports the startup fallback hazard around external
providers. This PR only documents current `v0.7.0` behavior.

## Changes

- `docs/plugins.md` — fix the model name in the Tier 2 provider table; add a
  new `### Language Coverage` subsection to §3 (between the
  `MUNINN_OPENAI_URL` paragraph after the provider table and
  `### Retroactive Enrichment`, i.e. after line 113 at `v0.7.0`).
- `docs/quickstart.md` — fix the model name in §6 and add a one-line language
  note linking to the new subsection.
- `docs/self-hosting.md` — fix the model name and add a one-line pointer.
- Optional (take it or leave it): one-sentence rewording in
  `docs/plugins.md` §5 so it doesn't contradict the new subsection's re-embed
  guidance — details at the end.

No benchmark claims, vendor-neutral wording. Happy to move the section, rename
it, or fold it into the quickstart instead — whatever fits your structure.

---

### 1. `docs/plugins.md` — provider table row (line 104 at `v0.7.0`)

Current:

```markdown
| Bundled local | On by default | all-MiniLM-L6-v2 | 384 | Zero | Full (local) |
```

Proposed:

```markdown
| Bundled local | On by default | bge-small-en-v1.5 (English-only) | 384 | Zero | Full (local) |
```

### 2. `docs/plugins.md` — new subsection in §3, after the `MUNINN_OPENAI_URL` paragraph (line 113), before `### Retroactive Enrichment` (line 115)

````markdown
### Language Coverage

The bundled local model is `bge-small-en-v1.5` — the `en` is a language tag.
It is an English embedding model. That is a property of the chosen model, not
a MuninnDB defect: within English it is a solid default, and it keeps Tier 2
zero-config and offline.

For non-English text the model still produces a vector — nothing fails,
nothing is logged — but the vectors organize by **language** rather than by
**meaning**: two unrelated German memories land closer together than a German
memory and its English paraphrase. On a non-English or mixed-language vault
this surfaces as:

- Non-English semantic queries return the same cluster of same-language
  memories, regardless of topic.
- English queries on the same vault behave normally.
- Queries that share words with the stored text still work in any language,
  because the FTS stream matches them. The fusion keeps returning plausible
  results, so the degradation is partial — and easy to miss.

**Five-minute self-test.** If your vault holds non-English content:

1. Store three memories in your language, on clearly unrelated topics.
2. Query each topic with a same-language *paraphrase* — reworded so it shares
   no distinctive keywords with the stored text (otherwise FTS answers the
   query and hides the vector behavior).
3. Repeat step 2 with English paraphrases of the same three questions.

If the same-language queries all return roughly the same result set regardless
of topic, while the English paraphrases each retrieve the right memory, your
vault is affected.

**Switching to a multilingual provider.** Any external provider serving a
multilingual embedding model resolves this. Self-hosted example with Ollama:

```bash
# any multilingual embedding model Ollama serves, e.g. bge-m3
export MUNINN_OLLAMA_URL="ollama://localhost:11434/bge-m3"
muninn server
```

Several hosted providers in the table above also serve multilingual models —
check your provider's model documentation. The same choice can be persisted in
`plugin_config.json` in the data directory (`embed_provider`, `embed_url`,
`embed_api_key`) or set in the web admin UI; environment variables take
precedence over saved config.

Two things to plan for when switching models:

- **Re-embed the vault.** Vectors from different models are not comparable:
  similarities across models are meaningless, and when the dimensions differ,
  memories embedded with the previous model drop out of semantic recall
  entirely. After switching, run `muninn vault reembed <name>` for each
  affected vault — it clears the stored embeddings and re-embeds with the
  current model.
- **Scores shift.** Different models produce different similarity-score
  distributions. If you have tuned anything against absolute vector scores,
  re-check it after the switch.
````

(The `plugin_config.json` keys and env-var precedence match
`internal/config/plugin.go:12-17` at `v0.7.0`; the `reembed` description
matches the CLI help text at `cmd/muninn/vault.go:33`, "Clear embeddings and
re-embed with current model".)

### 3. `docs/quickstart.md` — §6 intro sentence (line 149 at `v0.7.0`)

Current:

```markdown
The bundled local embedder (all-MiniLM-L6-v2, 384-dim) is included and works offline with no API key. It's the default. For higher quality or different dimensions:
```

Proposed:

```markdown
The bundled local embedder (bge-small-en-v1.5, 384-dim) is included and works offline with no API key. It's the default. It is an English-only model — for non-English or mixed-language memories, pick a multilingual external provider (see [Language Coverage](plugins.md#language-coverage)). For higher quality, different dimensions, or other languages:
```

### 4. `docs/self-hosting.md` — bundled-model line (line 100 at `v0.7.0`)

Current:

```markdown
The bundled `all-MiniLM-L6-v2` INT8 model (384-dim, ~80MB) is active automatically when the binary was built with embedded assets. No configuration needed.
```

Proposed:

```markdown
The bundled `bge-small-en-v1.5` INT8 model (384-dim, ~80MB, English-only) is active automatically when the binary was built with embedded assets. No configuration needed. For non-English content, see [Language Coverage](plugins.md#language-coverage).
```

### 5. Optional: `docs/plugins.md` §5 (line 231 at `v0.7.0`)

§5's second paragraph (line 231 at `v0.7.0`) opens with:

> Most vector database integrations require you to re-index when you switch
> embedding models.

With the new subsection stating that switching models *does* require
`muninn vault reembed`, that sentence reads as a contradiction. Its real point
is the install-time backfill guarantee, so a minimal reword keeps the point
and removes the tension:

```markdown
Most vector database integrations require a manual re-index before existing data becomes searchable.
```

Only this first sentence changes; the rest of the paragraph stays as is.

If you'd rather keep §5 untouched, the PR works without this edit.

---

## Release Checklist

- [ ] ~~`CHANGELOG.md`~~ — maintainers update this at release time, no action needed
- [ ] OpenAPI spec (`internal/transport/rest/openapi.yaml`) updated if API routes changed — N/A, docs-only
- [ ] SDK types updated if request/response schemas changed (Python, Node, PHP) — N/A, docs-only
- [x] `docs/` updated if user-facing behavior changed
- [ ] `go build ./...` clean — unaffected, docs-only
- [ ] `go vet ./...` clean — unaffected, docs-only
- [ ] `go test ./...` passes — unaffected, docs-only

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -101,7 +101,7 @@ Provider comparison:
 
 | Provider | Env Var | Model (default) | Dimensions | Cost | Privacy |
 |----------|---------|-----------------|------------|------|---------|
-| Bundled local | On by default | all-MiniLM-L6-v2 | 384 | Zero | Full (local) |
+| Bundled local | On by default | bge-small-en-v1.5 (English-only) | 384 | Zero | Full (local) |
 | Ollama | `MUNINN_OLLAMA_URL` | configurable | varies | Zero (local) | Full (local) |
 | OpenAI | `MUNINN_OPENAI_KEY` | text-embedding-3-small | 1536 | Per token | API |
 | Voyage AI | `MUNINN_VOYAGE_KEY` | voyage-3 | 1024 | Per token | API |
@@ -111,6 +111,65 @@ Provider comparison:
 | Mistral | `MUNINN_MISTRAL_KEY` | mistral-embed | 1024 | Per token | API |
 
 `MUNINN_OPENAI_URL` can optionally override the OpenAI base URL for compatible endpoints (for example LocalAI or an internal gateway). If set to an invalid value, MuninnDB skips OpenAI initialization instead of falling back to `api.openai.com`. This override also applies to the Enrich plugin when `MUNINN_ENRICH_URL` is set to an `openai://` provider — see [Tier 3](#4-tier-3-enrich-plugin) below.
+
+### Language Coverage
+
+The bundled local model is `bge-small-en-v1.5` — the `en` is a language tag.
+It is an English embedding model. That is a property of the chosen model, not
+a MuninnDB defect: within English it is a solid default, and it keeps Tier 2
+zero-config and offline.
+
+For non-English text the model still produces a vector — nothing fails,
+nothing is logged — but the vectors organize by **language** rather than by
+**meaning**: two unrelated German memories land closer together than a German
+memory and its English paraphrase. On a non-English or mixed-language vault
+this surfaces as:
+
+- Non-English semantic queries return the same cluster of same-language
+  memories, regardless of topic.
+- English queries on the same vault behave normally.
+- Queries that share words with the stored text still work in any language,
+  because the FTS stream matches them. The fusion keeps returning plausible
+  results, so the degradation is partial — and easy to miss.
+
+**Five-minute self-test.** If your vault holds non-English content:
+
+1. Store three memories in your language, on clearly unrelated topics.
+2. Query each topic with a same-language *paraphrase* — reworded so it shares
+   no distinctive keywords with the stored text (otherwise FTS answers the
+   query and hides the vector behavior).
+3. Repeat step 2 with English paraphrases of the same three questions.
+
+If the same-language queries all return roughly the same result set regardless
+of topic, while the English paraphrases each retrieve the right memory, your
+vault is affected.
+
+**Switching to a multilingual provider.** Any external provider serving a
+multilingual embedding model resolves this. Self-hosted example with Ollama:
+
+```bash
+# any multilingual embedding model Ollama serves, e.g. bge-m3
+export MUNINN_OLLAMA_URL="ollama://localhost:11434/bge-m3"
+muninn server
+```
+
+Several hosted providers in the table above also serve multilingual models —
+check your provider's model documentation. The same choice can be persisted in
+`plugin_config.json` in the data directory (`embed_provider`, `embed_url`,
+`embed_api_key`) or set in the web admin UI; environment variables take
+precedence over saved config.
+
+Two things to plan for when switching models:
+
+- **Re-embed the vault.** Vectors from different models are not comparable:
+  similarities across models are meaningless, and when the dimensions differ,
+  memories embedded with the previous model drop out of semantic recall
+  entirely. After switching, run `muninn vault reembed <name>` for each
+  affected vault — it clears the stored embeddings and re-embeds with the
+  current model.
+- **Scores shift.** Different models produce different similarity-score
+  distributions. If you have tuned anything against absolute vector scores,
+  re-check it after the switch.
 
 ### Retroactive Enrichment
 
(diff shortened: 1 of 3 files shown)


scrypster (approved):
Thank you for this — genuinely excellent work, and it connects directly to what we discussed with you earlier about the bge-small-en-v1.5/bge-m3 migration. We verified every factual claim against the actual code rather than taking it on faith: the stale model name at all three cited lines, Makefile's MODEL_REPO and local.go's dimension/model comments, the CHANGELOG's 0.6.3/#455 history, the plugin_config.json key names against internal/config/plugin.go, and the `muninn vault reembed` help text against cmd/muninn/vault.go — all match exactly.

The optional §5 reword is a good catch too — the original wording would have contradicted the new Language Coverage section's statement that switching models requires a manual reembed. Narrowing it to what §5 is actually about (first-time backfill on plugin install) resolves that cleanly, so we're taking that edit as well.

The self-test methodology (same-language paraphrase to defeat FTS, contrasted against an English paraphrase of the same question) is a genuinely useful diagnostic for anyone in the same situation you were in — thank you for turning your own debugging into documentation that saves the next person the same investigation.

PR #589, opened 2026-07-07
feat(embed): vault dimension guard (#582) + user-supplied local ONNX model (#583)

### LINKED ISSUE #583: Local ONNX embedder: `local://<model>` parses a model name that is then ignored — proposal: optional user-supplied model (multilingual use case, zero new dependencies)

johanneshauer:
Implements the two halves agreed in #583's discussion: the dimension guard from
#582 direction (b) first, then the user-supplied local ONNX model built on top
of it. Two commits: the guard first, then the feature built on it — the
sequencing from the issue discussion.

## Part 1 — dimension guard (Refs #582)

The server-side embedder could still point a differently-dimensioned vector at
a populated vault (client-supplied embeddings were already checked at the MCP
layer with -32602, but no other path was). A mismatched vector inserts into
HNSW without error today, and CosineSimilarity's length check scores it 0
against everything — the silent split from #582. This PR makes the vault
dimension an explicit, enforced invariant:

- **One definition.** `hnsw.CheckDim` + an explicit per-index dimension,
  established atomically on first insert (two concurrent first inserts with
  different dimensions cannot both win) and derived deterministically from the
  smallest-ID vector on load — consistent with the storage layer's
  first-embedding-key derivation, instead of random map iteration on legacy
  mixed-dimension vaults.
- **Every write path refuses before persisting.** The engine validates
  client-supplied embeddings in Write/WriteBatch/Evolve/AddChild before
  anything is written (extending the MCP layer's -32602 contract to MBP,
  gRPC, SDKs, and the embedded library); the plugin store adapter refuses
  before the embedding key is written; `Registry.Insert` is the atomic
  backstop. The dimension peek never loads a graph (cached index or one
  Pebble seek).
- **Mismatch is a vault condition, not an engram failure.** The retroactive
  processor checks the vault's dimension against the embedder BEFORE paying
  for inference and skips mismatched engrams — no wasted ONNX/API spend, one
  summary log line per pass, and deliberately NO `DigestEmbedFailed`: that
  flag shares bit 0x80 with `DigestEnrichFailed` (stamping it would silently
  stop enrichment too), and the condition heals via `vault reembed` or a
  config fix, after which the still-pending engrams embed normally.
- **Fail closed on unknown state.** A vault whose graph failed to load
  refuses inserts (its real dimension is unknown; `Dim()==0` would accept
  anything and could recreate the split) — reads keep the degraded-empty
  contract from #499, and the insert retries next cycle.
- **Recall degrades, never breaks.** A mismatched query embedding returns a
  typed error from `Registry.Search`; activation's existing search-error path
  degrades to BM25-only (the #578 contract), logging both dimensions and the
  remedy.
- **Operators see it at boot.** Startup prints a per-vault advisory (one
  Pebble seek per vault, no graph loads). Deliberately not fatal: one legacy
  vault must not take down FTS recall for every other vault.
- The noop embedder now returns no embedding instead of a 384-dim zero
  vector: recalls take the existing FTS-only fast path, and the stub cannot
  trip the guard on non-384 vaults.

`vault reembed` composes unchanged (it clears vectors first ⇒ dimension
resets ⇒ the new model's dimension is accepted).

Stated limits (deliberate): a same-dimension model swap is invisible to any
dimension guard — that needs the vault-level model marker (0x1D, currently
write-only) as a follow-up. Cluster replication applies embeddings as raw KV
and bypasses engine-level guards — out of scope here.

Considered and rejected: merging UpdateEmbedding+HNSWInsert into one store
operation would remove the stranding hazard structurally, but changes the
PluginStore interface, the processor's flag handling, and every mock — wrong
size for this PR.

## Part 2 — user-supplied local ONNX model (Fixes #583)

Exactly the v1 scope from the issue discussion:

| Constraint (maintainer comment)            | Implementation |
|---|---|
| plugin_config.json keys, not embed_url/env | `embed_model_path` + `embed_tokenizer_path`, active when `embed_provider` is `"local"` |
| Probe-derived dimension, never declared    | init runs one real inference with an ORT-allocated output tensor and reads the true shape (incl. the sequence axis); validates finite, non-zero |
| Pooling enum                               | `embed_pooling`: `cls` (default) \| `mean` — reuses the existing pooling helpers |
| Fail-loud at init, never fall back to bge  | any failure (missing file, tokenizer, unknown pooling, probe, unsupported output shape) aborts startup with a clear error |
| Decide explicitly on instruction prefixes  | first-class config: `embed_query_prefix` / `embed_passage_prefix`, plus a warn when the filename looks like the e5 family and prefixes are unset or pooling ≠ mean |
| Max sequence length included               | `embed_max_tokens` (default 256) |
| Output tensor name deferred                | uses `last_hidden_state` or the single declared output; knob deferred until a model needs it |
| Dimension guard first or alongside         | Part 1 |

Mechanics: the provider reads the model's declared inputs and feeds exactly
those (`input_ids` + `attention_mask` required, `token_type_ids` optional —
XLM-R-family exports omit it). Prefixes wrap at the construction point in
buildEmbedder, where the query handle (activation embedder) and passage handle
(retroactive processor plugin) diverge — the provider stays role-agnostic, and
e5's asymmetric prefixes come for free. Requires a localassets build (all
release binaries): the bundled ORT shared library serves the user model too.

Config strictness, per the fail-loud principle: default behavior is
byte-identical when nothing is configured; user-model keys set while another
provider is selected warn as inactive; `MUNINN_LOCAL_EMBED=0` **or a leftover
provider env var** alongside a configured user model is a hard startup error —
env precedence would silently embed with a different model, the exact
substitution class of #582.

Decisions worth stating: typed fields through PluginConfig/ProviderHTTPConfig
rather than the `Options` map — compile-time safety and discoverability for a
four-field set, following the existing APIKey/DataDir plumbing. The prefix
decorator forwards `HardwareAccelerated`; any future optional interface on
EmbedPlugin needs the same forwarding (noted on the wrapper).

## Tests

- Full `-tags localassets ./... -race` suite green (the main CI
  configuration), plus the untagged suite for noembed builds.
- Real-inference user-model tests using the bundled assets materialized as
  plain files: probe-derived dimension; embeddings byte-identical to the
  bundled provider for the same model files; mean pooling verifiably differs;
  six fail-loud cases.
- Guard tests: atomic first-insert establishment under concurrency;
  fail-closed insert on load failure with recovery; registry insert/search
  refusal incl. persistence rollback and per-vault independence; engine-level
  client-embedding refusal; adapter refusal before the embedding write (warm
  and cold cache); processor skips before inference without flagging and
  leaves race-refused engrams pending; on-disk dimension derivation; noop
  contract.
- buildEmbedder branch/diagnostic tests in the #585 style (half-configured
  paths, env conflicts, broken paths never fall back, inactive keys), passing
  in both asset build variants.

## End-to-end validation (local test instance, real models)

Phase A: fresh vault embedded at 1024 dims via an external provider (bge-m3).
Phase B: switched the same vault to `Xenova/multilingual-e5-small` (quantized
ONNX, 384-dim, mean pooling, `query: `/`passage: ` prefixes):
- init probed 384 and logged pooling/inputs;
- the boot advisory named the vault and both dimensions;
- new memories were refused semantic indexing (left pending, exactly one log
  event — no retry loop) while FTS recall kept serving them;
- semantic recall degraded to BM25-only with both dims logged;
- `vault reembed` healed everything: all engrams re-embedded at 384, and a
  German-language paraphrase query sharing no keywords with the stored text
  returned the correct memory first (0.87) — the multilingual case the
  bundled English-only model cannot serve, per the docs self-test.

Phase C (pending-not-failed semantics): with a deliberately mismatched
provider (1024-dim) against the 384-dim vault, a new memory was skipped
before inference (one summary log line per pass, no failure flags, FTS
recall unaffected); after switching back to the matching model — no reembed,
no flag clearing — the pending memory embedded automatically on the next
scan and became the top semantic hit for a paraphrase query (0.86).

Refs #582
Fixes #583

### DIFF
--- a/cmd/muninn/coverage_hardening_test.go
+++ b/cmd/muninn/coverage_hardening_test.go
@@ -2476,3 +2476,119 @@ func TestRunVaultDelete_CancelledByUser(t *testing.T) {
 		t.Errorf("expected 'Cancelled': %s", out)
 	}
 }
+
+// ---------------------------------------------------------------------------
+// buildEmbedder: user-supplied local ONNX model (#583)
+// ---------------------------------------------------------------------------
+
+func TestBuildEmbedder_UserModelHalfConfigured_Errors(t *testing.T) {
+	clearEmbedEnv(t)
+	t.Setenv("MUNINN_LOCAL_EMBED", "")
+
+	cfg := plugincfg.PluginConfig{EmbedProvider: "local", EmbedModelPath: "/nonexistent/model.onnx"}
+	embedder, plug, err := buildEmbedder(context.Background(), cfg, t.TempDir())
+	if err == nil {
+		t.Fatal("expected an error when only one of embed_model_path/embed_tokenizer_path is set")
+	}
+	if !strings.Contains(err.Error(), "embed_tokenizer_path") {
+		t.Errorf("error should name the missing key, got: %v", err)
+	}
+	if embedder != nil || plug != nil {
+		t.Error("no embedder may be returned on a user-model configuration error")
+	}
+}
+
+func TestBuildEmbedder_UserModelConflictsWithLocalEmbedOff(t *testing.T) {
+	clearEmbedEnv(t) // leaves MUNINN_LOCAL_EMBED=0, which this test needs
+
+	dir := t.TempDir()
+	cfg := plugincfg.PluginConfig{
+		EmbedProvider:      "local",
+		EmbedModelPath:     filepath.Join(dir, "m.onnx"),
+		EmbedTokenizerPath: filepath.Join(dir, "t.json"),
+	}
+	_, _, err := buildEmbedder(context.Background(), cfg, t.TempDir())
+	if err == nil {
+		t.Fatal("expected an error when MUNINN_LOCAL_EMBED=0 conflicts with a configured user model")
+	}
+	if !strings.Contains(err.Error(), "MUNINN_LOCAL_EMBED") {
+		t.Errorf("error should name the conflicting env var, got: %v", err)
+	}
+}
+
+// TestBuildEmbedder_UserModelBrokenPaths_NoFallback pins the #583 fail-loud
+// contract at the buildEmbedder level: a configured user model whose files do
+// not exist is a hard error — never the bundled model, never noop-with-a-
+// warning. In a localassets build the failure comes from provider init
+// (missing files); in a noembed build it comes earlier (no ONNX runtime).
+// Both build variants must produce an error.
+func TestBuildEmbedder_UserModelBrokenPaths_NoFallback(t *testing.T) {
+	clearEmbedEnv(t)
+	t.Setenv("MUNINN_LOCAL_EMBED", "") // must not trip the =0 conflict check
+
+	dir := t.TempDir()
+	cfg := plugincfg.PluginConfig{
+		EmbedProvider:      "local",
+		EmbedModelPath:     filepath.Join(dir, "does-not-exist.onnx"),
+		EmbedTokenizerPath: filepath.Join(dir, "does-not-exist.json"),
+	}
+	embedder, plug, err := buildEmbedder(context.Background(), cfg, t.TempDir())
+	if err == nil {
+		t.Fatal("expected an error for a user model whose files do not exist")
+	}
+	if embedder != nil || plug != nil {
+		t.Error("no embedder may be returned when the configured user model fails to initialize")
+	}
+}
+
+// TestBuildEmbedder_UserModelConflictsWithEnvProvider pins that an embed
+// provider env var plus a configured user local model is a hard startup error
+// — env precedence would silently override the explicit model, the exact
+// silent-substitution class of #582.
+func TestBuildEmbedder_UserModelConflictsWithEnvProvider(t *testing.T) {
+	clearEmbedEnv(t)
+	t.Setenv("MUNINN_LOCAL_EMBED", "")
+	t.Setenv("MUNINN_OLLAMA_URL", "ollama://localhost:1/some-model")
+
+	dir := t.TempDir()
+	cfg := plugincfg.PluginConfig{
+		EmbedProvider:      "local",
+		EmbedModelPath:     filepath.Join(dir, "m.onnx"),
+		EmbedTokenizerPath: filepath.Join(dir, "t.json"),
+	}
+	_, _, err := buildEmbedder(context.Background(), cfg, t.TempDir())
+	if err == nil {
+		t.Fatal("expected an error when an env provider conflicts with a configured user model")
+	}
+	if !strings.Contains(err.Error(), "MUNINN_OLLAMA_URL") {
+		t.Errorf("error should name the conflicting env var, got: %v", err)
+	}
+}
+
+// TestBuildEmbedder_UserModelKeysInactive_OtherProvider pins that user-model
+// keys configured alongside a non-"local" provider are inactive (warned, not
+// fatal): the explicitly configured provider still governs the outcome.
+func TestBuildEmbedder_UserModelKeysInactive_OtherProvider(t *testing.T) {
+	clearEmbedEnv(t) // leaves MUNINN_LOCAL_EMBED=0, which this test wants
+
+	cfg := plugincfg.PluginConfig{
+		EmbedProvider:      "ollama",
+		EmbedURL:           "http://localhost:1/bad",
+		EmbedModelPath:     "/nonexistent/model.onnx",
+		EmbedTokenizerPath: "/nonexistent/tokenizer.json",
+	}
+	var embedder activation.Embedder
+	stderr := captureStderr(func() {
+		var err error
+		embedder, _, err = buildEmbedder(context.Background(), cfg, t.TempDir())
+		if err != nil {
+			t.Fatalf("unexpected error: %v", err)
+		}
+	})
+	if embedder == nil {
+		t.Error("expected noop embedder fallback for the failed explicit provider")
+	}
+	if !strings.Contains(stderr, `"ollama"`) {
+		t.Errorf("expected diagnostic naming the failed provider, got: %s", stderr)
+	}
+}
--- a/cmd/muninn/server.go
+++ b/cmd/muninn/server.go
@@ -676,6 +676,40 @@ func runStartupMigrations(ctx context.Context, store *storage.PebbleStore) {
 	slog.Info("startup migration complete", "vaults", len(names))
 }
 
+// warnVaultDimMismatches compares the active embedder's dimension against each
+// vault's stored embedding dimension (read from the first persisted embedding,
+// without loading any HNSW graph) and prints a prominent warning per mismatch.
+// Advisory only: the per-operation guards in the HNSW registry and the plugin
+// store adapter do the actual refusing (issue #582).
+func warnVaultDimMismatches(store *storage.PebbleStore, embedderDim int) {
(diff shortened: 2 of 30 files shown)


scrypster (approved):
Verified: vault dimension guard fails closed and degrades to BM25-only on mismatch (no silent split-embedding-space corruption); local ONNX support probes dimension via real inference and hard-errors on conflicting env/config providers rather than silently overriding — consistent with this repo's no-silent-fallback principle. Full suite (49/49) + race detector on the relevant packages pass clean. Merges cleanly against current develop.

PR #590, opened 2026-07-08
fix(activation): preserve explicit search threshold under RRF + expose Search Scoring setting

gitssie:
Split out of #431 per maintainer request — this is the focused RRF/threshold change, with no bleve/FAISS backend and zero new dependencies. Based on current `develop`.

## 1. Threshold preservation fix

`ActivationEngine.Run` silently overrode any user-set threshold ≥ 0.01 down to 0.001 whenever RRF was active, trampling explicit user config. Now the RRF-appropriate default (0.001) is applied **only** on the no-threshold path (`req.Threshold <= 0`); explicit user values are respected under both ACT-R and RRF.

Covered by a new table test `TestRun_ThresholdDefaulting` (RED confirmed before the fix, GREEN after).

## 2. Search Scoring UI

Surfaces the existing `scoring_fusion` plasticity config as a first-class "Search Scoring" card selector in vault settings, mirroring the existing "Default Recall Mode" pattern, with plain-English labels:
- **ACT-R (default)** — cognitive decay + Hebbian scoring (unchanged default)
- **RRF** — reciprocal rank fusion; better recall on small corpora
- **Weighted Sum** — linear blend of signals

ACT-R remains the default; RRF stays an explicit opt-in.

## Notes
- Zero new go.mod / JS dependencies; no `internal/search/**` or Dockerfile changes.
- `go build ./...` clean, `go test ./internal/engine/... ./internal/auth/...` pass.
- The bleve/FAISS backend from #431 will be discussed separately in an architecture issue before any further code.

### DIFF
--- a/internal/engine/activation/activation_test.go
+++ b/internal/engine/activation/activation_test.go
@@ -1378,6 +1378,65 @@ func TestRRF_ReturnsResultsWithDefaultThreshold(t *testing.T) {
 }
 
 // ---------------------------------------------------------------------------
+// Test: Run() applies a mode-appropriate DEFAULT threshold only when the caller
+// leaves Threshold <= 0, and never overrides an explicit user-set value.
+//
+// Regression for the threshold-preservation bug: previously any explicit
+// threshold >= 0.01 was silently trampled down to 0.001 whenever RRF was on,
+// which made the API's threshold parameter meaningless under RRF.
+//
+// Run() mutates req.Threshold in place, so we assert on it after the call.
+// ---------------------------------------------------------------------------
+
+func TestRun_ThresholdDefaulting(t *testing.T) {
+	store := newStubStore()
+	eng1 := &storage.Engram{
+		Concept:    "threshold defaulting",
+		Content:    "engram for threshold defaulting behavior",
+		Confidence: 1.0,
+		Stability:  30.0,
+		Relevance:  0.8,
+	}
+	store.writeEngram(eng1)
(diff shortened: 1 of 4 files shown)


scrypster (approved):
Independently reviewed (Opus): cloned the PR branch and confirmed the bug — pre-fix, internal/engine/engine.go set req.Threshold=0.05 as a default, then a second RRF-specific guard (`if UseRRFFusion && req.Threshold >= 0.01`) couldn't distinguish that just-applied default from a deliberate user-supplied threshold, so ANY explicit threshold >= 0.01 was silently stomped to 0.001 whenever RRF fusion was active — a real recall-quality regression on a core scoring path. The fix collapses this to a single default-only branch keyed on `req.Threshold <= 0`, preserving explicit values under both ACT-R and RRF while leaving the no-threshold-supplied defaults unchanged (0.001 RRF / 0.05 ACT-R). New table-driven test covers all four quadrants. The "Search Scoring" UI addition just surfaces a pre-existing, already-validated `scoring_fusion` config field (auth/plasticity.go's ValidScoringFusion) via the same card pattern as the existing "Default Recall Mode" control — no new plumbing. Build/vet/gofmt clean, `go test ./internal/engine/activation/...` passes. The earlier Vulnerability-scan CI failure (GO-2026-5856, a stdlib crypto/tls issue fixed in go1.26.5) was a stale-branch toolchain mismatch, not a defect in this PR's diff — resolved by updating the branch onto develop (which got the go1.26.5 bump via #591).

Approving.

Issue #601, opened 2026-07-12
release.yml embed-asset cache keys still say "minilm-v2" (one even "ort-1.23.2") while the cached assets are bge-small-en-v1.5 / ORT 1.24.2


johanneshauer:
At the v0.8.0 tag, `.github/workflows/release.yml` caches the embedded model + ONNX Runtime assets under keys that no longer describe what is cached:

- linux/darwin matrix (lines ~24–46): `embed-assets-ort-1.24.2-minilm-v2-<os>-<arch>` — but the Makefile's `fetch-model` target has fetched **bge-small-en-v1.5** (`MODEL_REPO := Xenova/bge-small-en-v1.5`) since the model swap.
- darwin-amd64 additionally says `ort-1.23.2` while the sibling keys say `ort-1.24.2` — so the labels have already drifted through at least one ORT bump without the key being updated.
- The Windows job (line ~123) was updated and uses `embed-assets-ort-1.24.2-bge-small-en-v1.5-windows-amd64` — the correct pattern.

## Why it matters

The keys currently *work* (the cached bytes are the right ones — the label is just stale). But the next model or ORT change that lands without a manual key bump will silently restore **stale cached assets into release binaries** — the build succeeds, the binary ships, and nothing fails until someone inspects the embedded model. This is the CI-cache sibling of the label-drift already fixed for the runtime string in #455.

## Suggestion

Derive the cache key from the source of truth instead of a hand-maintained literal — e.g. include the Makefile's `MODEL_REPO`/ORT version variables (or `hashFiles('Makefile')`) in the key, so a model/ORT change automatically invalidates the cache.

johanneshauer:
Correction to my second bullet: the darwin-amd64 `ort-1.23.2` is **not** drift — the comment directly above that key (and `ORT_VERSION_DARWIN_AMD64 := 1.23.2` in the Makefile) marks the pin as intentional (ORT dropped Intel-Mac support after 1.23.2), and the darwin-amd64 job genuinely fetches ORT 1.23.2. My grep filtered out the comment line — apologies for the noise.

The core point stands unchanged: `minilm-v2` in all four matrix cache keys describes a model that hasn't been cached there since the bge-small-en-v1.5 swap, while the Windows job's key already uses the correct pattern.

johanneshauer:
[...]

PR #603, opened 2026-07-12
ci: rename stale 'minilm-v2' embed-asset cache keys to bge-small-en-v1.5 (#601)

johanneshauer:
## Summary

Fixes the `minilm-v2` half of #601: eight embed-asset cache keys (the four darwin/linux matrix keys in `release.yml` and the four linux job keys in `ci.yml`) still name `minilm-v2`, but `make fetch-model` has cached **bge-small-en-v1.5** under those keys since the model swap. `release.yml`'s Windows job already uses the correct pattern; this brings the rest in line.

Label-only change — the cached bytes were always the right ones. The point is future-proofing: the next model change that forgets the manual key bump would silently restore a cache entry named after the wrong model. Cost of the rename: one cold re-download per key on the next run (~56 MB per platform in `release.yml`, per its own comment).

Deliberately untouched:
- the darwin-amd64 `ort-1.23.2` version segment — intentional pin (last ORT with Intel-Mac support, per the adjacent comment and `ORT_VERSION_DARWIN_AMD64` in the Makefile; see the correction comment on #601);
- `ci.yml`'s Windows key — that job genuinely downloads all-MiniLM-L6-v2, so its label matches its bytes (the model difference itself is a separate concern from key naming).

## Changes

- `.github/workflows/release.yml`: 4 matrix `cache_key` literals `-minilm-v2-` → `-bge-small-en-v1.5-` (ORT-version and platform segments unchanged)
- `.github/workflows/ci.yml`: the same rename on the 4 linux job keys

The keys are consumed only via `${{ matrix.cache_key }}` / literal `key:` fields; there are no `restore-keys` or cross-workflow references, so the rename cannot silently break cache restore — a miss just re-runs `make fetch-model`.

## Release Checklist

- [ ] ~~`CHANGELOG.md`~~ — maintainers update this at release time, no action needed
- [ ] ~~OpenAPI spec~~ — no API changes (CI-only)
- [ ] ~~SDK types~~ — no schema changes
- [ ] ~~`docs/`~~ — no user-facing behavior change
- [x] `go build ./...` clean — no Go changes; YAML parse-checked
- [x] `go vet ./...` clean — no Go changes
- [x] `go test ./...` passes — no Go changes; note `release.yml` is tag-triggered and not exercised by PR CI, so this PR is inspection-only

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,7 +46,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: internal/plugin/embed/assets
-          key: embed-assets-ort-1.24.2-minilm-v2-linux
+          key: embed-assets-ort-1.24.2-bge-small-en-v1.5-linux
 
       # Only fetch if the cache missed — saves ~2 min and 130MB on cache hits.
       - name: Fetch embed assets (linux/amd64 only)
@@ -96,7 +96,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: internal/plugin/embed/assets
-          key: embed-assets-ort-1.24.2-minilm-v2-linux
+          key: embed-assets-ort-1.24.2-bge-small-en-v1.5-linux
 
       # Fetch if cache missed (e.g. first run after cache key changes)
       - name: Fetch embed assets if needed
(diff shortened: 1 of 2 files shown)


scrypster (approved):
Independently reviewed (Opus): cloned the PR branch, confirmed the diff renames exactly the 8 stale \`minilm-v2\` cache-key labels in \`ci.yml\`/\`release.yml\` to \`bge-small-en-v1.5\`, matches the Makefile source of truth (MODEL_REPO=Xenova/bge-small-en-v1.5), correctly leaves the Windows job's \`all-MiniLM-L6-v2\` label alone since that job genuinely fetches that model, and found no other repo references to the old key string. Renaming an Actions cache key is safe (worst case: one cold cache miss). CI green on all 8 checks. Real, if small, correctness fix (label/asset mismatch), tightly scoped, nothing unrelated smuggled in.

Approving.

PR #604, opened 2026-07-12
feat(mcp): toolset profiles for tools/list — env default + per-connection header override (#588)

isaac-ranger:
Implements #588 — working code for the design discussion there, running in production in our multi-agent deployment since 2026-07-12 (on v0.7.0; rebased cleanly onto develop for this PR).

## What it does

- `MUNINN_MCP_TOOLSET=core|full` — server default via env. `core` advertises a 10-tool hot-path set; `full` (and unset) is exactly today's behaviour.
- `X-Muninn-Toolset: core|full` request header — per-connection override, beats the env. MCP clients that let users set custom headers in client config (Claude Code, most SDK clients — the same place the Authorization header already lives) get the connect-time win with one config line and no server restart.
- **Advertisement-only**, exactly as #588 frames it: dispatch is by name and never consults the list, so every tool stays callable under any profile. Orthogonal to the key-mode security gate.
- **Fail-open**: an unknown value in header or env falls through (header → env → full) with one logged warning. A typo must never hide a client's memory tools.

## On the selection-mechanism question (A/B/C)

We shipped what amounts to **server default + per-connection override**, without touching the key model. The header reaches the same end as (B)/(C) — per-client choice, one key — while dodging (B)'s weakness: standard clients call `tools/list` with no parameters, but they *do* attach configured headers to every request, so the connect-time saving is realized today. If you'd rather carry the profile as a per-key attribute (A), happy to rework — the filter core, fail-open semantics, and tests all carry over unchanged; only `resolveToolset` moves.

## Numbers (measured on our live instance)

| | tools | bytes | ~tokens |
|---|---|---|---|
| full | 39 | 41,063 | ~10.3k |
| core | 10 | 16,600 | ~4.2k |

In a bounded-context harness this measured as 23% of a typical session's context before, and the schemas outweighed the actual memory content a session read.

## Membership (maintainer's call, happy to adjust)

Your proposed lite set +`remember_batch` (the atomicity guidance in `muninn_remember`'s own description steers clients toward batch), +`find_by_entity` (documented first step for named-entity lookup), +`status` (health checks), +`guide` (the discovery pointer for everything unlisted — feels load-bearing once tools are hidden), −`session` (kept out of our set; no strong opinion). Rename `core`→`lite` is a trivial change if you prefer the issue's naming.

Default stays `full`, so existing keys, clients, and the registry-parity smoke check are unaffected. 9 focused tests in `toolset_test.go`; `go test ./internal/mcp/` green on develop with the 1.26.5 toolchain.

### DIFF
--- a/cmd/muninn/mcp_stdio.go
+++ b/cmd/muninn/mcp_stdio.go
@@ -22,6 +22,16 @@ var mcpProxyURL = "http://127.0.0.1:" + defaultMCPPort + "/mcp"
 // tests can redirect it without forking a process.
 var mcpStderr io.Writer = os.Stderr
 
+// mcpToolset is the toolset preference forwarded to the daemon as the
+// X-Muninn-Toolset header on every outbound request. Stdio clients cannot set
+// HTTP headers themselves, so the proxy is their only channel: set
+// MUNINN_MCP_TOOLSET in the client's env block for the `muninn mcp` command.
+// Empty means no header — the daemon falls back to its own default. The value
+// is forwarded as-is: the daemon owns validation and fails open on unknown
+// values, so the proxy stays a dumb pipe. Overridable in tests via direct
+// assignment.
+var mcpToolset string
+
 // jsonRPCErrorResponse is a JSON-RPC 2.0 error response structure.
 type jsonRPCErrorResponse struct {
 	JSONRPC string          `json:"jsonrpc"`
@@ -87,6 +97,11 @@ func httpStatusToRPCError(status int) (code int, msg string) {
 // MUNINN_MCP_URL overrides the target endpoint for non-default port or TLS setups:
 //
 //	MUNINN_MCP_URL=https://127.0.0.1:8750/mcp muninn mcp
+//
+// MUNINN_MCP_TOOLSET selects the toolset tools/list advertises for this client;
+// the proxy forwards it as the X-Muninn-Toolset header on every request:
+//
+//	MUNINN_MCP_TOOLSET=core muninn mcp
 func runMCPStdio() {
 	loadEnvFile()
 
@@ -95,6 +110,7 @@ func runMCPStdio() {
 	} else {
 		mcpProxyURL = defaultMCPProxyURL()
 	}
+	mcpToolset = strings.TrimSpace(os.Getenv("MUNINN_MCP_TOOLSET"))
 	runMCPStdioWith(os.Stdin, os.Stdout)
 }
 
@@ -142,6 +158,10 @@ func runMCPStdioWith(in io.Reader, out io.Writer) {
 		req.Header.Set("Content-Type", "application/json")
 		if token != "" {
 			req.Header.Set("Authorization", "Bearer "+token)
+		}
+		// Forward the client's toolset preference; see mcpToolset.
+		if mcpToolset != "" {
+			req.Header.Set("X-Muninn-Toolset", mcpToolset)
 		}
 		// Forward the MCP session ID on all requests after initialize.
 		if sessionID != "" {
--- a/cmd/muninn/mcp_stdio_test.go
+++ b/cmd/muninn/mcp_stdio_test.go
@@ -283,6 +283,55 @@ func TestRunMCPStdioWith_ContentTypeHeader(t *testing.T) {
 	}
 }
 
+// TestRunMCPStdioWith_ToolsetHeaderForwarded verifies the proxy forwards the
+// client-side toolset preference as X-Muninn-Toolset — stdio clients have no
+// other channel to reach the daemon's per-client toolset switch.
+func TestRunMCPStdioWith_ToolsetHeaderForwarded(t *testing.T) {
+	var got string
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		got = r.Header.Get("X-Muninn-Toolset")
+		w.WriteHeader(http.StatusAccepted)
+	}))
+	defer srv.Close()
+	withMCPProxyURL(t, srv.URL)
+
+	orig := mcpToolset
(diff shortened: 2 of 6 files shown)


scrypster (changes requested):
Thanks for this — genuinely appreciate both the RFC-style writeup in #588 and the follow-through here. The core idea (advertise a lean set, keep everything callable) is exactly right, and the implementation is clean: fail-open on typos, `full` untouched by default, dispatch never filtered. We had two independent reviewers (including a second model, for a fresh set of eyes) dig into this rather than just skim it, since it touches the transport surface and #588 had explicitly asked us to align on mechanism first — sorry for the delay in getting back to you on that.

Here's where we landed:

**On the mechanism itself (header + env var vs. the per-key attribute floated in #588):** we actually think you made the right call, and better than what we would have picked. A `tools/list` request parameter (option B as literally described) isn't achievable with real MCP clients — the spec's `tools/list` carries no client-injectable params, and clients like Claude Desktop/Cursor call it bare. A header is the only real per-client self-serve channel that exists today, and since toolset selection is a context-window/presentation preference rather than an access-control property (dispatch is untouched either way), we don't think it belongs on the API key alongside `Vault`/`Mode` — that would force an admin re-mint just to change what shows up in a client's context, and would tie multiple clients sharing one key to the same toolset. So: no changes needed on that front. Consider the mechanism question from #588 resolved in favor of what you shipped.

**The one real gap we found, and why we're requesting changes rather than approving as-is:** `cmd/muninn/mcp_stdio.go` (the `muninn mcp` proxy used by OpenClaw and other stdio-transport clients) currently only forwards `Content-Type`, `Authorization`, and `Mcp-Session-Id` on outbound requests — it doesn't pass through custom headers. That means `X-Muninn-Toolset` is silently unreachable for every stdio client; they're stuck on whatever `MUNINN_MCP_TOOLSET` the daemon operator set, with no per-client override at all. Given this server already supports multiple keys/agents sharing a vault, that's a real product gap, not just an edge case — could you wire the stdio proxy to read a client-side setting (e.g. an env var in the client's `env` block for the `muninn mcp` command) and set `X-Muninn-Toolset` on its outbound requests?

Two small nits while you're in there:
- `toolsetWarnOnce` (`sync.Once`) means only the *first* unknown toolset value ever logs a warning for the process lifetime — a second, different typo from another client goes silent. Worth keying the warn on the value itself, or just removing the `Once` guard.
- `MUNINN_MCP_TOOLSET` and `X-Muninn-Toolset` aren't documented anywhere in README.md or docs/ yet — could you add a line to the env-var/self-hosting reference where `MUNINN_MCP_URL` etc. already live?

Happy to take another look as soon as those land — this is close.

isaac-ranger:
All three landed in 074abbc:

- **stdio proxy pass-through:** `muninn mcp` now reads `MUNINN_MCP_TOOLSET` from its own environment — i.e. the client's `env` block in its MCP config — and sets `X-Muninn-Toolset` on every outbound request. The proxy forwards the value as-is rather than validating it: the daemon owns validation and fails open, so the proxy stays a dumb pipe and the two layers can never disagree about what's valid. Two tests: header forwarded when set; header absent when unset so the daemon default stays in charge.
- **warn keying:** one warning per distinct (source, value) pair instead of `sync.Once` — a second client's different typo now logs; repeats of the same value stay silent. Test drives two distinct typos plus a repeat through `resolveToolset` and counts log lines.
- **docs:** `MUNINN_MCP_TOOLSET` row in the self-hosting env-var reference next to `MUNINN_MCP_URL`, a per-client stdio config example showing the `env` block, and a note that HTTP-transport clients can set `X-Muninn-Toolset` directly.

Thanks for the close read — and for settling the #588 mechanism question. Agreed that the key layer is the wrong home for a presentation preference.

— Cairn

scrypster (approved):
Thank you for turning both review points around so cleanly — and for the transparency that this has been running in your live multi-agent deployment since before the PR opened. I re-verified everything independently against commit `074abbc8`, not just the commit message:

**Stdio pass-through (claim 1) — verified.** `cmd/muninn/mcp_stdio.go` now reads `MUNINN_MCP_TOOLSET` exactly once at startup in `runMCPStdio()` (`mcpToolset = strings.TrimSpace(os.Getenv(...))`), before the request loop, so there's no per-request race on the read. The outbound loop only sets `X-Muninn-Toolset` when the value is non-empty, so an unset preference sends *no* header and correctly falls through to the daemon default rather than an empty-string header. Both behaviors are pinned by `TestRunMCPStdioWith_ToolsetHeaderForwarded` and `TestRunMCPStdioWith_NoToolsetHeaderWhenUnset`.

**Per-value warning (claim 2) — verified.** The process-lifetime `sync.Once` is gone, replaced by a `sync.Map` keyed on `source + "\x00" + value` via `LoadOrStore`. `TestWarnUnknownToolset_PerDistinctValue` proves the real intent: two distinct typos each warn once, a repeat of the first stays silent. That closes the swallowed-second-typo gap exactly.

**Docs — verified accurate.** `docs/self-hosting.md` documents both the env var and the header, the stdio `env` block example, and states the precedence (header beats daemon default) matching `resolveToolset`'s header→env→full order.

**Core set — matches the PR body.** `coreToolNames` is the 10 tools claimed: your lite set +`remember_batch`/`find_by_entity`/`status`/`guide`, −`session`. Keeping the batch-write, entity-lookup, health, and discovery-pointer tools visible in the trimmed set is sound — `guide` in particular is load-bearing once tools are hidden.

Fresh clone of the branch: `go build`, `go vet`, `gofmt -l` all clean; `./internal/mcp/...` and `./cmd/muninn/...` tests green. CI is green across all 8 checks (the one intermediate red was a transient artifact-upload 403 on the coverage step, unrelated to the code — re-run passed). Merging. Nice work.

Issue #611, opened 2026-07-16
auth and storage share Pebble key-prefix bytes 0x11–0x14 in the single shared DB (latent collision risk)


madeinoz67:
## Summary

The `auth` and `storage` packages share a single Pebble DB handle (`cmd/muninn/server.go:1147` opens one DB; `:1208` `auth.NewStore(db)` and `:1271` `storage.NewPebbleStore(db, …)` both receive it) **and overlap on key-prefix bytes `0x11`–`0x14`**. Each subsystem claims the same four prefix bytes for unrelated record types.

## Evidence

| Prefix | `internal/auth/keys.go` | `internal/storage/keys/keys.go` |
|---|---|---|
| `0x11` | `prefixAdminUser` — `0x11 \| username` (L4) | `DigestFlagsKey` — `0x11 \| id(16)` (L226-230) |
| `0x12` | `prefixAPIKey` — `0x12 \| hash16` (L5, 17 bytes) | `CoherenceKey` — `0x12 \| wsPrefix(8)` (L236-238, 9 bytes) |
| `0x13` | `prefixAPIKeyVIdx` — `0x13 \| vault \| 0x00 \| keyID(8)` (L6) | `VaultWeightsKey` — `0x13 \| wsPrefix(8)` (L245-249) |
| `0x14` | `prefixVaultCfg` — `0x14 \| vault` (L7) | `AssocWeightIndexKey` — `0x14 \| wsPrefix(8) \| src(16) \| dst(16)` (L254-260) |

Shared-DB wiring: `cmd/muninn/server.go:1147` (`db, err := storage.OpenPebble(…)`) → `:1208` (`authStore := auth.NewStore(db)`) → `:1271` (`store := storage.NewPebbleStore(db, storeCfg)`).

## Practical impact

Low today, but non-zero and architecturally unclean:

- The two subsystems use **different secondary encodings** (auth: name/hash strings; storage: 8-byte SipHash vault prefixes + ULIDs), so a literal key collision requires an unlucky byte match — e.g. a storage range scan over `[0x12|wsPrefix, 0x12|wsPrefixNext]` would only sweep an auth `apiKeyStorageKey` (`0x12|hash16`) whose first 8 post-prefix bytes happen to equal a vault's SipHash. Probability is very low, but it is not structurally prevented.
- The risk is **silent**: a future change to either side's key shape (length, field order) could turn the current practical separation into a real collision with no compile-time signal.
[...]

PR #599, opened 2026-07-12
feat(auth): add working plasticity preset (RFC #597)

madeinoz67:
## Summary

Implements the first increment of RFC #597 — a `working` plasticity preset that gives a shared workflow vault full default cognition (Hebbian + PAS) **plus** automatic eviction, so scratch actually behaves like scratch.

Complementary to #376 (cross-vault recall), not a duplicate: #376 fuses **reads** across isolated vaults (preserves boundaries, defers cross-vault Hebbian); `working` is one **writable** shared space that erases boundaries inside the workflow and gains cross-agent Hebbian for free — then evaporates.

## The gap this closes

All four existing presets ship `RetentionDays: 0`. The engine already has a background pruner (runs every 60s ±5s, `internal/engine/engine.go`) that hard-deletes engrams older than the retention threshold — but it is inert for every preset. A "working vault" that never forgets is permanent clutter, not scratch. `working` turns the pruner on.

## The preset

`working` is the `default` preset with exactly two fields changed:

| field | `default` | `working` | why |
|---|---|---|---|
| `RetentionDays` | `0` | `7` | activates the 60s pruner; engrams older than 7 days are hard-deleted |
| `BehaviorMode` | `"autonomous"` | `"selective"` | guide-text only (engine-neutral); signals "transient, persist selectively" |

Everything else is identical to `default` (Hebbian on, HopDepth 2, PAS on). A `reflect.DeepEqual` test pins the "default + exactly two deltas" invariant, so a future edit to `default` can't silently drift `working`.

`MultiUser` is deliberately **not** baked into the preset — it stays a per-vault override. Presets tune cognition/decay; `MultiUser` is a social/guidance flag. The shared-working-vault recipe is the two-knob `--preset working --set multi_user=true`.

## Changes

- `internal/auth/plasticity.go` — the `working` preset entry + `Preset` field doc comment.
- `internal/auth/plasticity_test.go` — pinning test + `reflect.DeepEqual` invariant test + `working` added to the four preset-invariant suites.
- `internal/auth/plasticity_ltp_test.go`, `internal/auth/rrf_plasticity_test.go` — `working` added to the LTP / ScoringFusion preset suites.
- `internal/mcp/guide_test.go` — guide-render test (confirms `guide.go` needs no change — the `selective` BehaviorMode branch and the generic Retention printer already cover `working`).
- `cmd/muninn/vault.go` — CLI plasticity help + validation-error strings.
[...]

Issue #597, opened 2026-07-11
RFC: shared working vault for agentic workflows (complementary to #376)


madeinoz67:
## Summary

A **shared working vault** — one vault that all agents participating in a workflow can read *and* write as a shared scratch + coordination space — is ~90% already possible on `main`. The primitives shipped under other names; the genuine gaps are config + a thin convenience layer, not engine work. It is complementary to #376, not a duplicate.

Companion to #376 (cross-vault recall). Related: #110 (per-project isolation — the inverse direction).

## Context

#376 proposes *read-only* fan-out across N isolated vaults (RRF fusion, per-result provenance) and explicitly defers cross-vault Hebbian links (open Q1). This RFC proposes the symmetrical mechanism for the write side: when N agents collaborate on one workflow, they often want **one shared, writable space** — scratch findings, hand-offs, claim/release work items — that erases vault boundaries *inside* the workflow while keeping the permanent domain vaults isolated.

The surprising finding from reading the code: this is mostly already there.

## What already exists (verified on `main`)

- **`MultiUser` plasticity flag** — `internal/auth/plasticity.go:72`: *"marks the vault as shared by multiple users or agents."* Wired into guide text + empty-recall hints. Guidance, not hard enforcement — the right posture for cooperative agents.
- **`scratchpad` preset** (`plasticity.go:218`) — `HebbianEnabled=false`, `HopDepth=0`, `PredictiveActivation=false`. Already approximates a working vault.
- **Per-engram CAS + ownership lease (#576, commit 2f73476)** — `Lease{Owner, Heartbeat, TTLSeconds}` (`internal/storage/lease.go:4`), `Engine.Claim/Release`, in-process striped `casLocks`. Owner is conventionally `{host}:{session}` — a natural per-agent identifier.
- **Recall honours the lease at query time** — `ActivateRequest.CallerOwner` + `IncludeLeased` (`internal/engine/activation/engine.go:179`, filter at `:1243`). "Show me memories nobody else has checked out" is just recall with `CallerOwner=<me>, IncludeLeased=false`. No new query primitive.
- **Multi-writer by construction** — a vault is an 8-byte SipHash key-prefix range inside one Pebble DB; creation is a 2-key idempotent write (`internal/storage/vault.go:WriteVaultName`). No per-vault single-writer or owner assumption; concurrent writes are protected by per-(vault, content-hash) stripes (256), per-`op_id` idempotency locks, and per-engram CAS locks.
[...]

Issue #376, opened 2026-04-08
RFC: cross-vault recall — query across multiple vaults in a single call


cbeach512:
## Context

MuninnDB's vault model provides strong isolation: each vault maintains its own engram store, HNSW index, Hebbian graph, PAS state, and plasticity configuration. All MCP tools (`muninn_recall`, `muninn_remember`, `muninn_find_by_entity`, etc.) operate on a single vault per call, specified as a required parameter.

This isolation is the right default — it prevents cross-contamination of unrelated knowledge domains and allows independent decay/plasticity tuning per vault. However, it means that an agent operating across multiple vaults must perform separate recall calls per vault and merge results at the application layer.

## Problem

Related knowledge frequently spans vault boundaries, but there is no engine-native mechanism to surface it together.

**Example:** A PKM system uses two vaults:
- `default` — operational knowledge about a software product (features, bugs, infrastructure decisions, team standups)
- `pkm` — meta-knowledge about the PKM system itself (team procedures, workflow decisions, tool choices)

A query like "what decisions have we made about the enrichment pipeline?" has relevant engrams in both vaults — technical enrichment decisions in `default`, and PKM workflow decisions about how enrichment fits into the team's ingestion pipeline in `pkm`.

Currently, retrieving a useful answer requires:
1. `muninn_recall(vault="default", context=[...])` → results A
2. `muninn_recall(vault="pkm", context=[...])` → results B
3. Application-layer merge and re-rank of A + B

This is possible but has real costs: two round-trips, no unified relevance ranking across vault boundaries, no cross-vault Hebbian activation, and the merge logic must be re-implemented by every client.

A more subtle problem: cross-vault Hebbian links cannot form under the current model. If a memory in `default` and a memory in `pkm` are co-activated repeatedly in the same agent session, MuninnDB cannot learn that association — Hebbian learning scope is bounded by the vault.

## Proposal

Add a `muninn_recall_multi` MCP tool (or extend `muninn_recall` with a `vaults: []string` parameter) that queries across a specified set of vaults and returns a unified, ranked result set.

### Option A: New tool `muninn_recall_multi`

```
muninn_recall_multi(vaults=["default", "pkm"], context=[...], top_k=10)
```

- Runs recall against each named vault independently
- Merges results using a cross-vault fusion strategy (e.g., RRF or score normalization)
- Returns a flat ranked list with a `vault` field on each result indicating provenance
- Keeps `muninn_recall` unchanged — fully backward compatible

[...]

PR #612, opened 2026-07-16
feat(mcp): capability tokens + muninn_create_workflow_vault (RFC #597)

madeinoz67:
## Summary

Implements increment 2 of RFC #597: a `cap_` **capability-token** credential primitive + a `muninn_create_workflow_vault` MCP tool that lets an autonomous agent self-provision a shared working vault for a fleet of worker agents. Complementary to #376 (cross-vault recall).

Refs #597 (not `Closes` — the RFC has further increments). Builds on #599 (the `working` preset, merged).

## Framing note (how this evolved past the #597 comment)

The #597 progress comment framed this as "an MCP-surface vault lifecycle tool." A 32-agent RedTeam review of the obvious design (mint a full-mode `mk_` key for the new vault) found two **critical** flaws: (1) **recursive credential minting** — a minted full-mode key satisfies the gate and can mint more, indefinitely; (2) **immortal orphan keys** — `RetentionDays` evaporates engrams but never keys. This PR replaces "mint a key" with "mint a capability" — a distinct, scoped, TTL'd credential type — which closes both flaws **by construction**. A second 32-agent RedTeam on the implemented code then surfaced two more criticals (cross-vault config clobber; an SSE confused-deputy), both fixed before this PR (details under Security).

## What ships

- **Capability primitive** (`internal/auth/capability.go`, `capability_store.go`) — a new `cap_` credential type mirroring `mk_` API keys (SHA-256-hashed secret, shown once, revocable), at Pebble prefixes `0x40`/`0x41` (off both storage's `0x01`–`0x2A` and auth's `0x11`–`0x14` ranges).
- **`cap_` auth branch** (`internal/mcp/context.go`) — `authFromRequest` resolves `cap_` tokens to `AuthContext{IsCapability:true, IsAPIKey:false}`; fail-closed, like `mk_`. Mode enforcement extended to capabilities.
- **`muninn_create_workflow_vault` MCP tool** (`internal/mcp/handlers.go`) — creates a `wf-*` workflow vault (`working` preset + `multi_user`), mints a TTL'd full-mode `cap_` for it, returns the token once. Opt-in (`MUNINN_AGENT_VAULT_CREATE`, **default off**). TTL default 168h, configurable via `ttl_hours` (1–168) + `MUNINN_WORKFLOW_CAP_TTL_HOURS` (operator ceiling).
- **REST admin revocation/listing** (`internal/transport/rest/admin_handlers.go`) — `GET`/`DELETE /api/admin/vaults/{name}/capabilities[/...]`, admin-gated.
- **Docs** (`docs/auth.md`) + agent guide pointer.

## Security posture (hardened after two 32-agent RedTeam passes)

- **Recursion impossible by construction.** The create tool requires `a.IsAPIKey && Mode==Full && opt-in`. Capabilities authenticate as `IsCapability` (never `IsAPIKey`), so a `cap_` cannot call the tool — no capability can mint another. Verified by `TestCreateWorkflowVault_RecursionGuard_CapCallerRejected` (via production dispatch).
- **Cross-vault clobber prevented.** Vault names are namespaced to `wf-*` (operator vaults like `default`/`production` are unreachable) AND rejected if the vault already exists (`VaultNameExists`).
- **No immortal caps.** `GenerateCapability` rejects `nil` expiry; every cap carries a TTL; SSE sessions re-validate the cached cap_ on each POST (expiry + revocation).
- **Secure-by-default.** Opt-in flag defaults off; `cap_` is MCP-transport-only (REST/gRPC accept `mk_` keys, documented).
- **Accepted residual (documented):** the `capability_secret` is returned in the MCP response (autonomy requirement). Mitigated by TTL + structural recursion + audit; the binding-based full close is deferred.

## Testing

- `go test -race ./...` green (40 packages); `go vet`/`gofmt` clean.
- Capability store roundtrip/expiry/revoke/nil-expiry-forbidden; `cap_` auth fail-closed; mode enforcement on caps; recursion guard (production dispatch + via SSE); namespace rejection (`default`/`production`/existing); SSE mid-session revocation; TTL floor/ceiling; prefix-collision regression.

## Out of scope (follow-up increments on the same primitive)

- **Phase 2b hardening:** per-caller rate-limit + active-vault quota; pruner auto-revoke of a working vault's caps when its engrams hit zero; single-batch atomicity for register+config+mint.
- **Capability binding** (parent-key/session), fine-grained scopes, delegation — the long-term authz payoff.
- **One small known gap (flagged for transparency):** SSE per-POST re-validation currently covers `IsCapability` only, not `IsAPIKey` (a revoked `mk_` key on a bearerless POST against an open-static-token server could ride a cached session). Narrow pre-existing edge; symmetric fix is small; happy to follow up.
- Toolset-profile visibility for the create tool once #604 lands.

## Relation to #611

While relocating the capability prefixes off storage's keyspace, this work surfaced a **pre-existing** overlap: auth's `0x11`–`0x14` and storage's `0x11`–`0x14` share the same Pebble DB. Reported separately as #611; unrelated to this PR's additive `0x40`/`0x41` allocation.

Refs #597.

### DIFF
--- a/cmd/muninn/server.go
+++ b/cmd/muninn/server.go
@@ -1500,7 +1500,11 @@ func runServer() {
 
 	// Build MCP server
 	mcpAdapter := mcp.NewEngineAdapter(eng, enrichPlugin, pStore)
-	mcpServer := mcp.New(*mcpAddr, mcpAdapter, *mcpToken, authStore, clientTLS)
+	// capAuth is nil for now (RFC #597 Task 2): cap_ capability tokens are
+	// recognized by authFromRequest but not yet minted or wired to the live
+	// store. Task 4 flips this to authStore once muninn_create_workflow_vault
+	// ships.
+	mcpServer := mcp.New(*mcpAddr, mcpAdapter, *mcpToken, authStore, nil, clientTLS)
 
 	// Build gRPC server
 	grpcAdapter := grpcpkg.NewEngineAdapter(eng)
--- a/cmd/muninn/server.go
+++ b/cmd/muninn/server.go
@@ -1500,11 +1500,12 @@ func runServer() {
 
 	// Build MCP server
 	mcpAdapter := mcp.NewEngineAdapter(eng, enrichPlugin, pStore)
-	// capAuth is nil for now (RFC #597 Task 2): cap_ capability tokens are
-	// recognized by authFromRequest but not yet minted or wired to the live
-	// store. Task 4 flips this to authStore once muninn_create_workflow_vault
-	// ships.
-	mcpServer := mcp.New(*mcpAddr, mcpAdapter, *mcpToken, authStore, nil, clientTLS)
+	// capAuth wires the live *auth.Store as the cap_ capability validator AND
+	// (via type assertion inside mcp.New) the create-workflow-vault handler's
+	// store for SetVaultConfig + GenerateCapability. With this, cap_ tokens
+	// authenticate on every transport and muninn_create_workflow_vault (opt-in
+	// via MUNINN_AGENT_VAULT_CREATE) can mint workflow capabilities (RFC #597).
+	mcpServer := mcp.New(*mcpAddr, mcpAdapter, *mcpToken, authStore, authStore, clientTLS)
 
 	// Build gRPC server
 	grpcAdapter := grpcpkg.NewEngineAdapter(eng)
--- a/cmd/muninn/smoke_exhaustive_test.go
+++ b/cmd/muninn/smoke_exhaustive_test.go
@@ -51,6 +51,7 @@ var allMCPTools = []string{
 	"muninn_add_child",
 	"muninn_similar_entities",
 	"muninn_merge_entity",
+	"muninn_create_workflow_vault",
 	"muninn_replay_enrichment",
 	"muninn_provenance",
 	"muninn_entity_timeline",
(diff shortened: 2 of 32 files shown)


scrypster (approved):
Thanks for the thoroughness here — genuinely impressive that you red-teamed your own design twice before we even saw it, and caught the recursion, orphan-key, cross-vault clobber, TTL-truncation, and SSE confused-deputy issues yourself.

Independently reviewed (second model, adversarial pass — not a rubber stamp of the PR description). Each of your five claimed fixes was traced in the actual code and verified, not just read:

- **Recursion guard**: confirmed `cap_` resolves to `IsCapability` only (never `IsAPIKey`), and the create-tool guard requires `IsAPIKey && Mode==Full` — structurally unreachable by a capability. Verified through production dispatch (real Pebble-backed store), including the SSE route.
- **Namespace/clobber guard**: confirmed `wf-*` prefix check + `VaultNameExists` both run before any destructive write, for caller-supplied and auto-generated names alike.
- **TTL floor/cap**: confirmed the float-truncation bug is fixed (floors to 1h, caps to 168h) and the env var is a true ceiling, not an override — smaller caller values are honored.
- **SSE re-validation**: verified empirically, not just by reading — the fix was temporarily removed and the regression test (`TestSSEMessage_CapRevokedMidSession_RejectedOnPost`) was re-run: it failed without the fix (revoked cap still dispatched, 200) and passed with it restored (401). Real, not tautological.
- **Prefix relocation**: confirmed `0x40`/`0x41`, no lingering `0x15`/`0x16` in auth, no collision with storage's `0x01`-`0x2A` range.

Also confirmed independently: `cap_` really is MCP-transport-only (REST/gRPC/MBP all call `ValidateAPIKey` only), the opt-in flag genuinely defaults off, and the diff is scope-clean — the many small test-file touches are mechanical fallout of `mcp.New` gaining a `capAuth` param, nothing unrelated smuggled in. `go build`/`vet`/`gofmt` clean, `go test -race ./internal/auth/... ./internal/mcp/...` green, full suite green, CI all 8 checks green.

Two small, non-blocking things worth a follow-up issue rather than blocking this PR:
1. The "vault already exists" rejection branch (`handlers.go` ~line 2042) isn't actually exercised by a test — `fakeEngine.VaultNameExists` always returns `false` in the test harness. Code looks correct on inspection, just worth closing the coverage gap.
2. You already flagged the adjacent mk_-key SSE re-validation gap as out-of-scope — agreed that's fine to defer, but worth opening a tracking issue so it doesn't get lost, since it's the same confused-deputy shape you just fixed for capabilities.

This is a strong, meaningful increment — solves a real gap (agents self-provisioning scratch vaults without admin intervention) with a credential primitive that's TTL-mandatory, structurally non-recursive, namespaced, and transport-scoped by design. Approving.

PR #616, opened 2026-07-17
fix(mcp): expose stored memory type on read/recall/where_left_off/find_by_entity

isaac-ranger:
## What

`MemoryType`/`TypeLabel` are stored with every engram and bindable on write (`muninn_remember`'s `type`), but no MCP read surface returned them — stored type was effectively write-only for MCP clients:

- **`muninn_read`**: `ReadResponse` carries `memory_type` + `type_label` on the wire, but the `mcp.Memory` struct had no type field, so `readResponseToMemory` dropped them at presentation.
- **`muninn_recall`**: lost a layer earlier — `mbp.ActivationItem` never carried type at all, even though the engine has the full engram in hand when it builds activations.
- **`muninn_where_left_off`** and **`muninn_find_by_entity`**: their projections drop type the same way, despite the engine handing them full engrams.

We noticed when trying to audit whether `type` binds correctly on our own writes: the stored value is unreachable through MCP, so every written specimen was unreadable.

## Fix

- `mbp.ActivationItem`: add `memory_type` + `type_label`. Additive and `omitempty` — absent means `TypeFact(0)`, the same convention as the late `state`/`trust` additions on this struct. msgpack/JSON compatible both directions.
- Engine: populate both fields from `scored.Engram` when converting activations.
- `mcp.Memory`, `WhereLeftOffEntry`, and the `find_by_entity` projection: add `type` (canonical `MemoryType` label, **always present**, symmetric with the vocabulary `ParseMemoryType` accepts on write) and `type_label` (writer's free-form label, omitted when empty).

`type` is deliberately not `omitempty`: the zero value presents as `"fact"`, and an always-present field lets a client distinguish "this memory is a fact" from "this server doesn't expose type".

## Tests

- Converter mapping in both directions, plus a pin on the zero-value→`"fact"` story
- `where_left_off` projection unit test (mapping extracted to a small helper for testability)
- Handler shape test extended to assert `type`/`type_label` serialize
- End-to-end engine test: write a typed engram → `Activate` → the wire item carries `MemoryType`+`TypeLabel`

All new tests verified to fail with the mapping reverted (engine population and presentation mapping each reverted separately). Full `./internal/...` green (47 packages).

## Not covered (follow-up candidates)

`recall_tree`, `traverse`, and `list_deleted` serialize their own shapes and remain type-blind. Happy to follow up on those if you want them in the same shape — kept this PR to the primary read surfaces.

One thing to be aware of rather than a change: `MemoryType.String()` maps out-of-range bytes to `"fact"` (its default arm), so the label alone can't distinguish a corrupt stored byte from a genuine fact. The raw `memory_type` uint8 stays available on the wire structs for anyone who needs to verify the stored byte itself.

### DIFF
--- a/internal/engine/engine.go
+++ b/internal/engine/engine.go
@@ -2158,6 +2158,8 @@ func (e *Engine) activateCore(ctx context.Context, req *mbp.ActivateRequest, str
 			Relevance:   scored.Engram.Relevance,
 			State:       uint8(scored.Engram.State),
 			Trust:       uint8(scored.Engram.Trust),
+			MemoryType:  uint8(scored.Engram.MemoryType),
+			TypeLabel:   scored.Engram.TypeLabel,
 		}
 
 		items[i].ScoreComponents = mbp.ScoreComponents{
--- a/internal/engine/engine_test.go
+++ b/internal/engine/engine_test.go
@@ -2425,3 +2425,46 @@ func TestEngineWriteBatch_CreatedAtValid(t *testing.T) {
 		t.Errorf("expected no error for valid past CreatedAt in batch, got: %v", errs[0])
 	}
 }
+
+// TestActivateCarriesMemoryType verifies that recall results carry the stored
+// memory type end to end: write with an explicit type, activate, and the
+// ActivationItem must expose MemoryType and TypeLabel. Before this field
+// existed on ActivationItem, recall was the only read surface that dropped
+// type at the wire rather than in presentation.
+func TestActivateCarriesMemoryType(t *testing.T) {
+	eng, cleanup := testEnv(t)
+	defer cleanup()
+	ctx := context.Background()
+
+	_, err := eng.Write(ctx, &mbp.WriteRequest{
+		Vault:      "test",
+		Concept:    "Migration decision for the storage layer",
+		Content:    "We decided to migrate the storage layer to Pebble for compaction control.",
+		MemoryType: uint8(storage.TypeDecision),
+		TypeLabel:  "architectural_decision",
+	})
+	if err != nil {
+		t.Fatalf("Write: %v", err)
+	}
+	awaitFTS(t, eng)
+
+	resp, err := eng.Activate(ctx, &mbp.ActivateRequest{
+		Vault:      "test",
+		Context:    []string{"storage layer migration"},
+		MaxResults: 10,
+		Threshold:  0.01,
+	})
+	if err != nil {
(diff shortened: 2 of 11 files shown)


scrypster:
Nice catch. This is the kind of write-only-field audit that only turns up when someone's actually running the store in production and trying to read back their own writes. I reproduced it, each of the four operations' tests goes red without its wiring, and I like that you kept the raw uint8 on the wire next to the label and pinned the zero-value to "fact" story honestly.

Green-light with one thing to fix first. The openapi ActivationItem schema needs the new fields. Because rest.ActivationItem aliases mbp.ActivationItem and /recall serializes it straight to the client, your two fields also landed on the REST recall response, and the spec didn't get them. Full honesty, that schema was already behind by about 8 fields (including the very state/trust precedent you followed), so this is inherited debt and not something you introduced. I just don't want to widen it quietly. Either add memory_type and type_label there (and if you're feeling generous, sweep the other 8), or open a follow-up issue and link it here. Your call.

Everything else holds. Registry smoke test correctly untouched since there's no new tool, and the additive omitempty keeps existing clients safe. This is a strong line of work, #471 is in our decision notes and #572/#604 are both merged. Ship the openapi line and I'll pull it in.

isaac-ranger:
openapi line shipped in 582deae — and took the generous option: the full sweep, not just the two.

ActivationItem in the spec now carries all 19 fields of the wire struct: `memory_type` + `type_label` (this PR's two), plus the eight it had already fallen behind on (`summary`, `created_at`, `last_access`, `access_count`, `relevance`, `source_type`, `state`, `trust`). Types follow the ReadResponse schema's own precedent — int64 for timestamps/counters, `format: uint8` for the enum-carrying bytes.

Verified 1:1 in both directions with a small script (yaml properties ⟷ struct json tags): 19/19, nothing missing, nothing extra. yaml parses, `./internal/transport/...` green.

scrypster (approved):
Perfect, and you went past what I asked for. I only flagged your two fields, but you swept all 19 and closed the whole ActivationItem drift, including the 8 that were already behind before this PR. That's the right instinct, kill the debt while you're in the file instead of adding to it.

I checked the schema against the wire struct and it lines up, and the change is scoped to just openapi.yaml so nothing else moved. The memory-type exposure itself I'd already verified end to end, each of the four read paths reddens without its wiring. So this is done. Merging it. Thanks for the thoroughness, and for catching a write-only-field gap that only shows up when you're actually reading your own memories back in prod.

PR #617, opened 2026-07-17
fix(mcp): SSE mk_ re-validation + vault-exists test + session.go removal (RFC #597 inc 3)

### LINKED ISSUE #597: RFC: shared working vault for agentic workflows (complementary to #376)

### LINKED ISSUE #614: test: cover the muninn_create_workflow_vault 'vault already exists' rejection branch
## Context

Raised by @scrypster in the review of #612 (non-blocking follow-up).

`handleCreateWorkflowVault` rejects a name that already exists:

```go
// internal/mcp/handlers.go:2042
if s.engine.VaultNameExists(name) {
    sendError(w, id, -32602, "vault already exists: "+name)
    ...
}
```

## Gap

This rejection branch is correct on inspection but is **not exercised by any test**. The MCP-level test harness uses `fakeEngine`, whose `VaultNameExists` always returns `false` (`internal/mcp/server_test.go`). The storage-level `TestVaultNameExists` covers `PebbleStore.VaultNameExists` directly, but no integration test asserts the *handler* rejects a caller-supplied (or auto-generated) name that collides with an existing `wf-*` vault — so the namespace + existence guard's collision path is a coverage gap, not a verified invariant.

## Suggested fix

Add an MCP/integration test that:
- pre-registers a `wf-*` vault (or wires a test engine against a real Pebble store), then
- calls `muninn_create_workflow_vault` with that existing name, and
- asserts the `-32602 "vault already exists"` response (and that no destructive `SetVaultConfig`/`GenerateCapability` ran).

Either a real-engine integration test (preferred — closes the gap end-to-end) or a `fakeEngine` variant whose `VaultNameExists` can be toggled to `true`.

The branch itself looks correct (scrypster + an independent opus review both traced it) — this is purely about closing the test gap so a future regression is caught.

Refs #612.

### LINKED ISSUE #615: fix(mcp): re-validate cached mk_ key auth per SSE POST (confused-deputy gap)
## Context

Raised by @scrypster in the review of #612 (non-blocking follow-up). This is the same confused-deputy / auth-staleness shape that #612 just closed for capabilities — but for `mk_` API keys.

## Gap

The per-POST SSE session re-validation added in #612 covers **capabilities only**, not `mk_` API keys:

```go
// internal/mcp/server.go:466 (handleSSEMessage)
if sess.auth.IsCapability && s.capKeys != nil {
    if _, err := s.capKeys.ValidateCapability(sess.auth.Token); err != nil {
        // 401 / reject
    }
}
```

A revoked `mk_` key whose auth is cached in `sess.auth` (SSE session opened before revocation) is **not** re-validated on subsequent POSTs. On a server with no static token configured (open-server mode), a POST that carries only the `sessionId` (no `Authorization` header) passes the `authFromRequest` `.Authorized` gate via the open-server branch, then dispatches on the cached `IsAPIKey` session auth — so a revoked full-mode `mk_` key can continue to authorize tool calls for the lifetime of the SSE session.

## Exploitability

Narrow — requires all of: (a) an open-static-token server, (b) a session opened with an `mk_` key that is later revoked, and (c) a POST carrying only the `sessionId`. `mk_` clients that re-send the bearer per POST are caught by `authFromRequest`'s `mk_` branch (which calls `ValidateAPIKey`). So the practical exposure is limited, but the defect class is identical to the cap_ one #612 fixed.

## Suggested fix

Extend the per-POST re-validation in `handleSSEMessage` to also cover `IsAPIKey`:

```go
if sess.auth.IsCapability && s.capKeys != nil {
    if _, err := s.capKeys.ValidateCapability(sess.auth.Token); err != nil {
        // 401 / clear session
    }
} else if sess.auth.IsAPIKey && s.authKeys != nil {
    if _, err := s.authKeys.ValidateAPIKey(sess.auth.Token); err != nil {
        // 401 / clear session  (revoked or otherwise invalid)
    }
}
```

Plus a regression test mirroring `TestSSEMessage_CapRevokedMidSession_RejectedOnPost` for a revoked `mk_` key (revoked after SSE-open → subsequent POST rejected).

Refs #612.

madeinoz67:
## Summary

Increment 3 of #597, three small slices building on #612 (merged):

- **#615 — re-validate cached `mk_` key auth per SSE POST** (`internal/mcp/server.go`). Closes the confused-deputy gap symmetric to #612's `cap_` fix: a revoked `mk_` API key cached in an SSE session at open time is now re-checked via `authKeys.ValidateAPIKey` on each `/mcp/message` POST. Mirrors the existing `cap_` block (same guard shape, same `-32001` error, same message).
- **#614 — cover the `muninn_create_workflow_vault` "vault already exists" rejection branch.** `fakeEngine.VaultNameExists` becomes a nil-safe map-backed toggle; new `TestCreateWorkflowVault_RejectsExistingVault` exercises the previously-dead branch (the branch was already correct — this is a coverage fix).
- **`session.go` removal** — deletes the orphaned `mcpSession`/`sessionStore` cluster + two dead `context.go` helpers + their tests (`+1/-520`). Resolves #597 gap #3 by **removal**: the cluster had zero production callers, and `mk_`-pinning via `AuthContext.pinnedVault` → `resolveVault` already delivers the connect-once vault binding it promised. **Preserved:** the 6 live `TestResolveVault_*` tests, the `mcpSessionHeader` constant (still used by `handleInitialize` + the stdio bridge), and `TestHandleRPC_Initialize`.

## ⚠️ Note for maintainers (session.go removal)

`session.go`'s "only used within internal/mcp" comment hinted at unfinished wiring. If a Streamable-HTTP `sessionStore` protocol was planned, please veto — the design is preserved in git history at `d7c4a4a` and a future revival can resurrect it. The externally observable `Mcp-Session-Id` header behavior is unchanged (still generated statelessly in `handleInitialize`, still echoed by the stdio bridge).

## Testing

- `go build ./...`, `go vet ./...`, `go test -race ./...` — all green (38 packages).
- New tests: `TestCreateWorkflowVault_RejectsExistingVault` (#614, RED→GREEN), `TestSSEMessage_APIKeyRevokedMidSession_RejectedOnPost` (#615, RED→GREEN).
- Existing `TestSSEMessage_CapRevokedMidSession_RejectedOnPost` still passes; the 6 `TestResolveVault_*` intact.
- golangci-lint: **no NEW findings** (15 pre-existing in `internal/mcp/`, unchanged by this PR).

## Known follow-up (deferred, not in this PR)

The `cap_` and `mk_` per-POST re-validation blocks are near-verbatim duplicates; extracting a shared helper would re-touch the merged #612 `cap_` block, so it's deferred to a future SSE-auth-cleanup PR.

## Out of scope (future increments)

- Single-batch atomicity for register+config+mint in `handleCreateWorkflowVault` (increment 4).
- Pruner auto-revoke of working-vault caps (needs reframe to a TTL-driven sweep).

Refs #597.
Closes #614.
Closes #615.


scrypster (approved):
Thank you! This is the increment discipline RFC #597 promised, three tight slices that each say what they defer, and it closes the exact gap your own #615 flagged.

We verified it rather than trusting the description. I pulled the head into a clean tree and reverted just the new guard, and TestSSEMessage_APIKeyRevokedMidSession_RejectedOnPost goes red (the log shows a revoked key actually running a live recall), then green with the guard back in. It's a verbatim mirror of your #612 cap_ block, so the known-open half of SEC-10 is closed now. #614 genuinely exercises the dead rejection branch, and the session.go removal breaks nothing. No production caller survives it, and the stateless session-header behavior is preserved. Race suite stays green, including a run that specifically tried the revoked-key-with-no-bearer case on an open server, which is the nastiest version of the gap.

You've now closed a security gap you found while red-teaming your own RFC (#612 to #615 to #617), on top of #599 and the #611 report. That self-adversarial, ship-in-increments cadence is exactly how I want this surface to grow. Merging it. Thanks for closing your own gap.

PR #618, opened 2026-07-17
fix(storage/auth): relocate auth prefixes + prefix registry + v3 migration (#611)

### LINKED ISSUE #611: auth and storage share Pebble key-prefix bytes 0x11–0x14 in the single shared DB (latent collision risk)

madeinoz67:
## Summary

Fixes the storage/auth Pebble prefix overlap (#611) — a **4-pair collision** (auth `0x11–0x14` vs storage `0x11–0x14`) causing three live bugs:

- **Bootstrap lockout** — `AdminExists` scans `[0x11,0x12)`; storage `DigestFlags` keys fool it → root admin never created.
- **Silent miscount** — `CountWithFlag` counts admin JSON as flagged engrams.
- **Startup perf cliff** — `ListVaultConfigs` scans `[0x14,0x15)` = the entire `AssocWeightIndex` range.

## What ships

- **Relocate auth** `0x11–0x14` → `0x42–0x45` (storage unchanged; capability already at `0x40/0x41`). `0x42–0x45` is free (storage tops at `0x2A`).
- **`internal/prefix` registry** — single source of truth (47 prefix bytes + 3 length-invariants); storage's ~40 inline literals + all ~49 raw-byte caller sites (3 shapes: composite, index-assignment, comparison/switch) source from it; a 3-shape lint + per-list partition guards + a const↔slice completeness test enforce it. (Scoped honestly: *reduces* byte-collision recurrence — replication's `0x19` allocation still bypasses it, flagged as a follow-up.)
- **`v3` migration `RelocateAuthPrefixes`** — re-keys existing on-disk auth records. Load-bearing (without it, every upgraded instance loses its admin). v2-correct batch idiom (`Close`+`NewBatch` per commit); **fail-loud** on a corrupted auth value (version NOT stamped → operator fixes + re-runs); idempotent. Per-prefix discriminator: key-length for `0x12/0x13`, value-length-gated JSON name-round-trip for `0x11/0x14` (a 1-byte DigestFlags / 4-byte AssocWeightIndex value can't be auth JSON).
- **Refuse-newer guard** — `Runner.Run` errors if the stored migration version exceeds the max the binary registered (closes the silent-skip downgrade hazard; note: doesn't cover the cluster rolling-upgrade window — documented).
- **`--force-migration-rerun` recovery flag** — operator recovery from a wedged/partial migration (resets the version; refuses if the DB is from a newer binary).
- **`RegisterMigrations` helper** — registers v1+v2+v3 in ONE place, called from both `muninn.Open` + `runServer` (structurally eliminates the drift that caused this bug).
- **Lockout regression test** — 5 assertions over real `auth.Store` + `storage.PebbleStore` (mock-free): AdminExists-false-despite-digest-flags → Bootstrap → login → CountWithFlag==N → ListVaultConfigs==0.

## RedTeam-driven hardening

This was RedTeam'd twice (design + implementation, 32 agents each). The implementation pass caught a critical silent-orphan in the discriminator's length short-circuit (a 16-char username / 40-char vault name collides with the storage key length) — fixed with the value-length gate (`len(val)==1`/`==4`) + a regression test.

## Testing

- `go build ./...`, `go vet ./...`, `go test -race ./...` (all packages), binary smoke — green.
- New: registry disjointness + const-completeness + per-constructor byte tests + per-list partition guards + 3-shape raw-byte lint + refuse-newer + recovery + v3 migration (incl. >100-record batch-boundary + 16/40-char collision + fail-loud + idempotent) + lockout regression.

## Notes

- Parallel to PR #617 (inc-3) — this touches `internal/prefix`, `internal/storage`, `internal/auth`, `internal/storage/migrate`; #617 touches `internal/mcp`. Disjoint.
- 11 commits, ordered: C1 (behavior-noop refactor: registry + storage/keys + caller sites + guards/lint) then C2 (behavior change: auth relocate + refuse-newer + v3 + recovery + register + lockout test).

Closes #611.

### DIFF
--- a/cmd/muninn/server.go
+++ b/cmd/muninn/server.go
@@ -953,6 +953,7 @@ func runServer() {
 	uiAddr := flag.String("ui-addr", uiAddrDefault, "Web UI HTTP listen address")
 	mcpToken := flag.String("mcp-token", "", "Bearer token override for MCP auth (leave empty to read from MUNINN_MCP_TOKEN env var or ~/.muninn/mcp.token)")
 	dev := flag.Bool("dev", false, "serve web assets from ./web directory (development mode)")
+	forceMigrationRerun := flag.Bool("force-migration-rerun", false, "Reset the stored migration version to 0 and exit without starting the server. The next normal start re-applies every registered migration (all are idempotent). Operator recovery path for a wedged/partial migration (#611). Always back up the DB before a migration-bearing upgrade; use this flag to recover from a wedged/partial migration.")
 	backupInterval := flag.String("backup-interval", "", "Automated backup interval (e.g. 6h, 30m); empty = disabled")
 	backupDir := flag.String("backup-dir", "", "Directory to write automated backups into")
 	backupRetain := flag.Int("backup-retain", 5, "Number of automated backups to keep")
@@ -1152,6 +1153,27 @@ func runServer() {
 	// NOTE: db.Close() is NOT deferred here because store.Close() (called
 	// during the ordered shutdown sequence) internally closes the Pebble DB
 	// after flushing its own background workers.
+
+	// --force-migration-rerun: reset the stored migration version to 0 and
+	// exit. The next normal start re-applies every registered migration.
+	// This is the operator recovery path for a wedged/partial migration
+	// (#611, Task 7b / RT5): if a version was stamped but the operator needs
+	// to force a re-run (e.g. recover from a partial state), this resets
+	// the marker so the existing Runner re-applies the migrations on the
+	// next Open. It does NOT run migrations itself — that keeps the path
+	// simple and reuses the hardened Runner's fail-loud semantics.
+	if *forceMigrationRerun {
+		if err := migrate.ForceRerunMigrations(db); err != nil {
+			slog.Error("force-migration-rerun failed", "err", err)
+			_ = db.Close()
+			os.Exit(1)
+		}
+		slog.Info("migration version reset to 0; all registered migrations will re-run on next start",
+			"data_dir", *dataDir)
+		fmt.Fprintln(os.Stderr, "migration version reset to 0; re-run on next start. Re-run `muninn start` (without this flag) to apply.")
+		_ = db.Close()
+		os.Exit(0)
+	}
 
 	if err := replication.CheckAndSetSchemaVersion(db); err != nil {
 		slog.Error("schema version check", "err", err)
--- a/cmd/muninn/server.go
+++ b/cmd/muninn/server.go
@@ -1182,16 +1182,7 @@ func runServer() {
 
 	// Run versioned schema migrations before the storage layer is built.
 	migRunner := migrate.NewRunner(db)
-	migRunner.Register(migrate.Migration{
-		Version:     1,
-		Description: "backfill embed_dim in ERF records for existing embeddings",
-		Up:          migrate.BackfillEmbedDim,
-	})
-	migRunner.Register(migrate.Migration{
-		Version:     2,
-		Description: "backfill relationship entity index (0x26) for GetEntityAggregate optimisation",
-		Up:          migrate.BackfillRelEntityIndex,
-	})
+	migrate.RegisterMigrations(migRunner)
 	if applied, err := migRunner.Run(); err != nil {
 		slog.Error("migration failed", "err", err)
 		db.Close()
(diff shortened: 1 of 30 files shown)


scrypster (approved):
This is the fix #611 actually needed, a migration instead of an assertion. The relocation lands at 0x42 and up, clear of storage's 0x01 to 0x2A and the capability 0x40/0x41, and you swapped the old storageMaxPrefix guard for a real cross-package disjointness test that tightens itself as prefixes get added.

I checked the part that matters most, and then I went at it harder than usual because it's a real on-disk migration. The on-disk keys are physically re-keyed, not just the constants swapped (which would have stranded every existing deployment's admin users, API keys, and vault configs). I RED-checked the lockout regression, a pre-auth DB with digest flags but no admin still bootstraps root the way it should. And I tried to construct the data-loss case: a storage record that could get mis-relocated into the auth range, or an auth record (long username, long vault name) that could get skipped. The fixed-length storage keys close every one of those. VaultWeights is always 9 bytes and the APIKey vault index is always at least 11, so there's no length a storage key could hit that lands it in the auth bucket. Corrupted auth values fail loud instead of orphaning. Idempotent per-record, crash-safe, and it runs before Bootstrap so an existing admin gets relocated before AdminExists ever scans. Race suite green across auth, storage, and prefix.

Approving. One thing worth its own follow-up issue so it doesn't get lost: the 0x19 idempotency/replication overlap is the same class of bug, and you've honestly scoped it out of this PR. Thanks for driving this all the way to a real migration.

PR #621, opened 2026-07-18
fix(hnsw): late inserts born with zero in-edges stay permanently unreachable — protect the fresh back-edge in prune, repair orphans on load

### LINKED ISSUE #620: HNSW: incrementally-inserted nodes accumulate zero in-edges and stay unreachable (recall silently misses new memories)
### Version
v0.8.0 (built from the tag; single-node, cluster `enabled: false`)

### Summary
Memories written after startup are not returned by `recall` in **any** mode (semantic / balanced / deep). Only `find_by_entity` (the entity reverse-index) surfaces them. An FTS reindex and a plain restart do **not** fix it.

### Root cause (as far as I traced)
On restart the graph loads with unreachable nodes:

```
hnsw: loaded graph from pebble nodes=4269 vectors=4269 reachable_from_entry=4166 rebuilt=false
```

~100 nodes are unreachable from the entry point. The disconnect-rebuild at `internal/index/hnsw/hnsw.go:832`:

```go
if vectorCount > 1 && reachable*2 < vectorCount {
    // rebuild from vectors
}
```

only fires below **50%** reachability, so a steady trickle of unreachable nodes (added by the retroactive embed processor) never crosses the threshold and is never repaired. The code already documents the failure mode at `hnsw.go:493`:

> // nodes then accumulate zero in-edges and become unreachable

So incrementally-inserted nodes appear to get forward links but no reachable back-edges; greedy search from the entry point can't reach them. FTS masks it and `status` stays green.

### Repro
1. Configure an embedder (OpenAI `text-embedding-3-small`) so embeddings are produced by the retroactive processor.
2. After startup, write a memory containing a unique token.
3. Wait for `retroactive processor: complete plugin=embed-openai`.
4. `recall` by the unique token (any mode) → 0 results. `find_by_entity` on an extracted entity → found.
5. Restart → `reachable_from_entry` < `nodes`, `rebuilt=false` → still 0 results.

Observed on a vault of ~4500 engrams; each restart left ~100 recently-added nodes unreachable.

### Impact
New memories are effectively invisible to `recall` until enough accumulate to cross the 50% threshold (or a `vault clone`, which runs the full `reindexVault` HNSW rebuild). The failure is silent — FTS and `status` stay green.

### Possible fixes
- Ensure incremental insert adds reachable bidirectional links (the real fix).
- And/or lower / parameterize the rebuild threshold (`reachable*2 < vectorCount`), or add a periodic reachability check + rebuild.
- Expose an in-place HNSW rebuild command — currently only `vault reindex-fts` is exposed; the HNSW rebuild only runs via `vault clone` (`reindexVault`) or the >50%-disconnected startup path.

Happy to provide more logs or test a fix.

isaac-ranger:
Fixes #620.

We can confirm this report from production, and add one fact that narrows it: **the defect predates v0.8.0.** Our daemon runs a develop build from 2026-07-01 (post-#471, post-#545) and shows the same shape on every restart, on every vault large enough to have dense regions:

```
hnsw: loaded graph from pebble nodes=1527 vectors=1527 reachable_from_entry=1515 rebuilt=false   (jul 15)
hnsw: loaded graph from pebble nodes=1540 vectors=1540 reachable_from_entry=1527 rebuilt=false   (jul 16)
hnsw: loaded graph from pebble nodes=1549 vectors=1549 reachable_from_entry=1536 rebuilt=false   (jul 17)
```

Constant gap, never repaired, `rebuilt=false` forever — three other vaults on the same daemon show 3–4 orphans each. So #471's four integrity fixes closed the truncation-shaped loss but left this residual, and it bites well below the reporter's ~100-node scale.

## Root cause, confirmed at insert time

`pruneNeighbors` keeps the `maxConn` **nearest** edges. When a new node's nearest neighbors all sit in a dense cluster, every neighbor's layer-0 list is already full of mutually-closer members — the just-appended back-edge loses the distance contest at *every* neighbor, and the node is born with zero in-edges. That's fatal and **permanent**: search reaches nodes only through in-edges, and inserts discover neighbors *by graph search*, so no future insert can ever find the orphan to re-link it. No restart short of the >50% rebuild threshold ever repairs it.

The escape hatch explains the observed trickle rate: a node that draws an upper HNSW level (p ≈ 1/16 per level) gets back-edges on sparse upper layers that never overflow, so ~94% of inserts into a dense-enough region are at risk, the rest survive. In our repro below, 18/20 late inserts into a dense cluster were unfindable by their own embedding **the moment they were written** — the other 2 had drawn upper levels.

## The fix, two parts

**1. Insert — protect the fresh back-edge in the prune.** `pruneNeighbors` gains a `protect` id that is always retained (it takes one of the `keep` slots). Connectivity of a just-linked node beats the marginal distance optimality of the single edge it displaces. This guarantees every insert is born reachable.

**2. Load — repair small orphan sets instead of ignoring them.** `LoadFromPebble` now re-links each unreachable node (those below the existing 50% full-rebuild threshold) exactly the way a fresh insert would, then re-persists — including clearing the orphan's stale persisted layer keys first, so a re-drawn level can't resurrect ghost edges on the next load. Cost is one insert per orphan; zero when the graph is healthy. This also heals damage that already exists on disk — like the 13-node gap above — on the first restart after upgrade, and addresses the operational need behind "expose an in-place HNSW rebuild" for the restart case. The load log line gains `repaired=N` so operators can watch it happen.

The repair branch writes during load; precedent is the existing rebuild branch, which already re-persists healthy lists from the same code path.

## Honest limits

The protect only guards the edge being appended *right now*. Later inserts pruning the same neighbor lists can still evict an old node's last in-edge over long churn — the current design has always allowed that, and this PR does not add in-degree bookkeeping to prevent it continuously. What changes is that the damage is no longer permanent: any residual orphan heals at the next load. If you'd like a follow-up that makes pruning in-degree-aware (never evict an edge whose target would drop to zero layer-0 in-edges), happy to explore it — it needs per-node in-degree state and some care around deletes, so it felt like its own change.

## Tests

- `TestInsert_DenseCluster_LateInsertsStayReachable` — 40-node tight cluster + 20 late inserts in distinct directions; asserts the user-visible contract (a just-inserted vector is findable by its own embedding, immediately). Pre-fix: 18/20 fail with zero layer-0 in-edges.
- `TestLoadFromPebble_FewOrphans_RepairedOnLoad` — persists a healthy 60-node graph, strips all in-edges to 3 level-0 victims on disk (the exact #620 end state: forward-only nodes), asserts reload reaches 60/60, victims findable, and — via a third load — that the repair persisted.
- Full `./internal/...` green (47 packages, two consecutive runs), hnsw package under `-race -count=2`, and 8 consecutive non-race runs to shake out flakes (found and fixed one in my own new test: a missing `idx.Close()` racing the test DB close).

— Cairn (running MuninnDB in production for a five-agent household; context in #609)

### DIFF
--- a/internal/index/hnsw/hnsw.go
+++ b/internal/index/hnsw/hnsw.go
@@ -488,11 +488,15 @@ func (idx *Index) Insert(id [16]byte, vector []float32) {
 			maxConn := maxConnections(l)
 			if len(nbNode.layers[l]) > maxConn {
 				// Prune: keep the maxConn NEAREST neighbors by distance to this
-				// node's vector. Plain truncation ([:maxConn]) would always drop
-				// the just-appended edge once the layer is full — late-inserted
-				// nodes then accumulate zero in-edges and become unreachable,
-				// silently degrading the graph as the vault grows.
-				nbNode.layers[l] = idx.pruneNeighbors(nbNode.vec, nbNode.layers[l], maxConn)
+				// node's vector — but always retain the just-appended edge. The
+				// new node's only path into the graph is a back-edge from one of
+				// these neighbors; in a dense region every neighbor's list is
+				// already full of mutually-closer nodes, so distance-only
+				// pruning would drop the fresh edge at every neighbor and the
+				// node would be born with zero in-edges — permanently invisible
+				// to graph search, since future inserts can only discover
+				// neighbors by traversing the graph (#620).
+				nbNode.layers[l] = idx.pruneNeighbors(nbNode.vec, nbNode.layers[l], maxConn, id)
 			}
 			nbNode.mu.Unlock()
 			mutated[nb.id] = nbNode
@@ -519,9 +523,13 @@ func (idx *Index) Insert(id [16]byte, vector []float32) {
 
 // pruneNeighbors keeps the keep nearest neighbor ids to baseVec, measured by
 // cosine distance. Neighbors whose node or vector is missing sort last so they
-// are pruned first. Caller must hold the owning node's mutex; neighbor vectors
-// are immutable after insert, so reading them without their locks is safe.
-func (idx *Index) pruneNeighbors(baseVec []float32, ids [][16]byte, keep int) [][16]byte {
+// are pruned first. The protect id, when present in ids, is always retained
+// (it takes one of the keep slots): connectivity of a just-linked node beats
+// the marginal distance optimality of the edge it displaces. Pass the zero
+// value for no protection — engram ids are ULIDs, which are never zero.
+// Caller must hold the owning node's mutex; neighbor vectors are immutable
+// after insert, so reading them without their locks is safe.
+func (idx *Index) pruneNeighbors(baseVec []float32, ids [][16]byte, keep int, protect [16]byte) [][16]byte {
 	if len(ids) <= keep {
 		return ids
 	}
@@ -531,6 +539,10 @@ func (idx *Index) pruneNeighbors(baseVec []float32, ids [][16]byte, keep int) []
 	}
 	scored := make([]nd, 0, len(ids))
 	for _, nid := range ids {
+		if nid == protect {
+			scored = append(scored, nd{nid, -1.0}) // always sorts first
+			continue
+		}
 		n := idx.nodes[nid]
 		if n == nil || len(n.vec) == 0 || len(baseVec) == 0 {
 			scored = append(scored, nd{nid, 2.0}) // missing vector: prune first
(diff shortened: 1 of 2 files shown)


scrypster (approved):
This is great, and it's the perfect follow-on to your own #471 work. That production trace (reachable_from_entry=1536 of 1549, rebuilt=false) is the kind of thing tests alone would never have surfaced. It names the exact residual your four integrity fixes left behind, and it explains why it was permanent: search only reaches nodes through in-edges, inserts find their neighbors by searching, so a node born with zero in-edges can never heal itself.

I didn't take it on faith. I reproduced both tests red on the pre-fix code (18 of 20 late inserts unfindable, 57 of 60 on the orphan reload), then green with the fix, including -race -count=5. The two-part shape is right. The prune protect guarantees a node is born reachable, and the load-repair is gated on !rebuilt so it fills the gap the majority-disconnect rebuild leaves open instead of duplicating it. Clearing the stale higher-layer keys before re-insert so you don't get ghost edges was a nice touch.

Your note about the honest limits is the right call. If churn ever makes an in-degree-aware prune worth it, that's a separate change. Approving as-is. Two tiny optional things if you're ever back in here (a redundant BFS on the healthy-load path, and the test hard-asserting exactly 3 victims), neither blocks anything. Thanks for staying on this one, and for measuring everything.

PR #625, opened 2026-07-20
feat(grpc,engine): AdjustConfidence RPC with contradiction signaling (#559)

### LINKED ISSUE #559: feat(grpc,engine): AdjustConfidence RPC with contradiction signaling
## Summary

MuninnDB's internal contradiction worker detects inconsistencies via graph traversal — it only catches contradictions between engrams that are already linked. There is no API for an external system to signal that two engrams contradict each other based on semantic content.

## Context

Raised via community integration work (go-rag bridge, https://github.com/madeinoz67/go-rag). When a RAG system retrieves two document chunks that contain opposing factual claims, it should be able to signal that to MuninnDB so Bayesian confidence is updated and a `CONTRADICTS` edge is created. This is additive to the internal contradiction worker — a different detection modality, not a replacement.

## Proposed change

```protobuf
message AdjustConfidenceRequest {
  string vault              = 1;
  string engram_id          = 2;
  float  delta              = 3;  // positive or negative, clamped to [0,1] after apply
  string reason             = 4;
  string contradicted_by_id = 5;  // optional: creates a CONTRADICTS edge between the two engrams
}

message AdjustConfidenceResponse {
  float new_confidence = 1;
}

service MuninnDB {
  // ... existing RPCs ...
  rpc AdjustConfidence(AdjustConfidenceRequest) returns (AdjustConfidenceResponse);
}
```

## Behavior

1. Read current confidence for `engram_id`
2. Apply delta, clamp to `[0.0, 1.0]`
3. Call `UpdateConfidence` (already exists at `storage/engram.go:765`)
4. If `contradicted_by_id` is set: write a bidirectional `CONTRADICTS`-typed association edge between the two engrams and submit a contradiction event to the async confidence worker — consistent with how the internal contradiction detection works

## Files

- `proto/muninn/v1/service.proto`
- `internal/transport/grpc/engine_adapter.go`
- `internal/transport/grpc/server.go`
- `internal/engine/engine.go` — new `AdjustConfidence` engine method

## Notes

- `UpdateConfidence` storage primitive already exists — this is a thin engine + transport wrapper
- The `contradicted_by_id` path should reuse the same association write + confidence worker submission path as the internal contradiction worker for consistency

madeinoz67:
# feat(grpc,engine): AdjustConfidence RPC with contradiction signaling

Closes #559. Branch `feat/adjust-confidence-559` off `develop` @ `9868c62`.

## Summary

Adds an `AdjustConfidence` gRPC RPC so an external system (e.g. the go-rag bridge) can explicitly adjust an engram's confidence and optionally signal a contradiction — a different *detection modality* from the internal contradiction worker (externally-asserted vs graph-detected), additive to it.

```protobuf
rpc AdjustConfidence(AdjustConfidenceRequest) returns (AdjustConfidenceResponse);

message AdjustConfidenceRequest {
  string vault              = 1;
  string engram_id          = 2;
  float  delta              = 3;   // signed; clamped to [0,1] after apply
  string reason             = 4;   // audit string (caller-supplied)
  string contradicted_by_id = 5;   // optional; mirrors OnFound
}
message AdjustConfidenceResponse { float new_confidence = 1; }
```

## What it does

- Applies a signed `delta` to an engram's confidence, clamped to `[0,1]`, returning the new absolute value.
- When `contradicted_by_id` is set, **mirrors the internal `OnFound`** (`engine.go`, the `ContradictWorker` callback): writes the `0x0A` contradiction marker via `FlagContradiction` (the same primitive the internal worker uses) **and** submits `EvidenceContradiction` to the `ConfidenceWorker` for **both** engrams (Bayesian lowering). This reproduces the internal worker's effect, externally-triggered.
- A bare delta (`contradicted_by_id` empty) does **not** touch the worker — it's a direct `UpdateConfidence` only. Submitting there would let `processBatch` clobber the caller's explicit value.

## Design notes

- **Atomic write:** confidence (`0x01`+`0x02`) and the contradiction marker (`0x0A`) commit in **one Pebble batch** (`PebbleStore.UpdateConfidenceWithContradiction`) — no partial-failure window between them.
- **Validation before any write:** NaN/±Inf delta, self-contradiction, unknown `engram_id`, and unknown `contradicted_by_id` all reject before any Pebble write or worker submit — a rejected call mutates nothing.
- **Mirrors `Engine.Link`'s `RelContradicts` path** for vault resolution, `GetMetadata` existence checks, and the `EvidenceContradiction` submit pattern.
- Additive only: no migration, no new keyspace prefix, no new credential, no change to the internal contradiction/confidence workers or existing RPCs (+783/-0 across 12 files).

## ⚠️ Recall-inertness (please read before merge)

Mirroring the internal `OnFound` reproduces the internal worker's effect — which is itself **recall-inert**:

- Recall **never reads the `0x0A` marker** — `GetContradictions` is consumed only by the operator listing API, not recall scoring.
- Confidence **is not in recall scoring** — only the caller-supplied `MinConfidence` cutoff filter.
- Recall's contradiction dampening (`activation/profiles.go`, `RelContradicts`) traverses the **`0x09` association graph** — a different keyspace that neither this RPC nor the internal worker writes.

So neither half of this RPC dampens retrieval ranking today. **If the intent is recall dampening at retrieval time** (e.g. the RAG case: don't surface contradictory chunks), that needs a `0x09` `WriteAssociation(rel_type=CONTRADICTS)` write (which `profiles.go` dampens) or a scoring change — a separate, upstream-scoped decision. I did **not** build that here because it diverges from "additive to the internal worker" and is a maintainer-level call. Flagging for your steer; happy to follow up.

## Other notes for the maintainer

- **Auth:** mirrors `Link`/`Forget` (write-mode `mk_`/`cap_` via the existing interceptor). If you'd prefer admin-only (since it writes the previously-internal `0x0A` keyspace), easy to gate.
- **Non-idempotent delta:** the `delta` is non-idempotent by design; a gRPC retry re-applies it. The marker half is idempotent. If the bridge retries, an optional `idempotency_key` on the proto would close the response-loss window — recommendation, not blocking.
- **Concurrency:** `GetConfidence → clamp → UpdateConfidence` is a non-atomic RMW (same shape as `processBatch`); lost updates under concurrent writers are accepted under the cooperative trust model. The unified batch above closes the *partial-failure* window; a per-engram mutex would be the upgrade if real contention appears.
- **No `0x0A` provenance field** — externally-written markers are indistinguishable from internal-worker ones at query time. A `source`/provenance field is a worthwhile follow-up if external-write volume grows.

## Testing

- Unit (`internal/engine`): clamp, NaN/±Inf, self-contradiction, NotFound (engram + target), rejection-mutates-nothing, bare-delta-submits-nothing, contradiction-submits-evidence-×2, + RED-sanity (NaN guard shown load-bearing). 8/8.
- Storage (`internal/storage`): atomic write — confidence + marker commit together; marker absent when `hasContra=false`.
- gRPC (`internal/transport/grpc`): three-layer coverage (Server RPC, adapter ULID parsing, error→codes mapping); auth-interceptor matrix extended.
- Gates: `go build ./...`, `go vet ./...`, `go test -race ./...` all green. (CI doesn't run golangci-lint on Go; the new code mirrors the existing `defer batch.Close()`/`batch.Set` convention.)

## Out of scope (deliberate)

REST/MCP surfaces, recall dampening, firing the push trigger, hard concurrency atomicity (mutex/CAS), capability provenance — all flagged above as follow-ups.

### DIFF
--- a/internal/engine/engine.go
+++ b/internal/engine/engine.go
@@ -2531,6 +2531,85 @@ func (e *Engine) Link(ctx context.Context, req *mbp.LinkRequest) (*mbp.LinkRespo
 	return &mbp.LinkResponse{OK: true}, nil
 }
 
+// AdjustConfidence applies a signed delta to engram id's confidence (clamped to
+// [0,1]) and, when hasContra, mirrors the internal OnFound: persists the 0x0A
+// contradiction marker for (id, other) AND submits EvidenceContradiction to the
+// ConfidenceWorker for both engrams. Returns the new absolute confidence.
+//
+// All validation precedes any write. A bare delta (hasContra=false) does NOT
+// submit to the ConfidenceWorker (§D3 carve-out — processBatch would otherwise
+// clobber the explicit value).
+//
+// Mirrors Engine.Link's vault resolution + GetMetadata existence pattern +
+// cogWorkers() submit pattern. Source string is "external_contradiction" to
+// distinguish external bridge signal from the internal "contradiction_detected"
+// path emitted by Link/ContradictWorker.
+func (e *Engine) AdjustConfidence(ctx context.Context, vault string, id storage.ULID, delta float32, other storage.ULID, hasContra bool, reason string) (float32, error) {
+	wsPrefix := e.store.ResolveVaultPrefix(vault)
+
+	if math.IsNaN(float64(delta)) || math.IsInf(float64(delta), 0) {
+		return 0, fmt.Errorf("%w: delta is NaN or Inf", ErrInvalidArgument)
+	}
+	if hasContra && other == id {
+		return 0, ErrSelfContradiction
+	}
+
+	// Existence check — mirror Engine.Link's batched GetMetadata (nil meta = NotFound).
+	ids := []storage.ULID{id}
+	if hasContra {
+		ids = []storage.ULID{id, other}
+	}
+	metas, err := e.store.GetMetadata(ctx, wsPrefix, ids)
+	if err != nil {
+		return 0, fmt.Errorf("adjust confidence: read meta: %w", err)
+	}
+	for _, m := range metas {
+		if m == nil {
+			return 0, ErrEngramNotFound
+		}
+	}
+
+	// Current confidence (for the clamp + the audit prior).
+	current, err := e.store.GetConfidence(ctx, wsPrefix, id)
+	if err != nil {
+		return 0, ErrEngramNotFound
+	}
+
+	// Clamp to [0,1].
+	newConf := current + delta
+	if newConf < 0 {
+		newConf = 0
+	} else if newConf > 1 {
+		newConf = 1
+	}
+
+	// Atomic write (confidence + marker in one batch — no partial-failure window).
+	if err := e.store.UpdateConfidenceWithContradiction(ctx, wsPrefix, id, newConf, other, hasContra); err != nil {
+		return 0, fmt.Errorf("adjust confidence: write: %w", err)
+	}
+
+	// Mirror OnFound: submit EvidenceContradiction for both engrams (only when hasContra).
+	if hasContra {
+		if _, _, cw := e.cogWorkers(); cw != nil {
+			cw.Submit(cognitive.ConfidenceUpdate{
+				WS: wsPrefix, EngramID: [16]byte(id),
+				Evidence: cognitive.EvidenceContradiction, Source: "external_contradiction",
+			})
+			cw.Submit(cognitive.ConfidenceUpdate{
+				WS: wsPrefix, EngramID: [16]byte(other),
+				Evidence: cognitive.EvidenceContradiction, Source: "external_contradiction",
+			})
+		}
+	}
+
+	slog.Info("adjust_confidence",
+		"engram_id", id.String(), "delta", delta, "prior", current, "new", newConf,
+		"contradicted_by", other.String(), "has_contra", hasContra,
+		"reason", reason, "vault", vault)
+
+	return newConf, nil
+}
+
 // Forget implements mbp.EngineAPI.Forget.
 func (e *Engine) Forget(ctx context.Context, req *mbp.ForgetRequest) (*mbp.ForgetResponse, error) {
 	wsPrefix := e.store.ResolveVaultPrefix(req.Vault)
--- a/internal/engine/engine.go
+++ b/internal/engine/engine.go
@@ -2569,22 +2569,15 @@ func (e *Engine) AdjustConfidence(ctx context.Context, vault string, id storage.
 		}
 	}
 
-	// Current confidence (for the clamp + the audit prior).
-	current, err := e.store.GetConfidence(ctx, wsPrefix, id)
+	// Atomic delta write: the storage method holds the per-engram stripe lock
+	// across read+add+clamp+commit, closing the lost-update race (#559). The
+	// prior confidence read and the [0,1] clamp have moved INTO the locked
+	// storage method (UpdateConfidenceWithContradiction) — an external unlocked
+	// read here would re-open the race, so we do not call GetConfidence. The
+	// returned (prior, newConf) come from inside the locked section so the
+	// audit log below reports the same values that were committed.
+	prior, newConf, err := e.store.UpdateConfidenceWithContradiction(ctx, wsPrefix, id, delta, other, hasContra)
 	if err != nil {
-		return 0, ErrEngramNotFound
-	}
-
-	// Clamp to [0,1].
-	newConf := current + delta
-	if newConf < 0 {
-		newConf = 0
-	} else if newConf > 1 {
-		newConf = 1
-	}
-
-	// Atomic write (confidence + marker in one batch — no partial-failure window).
-	if err := e.store.UpdateConfidenceWithContradiction(ctx, wsPrefix, id, newConf, other, hasContra); err != nil {
 		return 0, fmt.Errorf("adjust confidence: write: %w", err)
 	}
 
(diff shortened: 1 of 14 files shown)


scrypster (changes requested):
Hey Stephen, this one's really close and I want it in, but the second pass turned up something real that needs a fix first.

The AdjustConfidence write doesn't take the per-engram CAS lock across its read-modify-write. I know UpdateConfidence has the same shape today, so this isn't something you conjured from nothing, but this PR adds a new external, higher-frequency trigger for that path, and we actually reproduced two failures with it. Fifty concurrent +0.01 adjusts from 0.0 landed at 0.02 to 0.06 instead of 0.50, so 44 to 48 of the 50 writes vanished. And running AdjustConfidence next to DeleteEngram on the same id, a few trials out of 200 left the deleted engram readable again, which is the exact #594 resurrection class where the unlocked Set of the engram key lands after the delete batch commits.

The fix is the one #594 used. Hold casLocks.For(id) across the read and the batch commit in UpdateConfidenceWithContradiction, mirroring DeleteEngram. And please add a concurrency regression test, the -race suite passes today only because nothing exercises the concurrent path.

Everything else is genuinely good and I checked it. NaN and Inf are rejected and the result clamps to [0,1], the contradiction signal stays cleanly additive to the internal worker with that external_contradiction source, the proto is additive and back-compatible, and the gRPC path enforces write-mode and vault pinning. So it's just this one thing.

One more and it's on us, not you. UpdateConfidence has the same unlocked-write exposure, so I'll open a separate issue to bring the internal confidence path under the same lock. Your PR just made it visible. Fix the locking and add the concurrency test and I'll pull it in. Strong continuation of your run here, thanks for taking #559.

madeinoz67:
@scrypster — resurrection fixed in 725b4c9: `casLocks.For(id)` held across the `GetEngram`→`Commit` in `UpdateConfidenceWithContradiction`, mirroring `DeleteEngram`. Regression test `TestUpdateConfidenceWithContradiction_NoResurrectionUnderConcurrentDelete` (200 trials × `-count=10`) goes RED without the fix (resurrects at iter 79/97) and GREEN with it.

**One thing surfaced while writing the test.** `DeleteEngram` invalidates its engram-cache entry *after* `mu.Unlock`, so a racing `UpdateConfidence` acquiring the stripe lock in that window reads a stale cached entry and writes it back — resurrecting the record even *with* the lock. Fixed additively by dropping `ps.cache.Delete` *under* the lock in `UpdateConfidenceWithContradiction`, before the `GetEngram` read. `DeleteEngram`'s own cache-invalidation ordering is a latent bug — its existing CAS-vs-delete test doesn't exercise it because `CompareAndSet` reads via `GetMetadata`, not `GetEngram`. Flagging for the separate internal-path issue you mentioned.

**The lost-update case (50× concurrent `+0.01`) is not fixed by the storage lock, and I don't want to claim it is.** `Engine.AdjustConfidence` reads confidence via `GetConfidence` *outside* the locked storage method and passes an absolute target, so no storage-internal lock can span that read. Empirically the lock alone concentrates the loss — serialized writers each writing the same stale absolute → 0.02 vs 0.04–0.07 without. Two ways to actually close it:

- **(a)** a storage-level `AdjustConfidence(delta)` that does read+add+clamp+write atomically under the stripe lock — keeps the lock in the storage layer, matching `DeleteEngram`.
- **(b)** hold the stripe lock at the engine layer across `GetConfidence`+write.

I lean **(a)**. Want me to implement it as the next commit, or do you prefer **(b)**?

madeinoz67:
@scrypster — lost-update closed in 4f158ea via approach (a): `UpdateConfidenceWithContradiction` now takes `delta` and does read+add+clamp+write atomically under the same `casLocks.For(id[:])` the resurrection fix (`725b4c9`) already holds — same shape as `CompareAndSet` from #594. `Engine.AdjustConfidence` no longer reads `GetConfidence` externally; it forwards `delta` and uses the `(prior, newConf)` the storage method returns from inside the locked section (prior feeds the audit log).

Proof — `TestUpdateConfidenceWithContradiction_NoLostUpdateUnderConcurrentDeltas`: 50 concurrent `+0.01` lands at `0.4999998` (float32 accumulation noise, 2e-7), stable across `-race -count=10`. Pre-fix was ~0.02 (last absolute-writer wins). The resurrection regression test (`NoResurrectionUnderConcurrentDelete`) still passes — the refactor is strictly additive inside the existing critical section; the lock shape from `725b4c9` is preserved verbatim.

One honest footnote: a direct *engine-layer* concurrency test (50 goroutines on `Engine.AdjustConfidence`) surfaces a separate, pre-existing race that's orthogonal to the confidence RMW — the engine's unlocked `GetMetadata` existence-check read races the storage method's post-commit `metaCache.Remove`/`cache.Set`. The stripe lock correctly serializes the confidence read-modify-write; only the auxiliary cache mutation and the existence-check read race. The storage-level test above is the canonical lost-update proof, so I landed that. Flagging the cache race as a follow-up — it's the same unlocked-read-vs-cache-invalidation shape as the `DeleteEngram` cache ordering I noted earlier, so it likely belongs in the internal-path issue you're opening.

scrypster:
Thanks for the re-work, @madeinoz67. Your three lock commits (725b4c9, 4f158ea, b9d68f4) fix the direct-path lost-update and the resurrection cleanly. I confirmed it under a 200-goroutine stress plus RED-sanity: the delta primitive serializes same-id callers and reads committed state, exactly as #559 needs.

One more path was still exposed. `AdjustConfidence(hasContra=true)` routes through the ConfidenceWorker, which persists via the absolute setter `PebbleStore.UpdateConfidence`. That one stayed an unlocked read-modify-write, so it resurrects a concurrently deleted engram about 1 in 500 under `-race`. Same class as #594, filed separately as #626.

I locked it to mirror your `UpdateConfidenceWithContradiction` exactly: `casLocks.For(id)`, cache drop before the authoritative read, `cache.Set` held under the lock, `replicateBatch` after unlock. Added the absolute-path resurrection test as the analogue of your delta-path one. build/vet/-race green, RED-sanity proven.

I also rebased your branch cleanly over develop (zero conflicts) so the merge stays linear. Your commit shas changed as a result of the rebase, so a `git fetch` + reset if you have the branch checked out locally.

Merging closes #559 and #626.

scrypster (approved):
Approving. All required checks and the vuln scan are green on the rebased head. The three lock commits fix the direct-path lost-update and resurrection (confirmed under 200-goroutine stress + RED-sanity), and the maintainer fixup locks the absolute UpdateConfidence path to close #626, with the resurrection test proving it under -race. Closes #559 and #626.

Issue #642, opened 2026-07-21
Enrichment circuit breaker counts context cancellation as provider failure


dpearson2699:
## Summary

`EnrichService.Enrich` executes the enrichment pipeline inside `breaker.Do`. Any non-nil pipeline error calls `RecordFailure`, including `context.Canceled` and `context.DeadlineExceeded` caused by the caller rather than by the provider.

This predates the fix for #640 and is intentionally out of scope for that narrowly focused PR.

## Current behavior

Repeated canceled or deadline-expired enrichment calls can increment `consecutiveFails` and eventually open the provider circuit. Five caller cancellations at the default threshold can therefore make healthy-provider calls fail with `circuit.ErrOpen` until the reset window elapses.

Relevant source:

- `internal/plugin/enrich/enrich.go`, `EnrichService.Enrich`
- `internal/engine/circuit/circuit.go`, `Breaker.Do`

## Expected behavior

Caller cancellation and deadline expiration should propagate through `errors.Is` without contributing to provider-health failure counts. Genuine provider, transport, parsing, and configuration failures should retain their existing breaker policy.

## Regression coverage

- Repeated `context.Canceled` results do not open the enrichment circuit.
- Repeated `context.DeadlineExceeded` results do not open the enrichment circuit.
- The cancellation/deadline cause remains discoverable through wrapping.
- Genuine provider failures still open the circuit at the configured threshold.

Issue #641, opened 2026-07-21
Enrichment parse-error logs can expose raw model output


dpearson2699:
## Summary

Enrichment response parsers include up to 160 bytes of raw model output in parse errors via `truncateForError`. Those errors are subsequently logged by the enrichment pipeline and may also appear in verbose LLM logs.

Because model output is derived from memory content, this can disclose memory text or other sensitive material in logs. This behavior predates the fix for #640 and is intentionally out of scope for that narrowly focused PR.

## Current behavior

Functions such as `ParseEntityResponse`, `ParseRelationshipResponse`, `ParseClassificationResponse`, and `ParseSummarizeResponse` construct errors containing a truncated response payload. `EnrichmentPipeline.Run`, `recordParseError`, and `recordComplete` can log those error strings.

Relevant source:

- `internal/plugin/enrich/parse.go`, `truncateForError` and parse errors
- `internal/plugin/enrich/pipeline.go`, stage warnings and verbose error logging

## Expected behavior

Parse failures should remain observable by stage and error category without logging raw model output, memory content, prompts, or credentials. If payload diagnostics are needed, they should use an explicit privacy-safe mechanism rather than default logs.

## Regression coverage

- Parse-error logs contain the stage and safe error classification.
- Raw model output is absent from normal and verbose logs.
- Successful parsing behavior and permanent malformed-output retry protection remain unchanged.

johanneshauer:
Is anyone already on this? If not, I'm happy to send a PR — you clearly left it out of #643's scope deliberately, so I don't want to cut across a plan you already have.

Shape I'd propose, matching the expected behavior above: parse errors keep their diagnostic value but stop carrying payload bytes. Concretely, `truncateForError` callers would produce errors with stage + a stable category (unmarshal failure / missing field / wrong type / empty response) plus response byte count — never response text — so `EnrichmentPipeline.Run`, `recordParseError`, and `recordComplete` can log them freely. If payload diagnostics are wanted at all, they'd live behind an explicit opt-in debug dump mechanism (off by default, clearly labeled), not in normal or verbose logs.

Regression coverage per your list: stage + category present in parse-error logs, raw model output absent from normal and verbose logs, and the malformed-output/retryability behavior from #640/#643 untouched.

dpearson2699:
> Is anyone already on this? If not, I'm happy to send a PR — you clearly left it out of #643's scope deliberately, so I don't want to cut across a plan you already have.
> 
[...]

Issue #390, opened 2026-04-13
Enrichment processor regression: ghost queue entries + circuit breaker deadlock (v0.4.12-alpha)


lshv:
## Summary

The retroactive enrichment processor retains references to soft-deleted/forgotten engrams in its queue, causing infinite retry loops that trip the circuit breaker and block enrichment for all memories — including new ones. This appears to be a regression of #197, #226, and #133.

## Environment

- Server: v0.4.12-alpha (Linux/Docker)
- Enrich provider: Google Gemini 2.5 Flash
- Embed provider: Voyage AI (voyage-3)
- Vault: ~670 engrams across 6 vaults

## Reproduction

1. Store memories, then soft-delete some via `muninn_forget`
2. Enable LLM enrichment (Google provider)
3. The retroactive processor picks up deleted engram IDs, sends their (empty/corrupt) content to the LLM
4. The LLM returns truncated or malformed JSON
5. After ~6 consecutive failures, the circuit breaker opens
6. **All** subsequent enrichment attempts — including for valid new memories — are rejected with `circuit breaker open`
7. On the next processor pass, it starts from the same deleted engrams again → deadlock

## Evidence

93 engram IDs in the enrichment queue were checked against both `work` and `default` vaults — **every single one returned 404 (engram not found)**. The queue is entirely ghost entries.

```
retroactive processor: failed to process engram plugin=enrich-google engram_id=01KN3NEJZCGZSACSWNF0ASK3HA
  error="enrich: all pipeline stages failed ... invalid entity response JSON"

retroactive processor: failed to process engram plugin=enrich-google engram_id=01KP3F540BM7D7409Q6WXXSQCZ
  error="circuit breaker open"
```

The second entry is a **new, valid memory** being blocked by the breaker that was tripped by the ghost entry above it.

### 1. Ghost entries in enrichment queue (regression of #197)
`muninn_forget` and content-hash deduplication (v0.4.11) remove the engram but do not remove it from the enrichment candidate set. The processor keeps trying to enrich IDs that no longer exist.

### 2. No skip-after-N-failures (regression of #226, #133)
The processor retries the same failed engram indefinitely with no backoff or skip logic. In our case, one engram was retried 25,000+ times in a single session.

### 3. Circuit breaker deadlock (new)
The circuit breaker correctly opens after consecutive failures, but:
- It blocks **all** enrichment, not just the failing engram
- When the breaker resets, the processor retries the **same ghost engram first**, immediately re-tripping the breaker
[...]

PR #643, opened 2026-07-21
fix(enrich): preserve transient provider failures

### LINKED ISSUE #640: HTTP 429 is treated as a permanent retroactive enrichment failure
## Summary

The v0.9.0 retroactive enrichment path appears to classify an HTTP `429 Too Many Requests` response as a permanent per-engram LLM failure. It then sets `DigestEnrichFailed`, preventing that engram from being retried automatically after the provider recovers.

Rate limiting is normally transient. A temporary provider throttle should delay work, not permanently exclude affected memories from enrichment.

## Affected version

- v0.9.0
- The relevant enrichment/retroactive processor implementation is unchanged from v0.8.0

## Steps to reproduce

1. Configure an enrichment provider against an HTTP test server.
2. Have the server return `429 Too Many Requests` for the first enrichment request, optionally with `Retry-After: 1`.
3. Return a valid enrichment response on later requests.
4. Submit an engram without enrichment digest flags and let the `RetroactiveProcessor` process it.
5. Inspect its digest flags and retry behavior after the provider is healthy.

## Actual behavior inferred from the v0.9.0 path

1. The provider returns a normal error for non-2xx responses, including 429.
2. `EnrichService.Enrich` records that error as a circuit-breaker failure.
3. `RetroactiveProcessor.processEnrichEngram` treats every provider/pipeline error other than `circuit.ErrOpen`, context cancellation, or deadline expiration as `errLLMFailed`.
4. `processBatch` responds to `errLLMFailed` by setting `DigestEnrichFailed`.
5. Future scans skip the engram permanently unless an operator manually clears flags/reprocesses the vault.

Several consecutive throttled requests can therefore both mark individual engrams permanently failed and open the global enrichment circuit.

Relevant source:

- `internal/plugin/retroactive.go`, `processEnrichEngram` and the `errLLMFailed` branch in `processBatch`
- `internal/plugin/enrich/providers.go` and provider-specific HTTP handling
- `internal/plugin/enrich/enrich.go`, circuit-breaker integration

## Expected behavior

- HTTP 429 is represented as a typed transient/rate-limit error.
- `Retry-After` is honored when present.
- Throttled engrams remain pending and are retried with bounded exponential backoff and jitter.
- Rate limiting does not set `DigestEnrichFailed`.
- The processor does not immediately move to many additional engrams in a way that expands the throttle event.
- Truly permanent per-engram failures, such as unrecoverable malformed model output after the applicable retry policy, can still be marked failed.

It may also be appropriate to classify network errors, HTTP 408, and retryable 5xx responses as transient, while keeping authentication/configuration failures visible as plugin-level unhealthy states.

## Suggested implementation

- Introduce typed provider errors carrying status code, retryability, and optional retry-after duration.
- Preserve that classification through the pipeline and `EnrichService`.
- Teach `processEnrichEngram`/`processBatch` to leave transiently failed engrams pending.
- Apply provider-wide backpressure before selecting the next engram.
- Expose throttled/retry counts and next-attempt time in observability rather than counting these as permanently failed memories.

## Regression coverage

- `429` followed by `200` eventually enriches the original engram without manual flag clearing.
- `DigestEnrichFailed` is not set after a 429.
- `Retry-After` delays the next provider request by at least the requested interval.
- Repeated 429s do not cause a tight loop or unbounded request burst.
- Permanent malformed-output behavior remains protected against infinite retries.

dpearson2699:
## Summary

- classify enrichment provider failures with typed status, retryability, and optional `Retry-After` metadata
- preserve provider errors through the pipeline, service, and retroactive processor
- apply provider-wide `Retry-After` or bounded exponential backoff with jitter
- leave systemic provider failures pending and stop the current batch instead of setting `DigestEnrichFailed`
- retain the permanent-failure guard for successful-but-malformed LLM output

## Root cause

Provider clients returned plain formatted errors for every non-200 response. The pipeline then flattened stage errors into strings, and `processEnrichEngram` classified every non-circuit/non-context error as `errLLMFailed`. `processBatch` consequently stamped `DigestEnrichFailed` after a 429, permanently excluding that engram from automatic enrichment.

The pipeline could also continue into later stages after a provider throttle, amplifying the rate-limit event across both stages and subsequent engrams.

## Error policy

- `429`, `408`, `500`, `502`, `503`, `504`, and non-context transport failures are retryable systemic provider failures.
- Valid `Retry-After` delta-seconds and HTTP dates are preserved; overflowing delta-seconds are rejected safely.
- Retryable failures still count toward the general circuit breaker. A throttled provider is unavailable, and the breaker remains a useful second safety net. Explicit provider-wide backpressure prevents burst amplification, while typed classification prevents the breaker from poisoning individual engrams.
- Nonretryable provider failures such as authentication/configuration responses are also systemic: they remain pending, stop the pass, and stay observable through the breaker rather than becoming per-engram failures.
- Successful HTTP responses with malformed/unusable model output remain permanent per-engram failures and continue to set `DigestEnrichFailed`, preserving the #390 infinite-retry protection.

Generation ordering prevents an older concurrent success from clearing a newer throttle window. Tests use injected time/wait seams and do not sleep.

## Verification

- RED-first regressions demonstrated the original permanent flagging, missing backpressure, concurrent throttle erasure, and duration overflow.
- `go test -count=1 ./internal/plugin/enrich ./internal/plugin ./internal/engine/circuit`
- `go test -race -count=1 ./internal/plugin/enrich ./internal/plugin ./internal/engine/circuit`
- `go build -tags localassets ./...`
- `go vet ./...`
- `gofmt -l .`
- `go test -count=1 ./...`

## Compatibility and operations

- No wire format, storage schema, digest-bit, configuration, or public protocol changes.
- Provider HTTP error strings no longer include response bodies, avoiding accidental content or credential exposure; status and typed metadata remain available.
- During provider outages, pending enrichment work pauses globally and resumes after recovery. Long-lived authentication/configuration failures surface through circuit health and do not silently consume or permanently exclude memories.

Follow-up issues for pre-existing, out-of-scope findings: #641 and #642.

Fixes #640

### DIFF
--- a/internal/plugin/enrich/anthropic.go
+++ b/internal/plugin/enrich/anthropic.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"io"
 	"net/http"
 	"time"
 
@@ -108,13 +107,12 @@ func (p *AnthropicLLMProvider) Complete(ctx context.Context, system, user string
 
 	resp, err := p.client.Do(httpReq)
 	if err != nil {
-		return "", fmt.Errorf("request failed: %w", err)
+		return "", providerTransportError(p.Name(), err)
 	}
 	defer resp.Body.Close()
 
 	if resp.StatusCode != http.StatusOK {
-		bodyBytes, _ := io.ReadAll(resp.Body)
-		return "", fmt.Errorf("anthropic returned status %d: %s", resp.StatusCode, string(bodyBytes))
+		return "", providerHTTPError(p.Name(), resp)
 	}
 
 	var messagesResp anthropicMessagesResponse
--- a/internal/plugin/enrich/enrich.go
+++ b/internal/plugin/enrich/enrich.go
@@ -2,8 +2,10 @@ package enrich
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
+	"math/rand"
 	"sync"
 	"time"
 
@@ -34,8 +36,10 @@ type LLMProviderConfig struct {
 // defaultBreaker thresholds: 5 consecutive failures open the circuit;
 // it probes again after 30 s.
 const (
-	breakerMaxFails   = 5
-	breakerResetAfter = 30 * time.Second
+	breakerMaxFails    = 5
+	breakerResetAfter  = 30 * time.Second
+	providerBackoffMin = time.Second
+	providerBackoffMax = time.Minute
 )
 
 // EnrichService implements plugin.EnrichPlugin.
@@ -54,6 +58,17 @@ type EnrichService struct {
 	// so it is always non-nil. OnStateChange may be wired after construction
 	// (before concurrent use) to emit logs and update the plugin registry.
 	breaker *circuit.Breaker
+
+	// backpressure is provider-wide: a transient response from any engram
+	// delays every subsequent call through this service. The function seams
+	// keep time-based behavior deterministic in tests.
+	backpressureMu    sync.Mutex
+	retryNotBefore    time.Time
+	retryableFailures int
+	backpressureGen   uint64
+	nowFn             func() time.Time
+	waitFn            func(context.Context, time.Duration) error
+	jitterFn          func(time.Duration) time.Duration
 }
 
 // NewEnrichService creates an EnrichService from a provider URL.
@@ -84,6 +99,9 @@ func NewEnrichService(providerURL string) (*EnrichService, error) {
 		provCfg:  provCfg,
 		name:     name,
 		breaker:  circuit.New(breakerMaxFails, breakerResetAfter),
+		nowFn:    time.Now,
+		waitFn:   waitContext,
+		jitterFn: addBackoffJitter,
 	}
 
 	return es, nil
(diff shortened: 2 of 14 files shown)


scrypster:
Thanks @dpearson2699, this is a clean fix for #640. The 429-to-retryable-with-backpressure path is right and your RED test proves it. I ran it locally and confirmed transient failures now survive as retryable.

While reviewing I found one adjacent gap the fix surfaced rather than caused. Non-retryable content 4xx (400/413/422/404) took the same break-and-pend path as systemic failures, but those don't latch `DigestEnrichFailed`. Since `ScanWithoutFlag` restarts from the keyspace head every pass, one content-4xx engram that sorts first re-breaks every pass and starves every follower behind it forever. That's the #587 wedge re-opening.

Rather than send it back, I took it the rest of the way since you'd left maintainerCanModify on, much appreciated. I split the classification: content 4xx now wraps `errLLMFailed` so it latches and continues, and only systemic conditions (429/5xx/408/transport/401/403) break-and-pend. Added a regression test that models the flag-driven scan and proves the follower drains. build/vet/-race/gofmt green, RED-sanity confirmed.

Merging closes #640 and the #587 wedge. I left #641 and #642 open, you were right to defer those.

scrypster (approved):
Approving. Required checks and the vuln scan are green on the rebased head. The maintainer fixup (content-4xx latching) is covered by the regression test and RED-sanity proven, as noted above. Closes #640 and the #587 wedge.

PR #644, opened 2026-07-21
fix(plugin): fence retroactive scans during vault clear

### LINKED ISSUE #639: Retroactive processor continues processing deleted engrams after a concurrent vault clear
## Summary

In v0.9.0, clearing or rebuilding a vault while a `RetroactiveProcessor` pass is active can leave that pass processing engrams captured by its existing scan iterator after those engrams have been deleted. The server then emits repeated `engram not found`/persistence failures for removed IDs and may continue working through stale items until the pass ends or the server is restarted.

This is related to the ghost-entry class addressed by #390, but the trigger here is a concurrent destructive vault operation against an already-open scan, rather than soft-deleted records being counted as pending on a later scan.

## Affected version

- Observed on v0.8.0
- The relevant `internal/plugin` implementation is unchanged in v0.9.0

## Steps to reproduce

1. Configure any enrichment plugin backed by a controllable test server.
2. Store enough unenriched engrams to keep a retroactive enrichment pass active.
3. Block one enrichment request so the processor retains an open `ScanWithoutFlag` iterator.
4. Clear/delete the vault contents while that pass remains active.
5. Release the blocked request and observe subsequent processor activity and logs.

The same test can be performed for the embedding processor with a blocked micro-batch.

## Actual behavior

The active pass can continue iterating over engram objects from the pre-clear snapshot. It attempts inference and/or persistence for IDs that no longer exist, producing repeated failures. For an external provider, this can also spend requests on content that the operator has already deleted.

The likely race is in `internal/plugin/retroactive.go`: `processBatch` obtains one iterator and consumes it for the lifetime of the pass, while destructive vault operations do not appear to invalidate or cancel that pass.

## Expected behavior

Once a vault clear/delete operation completes:

- No new embedding or enrichment call should start for an engram deleted by that operation.
- An active retroactive pass should stop or safely skip stale iterator entries.
- Deleted IDs should not generate a long stream of errors or affect the circuit breaker.
- New records added after the clear should be processed normally without requiring a restart.

## Suggested implementation options

- Associate processor scans with a vault generation/epoch and abort the pass when the generation changes.
- Cancel and restart the affected retroactive processors around a vault clear/rebuild operation.
- Revalidate an engram's existence immediately before invoking an external plugin and treat not-found as a benign skip.
- Ensure a clear operation wakes the processor after invalidating the old pass so newly written records are discovered promptly.

The existence check alone should be evaluated carefully: the record can still disappear between the check and the call, so cancellation/generation tracking may provide a stronger guarantee.

## Regression coverage

Add a deterministic concurrency test that:

1. Starts a retroactive pass with a plugin blocked on a channel.
2. Clears the underlying vault while the pass is blocked.
3. Releases the plugin.
4. Verifies that no additional stale engram is sent to the plugin.
5. Verifies that not-found entries are benign and do not increment systemic provider failures/open the circuit.
6. Writes one new engram and proves it is processed without restarting the server.

## Related

- #390 fixed soft-delete phantom counts and permanent retry loops, but does not appear to cover invalidating an iterator that was opened before a concurrent hard clear.

dpearson2699:
## Summary

- fence retroactive enrichment and embedding work with per-vault generations and admission guards
- make vault-clear admission context-cancellable and abort the whole clear before storage deletion when an admitted call does not drain
- keep enrichment admission release structurally scoped to a per-engram closure so every success, cancellation, permanent-failure, and provider-failure exit releases it
- preserve embedding micro-batching across vaults while filtering stale vault entries
- restore FTS admission before releasing/notifying processors, then discover post-clear records without restarting
- document that processor registration is correctness-critical, not observability-only

## Root cause

`ScanWithoutFlag` uses a fixed Pebble iterator snapshot. An active `RetroactiveProcessor.processBatch` therefore retained decoded pre-clear engrams after `Engine.ClearVault` deleted their storage records. Nothing coordinated the engine's per-vault destructive-operation mutex with plugin admission, so releasing one blocked provider call let the old iterator begin more external calls with deleted content. Persistence then failed with not-found errors and inflated processor error statistics.

A deterministic baseline RED run against the unfixed implementation reproduced the exact interleaving: after clear completed, the provider received all three pre-clear records and each persistence attempt logged not-found.

After #643 landed, its provider-failure stop paths also exposed a manual guard-release defect: returning or breaking from several enrichment branches left the vault admission held indefinitely. The new provider-failure regression failed before this follow-up with `BeginVaultClear` blocked after the pass had ended.

## Concurrency semantics

Each processor owns a guard per workspace prefix.

- A scan snapshots the current per-vault generation.
- Immediately before an enrichment call or embedding micro-batch, it enters the affected vault's admission guard and rechecks both the generation and clear state.
- Admission remains held through provider inference and persistence.
- Clear advances the generation, closes admission, and waits for admitted calls to drain.
- The wait observes the operation context. Cancellation reopens admission, advances the generation again, wakes the processor, and returns an error; `Engine.ClearVault` then releases any earlier processor boundaries and returns without deleting storage.
- A successful clear deletes storage and resets in-memory state while admission remains closed. Defer LIFO restores FTS writes before processor admission is reopened and processors are notified.

The two generation advances reject scans that predate the clear request and scans opened while clear holds the boundary. If plugin admission wins first, clear cannot complete until persistence finishes and then deletes that result. If clear wins, stale content cannot begin an external call. Synchronization remains scoped per vault.

## Branch decision and #640 helper

PR #643 for issue #640 is now merged into `develop`, and this branch was rebased onto current `develop`. Its `ProviderError` helpers apply to #639's systemic enrichment-error classification and are reused directly. They do not identify stale Pebble snapshots or establish a clear-completion boundary, so #639 adds no duplicate helper and no artificial stacked dependency.

## Review changes addressed

- rebased onto current `develop`, including #643
- replaced manual enrichment unlocks with a per-engram deferred release
- made `BeginVaultClear` context-aware and made `Engine.ClearVault` abort on boundary cancellation
- added a provider-failure lock-release regression
- added an `Engine.ClearVault` integration regression with a blocked provider, cancelled clear, successful retry, LIFO release, and post-clear rediscovery
- updated `SetRetroactiveProcessors` and server wiring comments to describe their correctness role

## Verification

- RED: provider failure ended the pass, then `BeginVaultClear` remained blocked on the unfixed code
- `go test ./internal/plugin -run 'TestRetroactiveProcessor_(ConcurrentVaultClear|PassOpenedDuringVaultClear|VaultClearDoesNotWait|ProviderFailureReleases|CancelledVaultClear)' -count=20`
- `go test ./internal/engine -run 'TestEngineClearVault_(MemoriesGone|BlockedRetroactiveProcessorHonorsCancellationAndRecovers)' -count=20`
- `go test -race ./internal/plugin ./internal/engine -count=1`
- `go test ./... -count=1`
- `go build ./...`
- `go vet ./...`
- `gofmt -l .`
- `git diff --check`

An independent repository code-review pass found no blocking correctness or scope issues.

Fixes #639


scrypster (changes requested):
This is good work and the fence design holds up. I went at the ordering specifically and could not break it. Snapshotting generations before the iterator opens is the safe direction, and the two-phase advance rejects both the pass that predates the clear and the one opened inside the window. Three of your four tests go red without the fence and reproduce #639 exactly, stale provider calls and all.

It cannot go in as it stands though, and the reason is not visible on your branch.

#643 landed on develop after you opened this. It restructured the control flow of the same `processEnrichEngram` block you are instrumenting, adding a `return true`, a `continue`, and a `break engramLoop`. Your guard is released by hand at six sites rather than by defer. The two merge textually clean and the merged tree passes CI, but a single enrich provider failure now exits without releasing `guard.mu.RLock()`. The reader count never drops, so the next `BeginVaultClear` for that vault blocks forever while holding the vault mutex, and `ClearVault` and `DeleteVault` for that vault never complete again without a restart. No concurrency needed to hit it, just one 500 from a provider.

What I need:

1. Rebase onto current develop and scope the read guard with defer inside a per-engram helper or closure, so no present or future `return`, `continue`, or `break` can escape it. Add a regression test that fails one enrich call, ends the pass, then asserts `BeginVaultClear` completes. Trap: a bare `defer` in the loop body defers to function exit, so it would hold the guard for the whole pass and block `ClearVault` that entire time. It has to be a nested func or a separate method. `flushMicroBatch` already gets this right because it is a closure.
2. Make the clear-boundary wait context-aware or bounded. `guard.mu.Lock()` ignores ctx, and the enrich path holds the read guard across a provider call whose HTTP timeout is 300 seconds. So a previously sub-second admin operation can now block for five minutes uncancellably while holding the vault mutex. If you make it abortable, abort the whole clear rather than proceeding with a call still admitted.
3. Add one test through `Engine.ClearVault` with a blocked processor. All four current tests call `BeginVaultClear` and `ClearVault` directly, so the defer LIFO ordering in `engine_vault.go` is untested. It is correct today and it is load-bearing, and one reordered defer breaks it silently.
4. `SetRetroactiveProcessors` and the `obsProcs` wire-up both say the registry is for observability. It is now required for correctness. Please update both comments so the next person adding a processor knows it has to be registered.

Two smaller things, not blocking. `vaultGuards` is never pruned, so it keeps an entry per vault ever cleared. And "unrelated vaults remain active" is a bit strong for the enrich path, since `ScanWithoutFlag` iterates all vaults and one stale engram breaks the whole pass rather than skipping that vault.

When it comes back I want a second independent pass on it before merge, since this is concurrency plus vault lifecycle and it has only had one set of eyes.

Good catch filing #639 and then fixing it.

scrypster (changes requested):
First, an apology: you pushed the requested changes on the 28th and we did not come back to it. That is on us.

All four landed, and they landed well. We checked out your first commit to confirm the guard-leak fix earns its keep, and the deadlock reproduced exactly as described, then went away on your head. The structurally-scoped release is the right shape, not just a defer moved around: expressing the three control-flow exits as returned actions means a future return or break cannot escape the release either.

We also proved the race is real rather than assuming it. With only the fence neutered, three engrams went to the enrichment provider after ClearVault had returned success, and a global entity record minted from that cleared content survived the clear permanently un-prunable, because 0x1F is deliberately excluded from clearVaultDataPrefixes and the mention decrement runs before the stale write lands. That second one is the precise reason your drain wait matters and not just the generation check, and it is worth a comment, because it is subtle and nothing currently says it out loud. Worth knowing too, since it changes how we would describe #639: vault-scoped key resurrection does not actually happen even unfenced, the ws-scoped writes fail not-found. The real damage is the egress, the global entity leak, and error stats reading zero while three engrams leaked.

One new thing, and it is the reason we run a second pass at all. The invalidation path calls rp.Notify on itself, so a rejected pass immediately wakes itself, and the notify branch has no backoff. We measured roughly 175,000 scan passes per second for the whole duration of a clear, per registered processor, and production registers two. ScanWithoutFlag is global, so unrelated vaults starve for that window too.

1. Delete the three rp.Notify calls on the invalidation path. The clear's own finish closure already wakes the processor at exactly the right moment, and so does the cancellation arm. We validated it: 52,372 passes in a 300ms window becomes 1, and both suites stay green under race.
2. Pin it, because nothing currently holds this in either direction. A counting wrapper over ScanWithoutFlag, modelled on your scanSignalingStore, asserts a bounded pass count while a boundary is held. Please do not fix it with a sleep or a backoff on that path, that trades a spin for a stall.
3. Rebase onto develop, 50 commits have landed since your merge-base. It auto-merges cleanly today but that should not be taken on faith at merge time.
4. One comment on BeginVaultClear noting the wait is bounded only by the caller's context, and that the enrich path can hold admission for the provider's full HTTP timeout.
5. Two stale comments: the "without stopping processing for unrelated vaults" claim (an invalidation now aborts the whole cross-vault pass), and the vaultGuards sizing note (an entry per vault ever scanned, not ever cleared, since lockCurrentVault creates them).

None of this is a no. The fence design survived the attack, including the ordering question: the two-generation protocol correctly rejects both a pass predating the clear and one opened inside the window, and taking the snapshot before opening the iterator is the right primitive. The spin is a feedback loop hiding behind a correct fence rather than a flaw in the fence. After these land we will run the second adversarial pass and get it in.

dpearson2699:
@scrypster you know you can land edits on the PR as well without waiting for me to make changes right? Just checkout the branch and push the changes you want

PR #645, opened 2026-07-21
fix(web): report authoritative cognitive worker state

### LINKED ISSUE #638: Admin dashboard misreports dormant cognitive workers as inactive
## Summary

In v0.9.0, the Admin dashboard's **Cognitive Engine Workers** card can report enabled workers as `Inactive`, presents **Temporal Scoring** as a background worker even though no corresponding worker statistic is returned, and says the workers are "Configured via server startup flags" even though the server starts the cognitive workers automatically.

This makes a healthy default installation look incorrectly configured.

## Affected version

- v0.9.0
- The same frontend behavior was present in v0.8.0

## Steps to reproduce

1. Start a fresh MuninnDB v0.9.0 server using normal defaults.
2. Open the Admin page before the cognitive workers have processed any qualifying work.
3. Inspect the **Cognitive Engine Workers** card.
4. Compare it with `GET /api/workers`.

## Actual behavior

- Hebbian Learning, Contradiction Detection, and Confidence Updates can appear as `Inactive` with zero processed even when their server-side state is legitimately dormant/idle and they are available to wake for work.
- Temporal Scoring always appears inactive with a dash because the frontend reads `ws.decay`, while the worker-status response does not expose a `decay` worker.
- The explanatory copy says the workers are "Configured via server startup flags," but there are no corresponding normal CLI startup flags and the server constructs/starts these workers automatically.

The v0.9.0 frontend determines activity using:

```js
active: stats && (stats.processed > 0 || stats.running)
```

However, a worker that is enabled but has processed zero items is not necessarily inactive, and the worker stats do not provide a matching `running` boolean.

Relevant source:

- `web/static/js/app.js`, `workerRows()`
- `web/templates/index.html`, Cognitive Engine Workers card
- `internal/cognitive/worker.go`, worker state definitions
- `cmd/muninn/server.go`, cognitive worker construction/startup

## Expected behavior

The dashboard should distinguish configuration and runtime state accurately:

- **Active/running**: currently processing
- **Idle**: enabled and waiting
- **Dormant**: enabled, backed off after inactivity, and able to wake
- **Disabled**: not configured/started

Temporal scoring should either:

1. Be removed from the background-worker table and described as query-time scoring, or
2. Be backed by an actual status field with accurate semantics.

The help text should describe how the workers are actually enabled/configured in v0.9.0.

## Suggested implementation

- Expose an explicit, stable status from `/api/workers` for each real worker, including an `enabled`/configuration signal if needed.
- Render the server-reported state rather than inferring activation from `processed > 0`.
- Remove the synthetic `ws.decay` row unless a real temporal worker exists.
- Replace the startup-flags copy with wording consistent with the default plasticity preset and actual server configuration.
- Consider showing a tooltip explaining that dormant workers wake when qualifying work arrives.

## Regression coverage

- A worker with state `Dormant` and `processed: 0` renders as **Dormant**, not **Inactive**.
- A configured idle worker with zero processed renders as enabled.
- The UI does not invent worker rows absent from the API contract.
- The explanatory copy matches the supported configuration surface.

dpearson2699:
## Summary

- add backward-compatible symbolic `status` and `enabled` fields to `GET /api/workers`
- render worker lifecycle state from the server in both Admin dashboard views
- distinguish active, idle, dormant, stopped, disabled, and unknown states without inferring from processed counts
- remove the synthetic Temporal Scoring worker row and describe ACT-R scoring as query-time plasticity configuration
- document the worker API contract and correct stale architecture/help text

## Root cause

The detailed Admin worker table inferred activity from `processed > 0` or a nonexistent `running` field, so healthy workers with no processed work appeared inactive. It also created a Temporal Scoring row from a nonexistent `decay` response field even though ACT-R temporal scoring runs during queries rather than in a background worker.

## Compatibility

Existing numeric `state` and telemetry fields are unchanged. The new `status` and `enabled` fields are additive. Disabled/unconfigured worker slots retain their legacy zero-valued counters and numeric state while now reporting `status: "disabled"` and `enabled: false`.

## Validation

- `go test -race ./internal/cognitive ./internal/engine ./internal/transport/rest ./web`
- `npx vitest run worker-status-utils.test.js time-utils.test.js plugin-config-utils.test.js` (42 tests)
- `go build ./...`
- `go vet ./...`
- `gofmt -l .`
- `node --check web/static/js/app.js`
- `node --check web/static/js/worker-status-utils.js`
- Vite build and OpenAPI lint

`go test ./...` was also attempted locally. Issue-relevant packages passed; unrelated tests that open loopback listeners were blocked by the execution sandbox with `bind: operation not permitted`.

Fixes #638


scrypster (changes requested):
Thank you for this, and for pulling the mapping out into its own module. That is what made the fix testable instead of something we take on faith, and it is why the frontend no longer guesses state from `processed` at all.

What I verified. Reverting each of the three production files in turn turns a different test red, so all three pieces are load-bearing rather than one fix plus scaffolding. I also ran develop's old `workerRows()` against a payload of three genuinely dormant workers and got "Inactive" three times, which is #638 exactly, and the new module maps the same payload to Dormant. The openapi update is there and Redocly is clean with no new warnings against base. Preset parity still passes after merging current develop in.

Two things before I merge.

1. `Normalize()` is applied at `rest/server.go:935` but not at `ui/server.go:437`, and both feed the same dashboard card through the same interface. It is harmless today because `Engine.WorkerStats()` already seeds the disabled stats, which is also why reverting `server.go` only reddens through the mock. A guard on one of two paths is the shape someone later assumes is universal. Either add it to the SSE path or drop it and lean on `WorkerStats()`. If you drop it, `TestHandleWorkerStats` will redden because it currently passes only on the zero-value mock being rescued, and the right fix there is to make the mock return `DisabledWorkerStats()` rather than to soften the assertion.
2. Add a clause to the `state` description in openapi saying zero is ambiguous and to check `enabled` first. Keeping legacy `state` is the right call, but zero now means both active and disabled, and a legacy client reading only that field sees a disabled worker as Active. That is the same bug class as #638 relocated.

Non-blocking. `workerStateBadge()` in `app.js` is dead now, worth deleting. And in the admin table `stopped` and `disabled` both bind to `badge-dormant`, so they look identical even though the module gives them distinct colors, which is not quite what #638 asked for.

One thing to be aware of rather than fix here: your new vitest file will not run in CI. Nothing runs `npm test` for `web/`, so `time-utils` and `plugin-config-utils` are in the same position. That is pre-existing and not yours to solve. The blocker was that vitest used to glob the Playwright specs, which got fixed on develop earlier today, so wiring it up is now cheap and I will take it as a follow-up.

Good to see you fixing issues you filed yourself. That plus #643 is a useful pattern.

dpearson2699:
Thanks, these were good catches. I pushed the requested changes and merged current `upstream/develop`.

I removed `Normalize()` entirely, so REST and SSE now rely on `Engine.WorkerStats()` as the single source of truth. The REST mock returns `DisabledWorkerStats()` explicitly. I also documented the ambiguous numeric zero in OpenAPI and the feature reference, removed the dead `workerStateBadge()`, and gave stopped and disabled separate badge classes and styles.

I red-checked the regressions before applying the production fixes. Dropping `Normalize()` before fixing the mock failed `TestHandleWorkerStats`, and the new OpenAPI and badge tests failed for their expected reasons.

After the fixes, these all pass:

- `go test -race ./internal/cognitive ./internal/engine ./internal/transport/rest ./web`
- 43 web Vitest tests
- `go build ./...`
- `go vet ./...`
- `gofmt -l .`
- Vite production build
- Redocly validation (the same 85 existing warnings, no errors)

The follow-up is in `0b7a427`.

scrypster (approved):
Thank you for the patience on this one, and I am sorry it sat. You pushed on the 28th and we did not come back to it, which is on us, not on you.

All four are in, and you took the better route on the first one. Deleting Normalize() entirely rather than adding it to the SSE path means there is one source of truth instead of two that happen to agree today. We checked: it exists nowhere in the tree now, and both consumers read Engine.WorkerStats directly. The mock got stronger rather than weaker too, which is the part people usually skip.

We reverted each piece in place to confirm the tests earn their keep, and all five went red: the mock, the engine seeding, the openapi description, the CSS badges, and the JS class mapping. Then we ran your old workerRows against the new module on the same payload to see the bug from outside the tests. The old path called three healthy dormant workers "Inactive", and it also called a genuinely active worker with zero processed items "Inactive", which nobody had shown before. The synthetic Temporal Scoring row that the server never sends is gone as well.

One aside that is now stale in your favour: your note about the vitest file not running in CI is not true any more. #679 landed the web-unit job, so worker-status-utils.test.js runs on every PR at no real cost.

We also checked the interaction with #772, which changed the same worker's flush cadence after you pushed. Both are atomic operations on the same Int32 and Stopped is only ever written on loop exit, so they cannot observe each other. #772's pinning tests pass alongside yours in the merged tree.

Three small things we are taking as follow-ups rather than sending back to you: the summary pill ranks stopped below dormant so a fully crashed set still reads "Dormant", the worker keys are hardcoded in the mapper, and an explicit null worker now renders as a Disabled row instead of being dropped (arguably the better behaviour, just noting it was not called out).

Merging. This is your second one in this area and the care shows, particularly wiring a CSS assertion to the embedded stylesheet so the class name and the rule cannot drift apart quietly.

Issue #596, opened 2026-07-10
Writes to a non-Cortex node are accepted locally and never forwarded (silent data divergence)


Sp00nman-420:
Issue 1 — Writes to a non-Cortex node are accepted locally and never forwarded (silent data divergence)

  Version: v0.7.0 and v0.8.0 · Severity: data loss, silent

  A muninn_remember sent to a Lobe returns 200 with an engram id, but the engram is written only to that node's
  local store. It is never forwarded to the Cortex and never replicated. No error, no warning in the log.

  Reproduce

  Three-node cluster, node-a = Cortex, node-b/node-c = Lobes.

  # 1. write to a LOBE via MCP
  curl -sk -X POST https://node-b:8750/mcp \
    -H "Authorization: Bearer $MUNINN_MCP_TOKEN" \
    -H 'Content-Type: application/json' \
    -H 'Accept: application/json, text/event-stream' \
    -d '{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{
          "name":"muninn_remember",
          "arguments":{"content":"written on a lobe","concept":"probe"}}}'
  # -> 200 {"id":"01K..."}

  # 2. read that id on the CORTEX and on the OTHER lobe
  curl -sk -X POST https://node-a:8750/mcp ... muninn_read {"id":"01K..."}   # -> not found
  curl -sk -X POST https://node-c:8750/mcp ... muninn_read {"id":"01K..."}   # -> not found

  # 3. read it back on node-b -> present

  Expected

  One of:
  - the Lobe forwards the write to the Cortex (transparent proxying), or
  - the Lobe rejects the write with an explicit "not the leader" error so the client can retry against the Cortex.

  Actual

  The write succeeds locally and diverges permanently. A client that reaches a replica — a load balancer without
  leader affinity, a stale DNS record, a VIP that fails over to a replica — loses data without any signal.

  Notes

  muninn_read on the Lobe returns the engram, so it is durably stored there; it simply never enters the replication
  stream. Replication in the other direction (Cortex → Lobes) works correctly.

  ---
  Issue 2 — Cluster replicates engrams but not configuration

  Version: v0.7.0 and v0.8.0 · Severity: correctness after failover

  Engrams replicate from the Cortex to the Lobes. Configuration written through the admin API does not, even though
  it is also a Pebble write.

  Reproduce

  # on the CORTEX
  curl -sk -b "$SESSION" -X PUT https://node-a:8476/api/admin/vault/probe/plasticity \
[...]

PR #646, opened 2026-07-21
fix(engine): carry entity links and relationships across evolve, funding the mention ledgers

### LINKED ISSUE #622: Evolve inherits Concept and Tags but drops entity links — evolved memories vanish from the entity index
### Version
Observed in production on a 2026-07-01 develop build; confirmed present at develop tip (`d7c4a4a`) by code read.

### Summary
`Evolve` inherits the predecessor's **Concept** and **Tags** but silently drops its **entity links**. The successor is written with no entity-index entries and no entity relationships, so an evolved memory vanishes from `find_by_entity` — permanently, in any deployment without an enrichment provider, and until re-extraction (which rebuilds from text, not from the curated links) in deployments with one.

### Where
`internal/engine/engine.go`, `Evolve` (~line 2863 at develop tip):

- `newEng` is built from `oldEng.Concept` / `oldEng.Tags` only; nothing reads the predecessor's entity links to carry them over.
- The normal write path stores caller-provided entities via `WriteEntityEngramLink` and entity-to-entity relationships in the 0x21 index (`engine.go` ~1070 and ~1598); `Evolve` calls neither, and its API surface accepts no inline entities either.

### Why it bites
The MCP guidance (and the tool descriptions) recommend "update with `muninn_evolve`, not forget+remember" — following the recommended path sheds the knowledge graph one rewrite at a time. We hit it in the field three separate ways before tracing it:

- A monitoring sentinel that was evolved once stopped being findable by its entity — it silently fabricated the exact absence it was planted to detect.
- Long-lived memories that get refined repeatedly (ours have evolve chains eight deep) end fully de-linked from every entity they mention.
- The operational workaround is a double write: evolve, then re-state the entities with a second write — which re-creates exactly the forget+remember shape evolve exists to replace.

Secondary observation, same audit: the successor's provenance reads as a plain `create` — the evolve lineage is visible only by finding the soft-deleted predecessor via `list_deleted` (the supersedes association exists, but nothing on the successor itself says it was born from an evolve).

### Suggested direction
Carry entity links across the evolve the same way Concept and Tags are carried: read the predecessor's entity links and write them for the successor inside the same atomic batch (and arguably accept optional inline `entities` on evolve, mirroring `remember`, for callers whose update changes what the memory is about). If you'd take a PR for this we're glad to build it — we already carry the read/write sides of both entity surfaces in #616 and the write path respectively, so the shape is familiar.

— Cairn (context: #609, #620)

isaac-ranger:
Fixes #622. Scope narrowed per review: this PR is now the write-path carry only. The inline-entities feature and the startup repair moved to their own PRs (stacked on this one) so the migration piece can get its independent pass without holding the fix hostage.

Evolve inherited Concept and Tags but wrote no 0x20/0x23 entity links and no 0x21/0x26 relationship records for the successor, so every evolved memory silently vanished from FindByEntity and entity scoring from the moment it was evolved.

**What this PR does**

- Evolve reads the predecessor's entity links and relationship records and queues them into the existing atomic batch via two new `StoreBatch` methods (`WriteEntityEngramLink`, `WriteRelationshipRecord`). No signature change.
- **The carried links fund the mention-count and co-occurrence ledgers** (review item 4 — the reversal). Both ledgers are one increment per link key created, one decrement per link key destroyed: `DeleteEngram` decrements unconditionally, so an unfunded carried link is a liability the pruner cashes in. Funding happens post-commit via a new `IncrementEntityMentionCount` (touches only MentionCount/UpdatedAt on the 0x1F record — never Type, Source or Confidence) and `IncrementEntityCoOccurrence`, mirroring `DeleteEngram`'s post-commit decrements and its 50-entity pair cap. Beyond the cap the ledger is best-effort on both sides (DeleteEngram's first-50 comes from map iteration), matching the pre-existing behaviour for >50-entity engrams rather than claiming to fix it.
- `TestEvolve_DoesNotDoubleCountMentions` is deleted and replaced by the conservation test you specified: write funds 1, evolve carries and funds to 2, hard-deleting the predecessor returns 1 **with the co-occurrence pair key still present** — the pair no longer vanishes from `ScanEntityClusters` while the successor still mentions both. The inverted reasoning is in the comment blocks on both the batch method and the funding site.
- `DigestEntities` is set on the successor so enrichment does not re-extract over the carried set.

**Verification**

Rebased on develop (d5f36e3, includes #655); `go vet -tags localassets ./...` clean and the two #655-added `Evolve` call sites compile — the branch is a fresh base, not the stale 21 July check. RED both ways: with the carry neutralized, `TestEvolve_CarriesEntityLinks`, `TestEvolve_CarriesRelationshipRecords` and the conservation test fail; with carry kept but funding disabled, the conservation test fails alone. Full `./internal/...` green, storage and engine under `-race`.

### DIFF
--- a/internal/engine/engine.go
+++ b/internal/engine/engine.go
@@ -3023,6 +3023,28 @@ func (e *Engine) Evolve(ctx context.Context, vault, oldID, newContent, reason st
 		Trust:      storage.TrustInferred, // all new MCP writes default to inferred
 	}
 
+	// Carry the predecessor's entity links and relationship records forward
+	// (#622): Evolve wrote no 0x20/0x23 links and no 0x21/0x26 relationship
+	// records for the successor, so every evolved memory silently vanished
+	// from FindByEntity and entity scoring from the moment it was evolved.
+	// The link and relationship keys are queued into the same atomic batch as
+	// the engram writes; the mention-count and co-occurrence ledgers are funded
+	// post-commit (see below).
+	var carriedEntities []string
+	var carriedRels []storage.RelationshipRecord
+	if err := e.store.ScanEngramEntities(ctx, wsPrefix, oldULID, func(name string) error {
+		carriedEntities = append(carriedEntities, name)
+		return nil
+	}); err != nil {
+		return storage.ULID{}, fmt.Errorf("evolve: scan predecessor entities: %w", err)
+	}
+	if err := e.store.ScanEngramRelationships(ctx, wsPrefix, oldULID, func(rec storage.RelationshipRecord) error {
+		carriedRels = append(carriedRels, rec)
+		return nil
+	}); err != nil {
+		return storage.ULID{}, fmt.Errorf("evolve: scan predecessor relationships: %w", err)
+	}
+
 	// Build the supersedes association (new → old).
 	supersedes := &storage.Association{
 		TargetID:      oldULID,
@@ -3046,8 +3068,62 @@ func (e *Engine) Evolve(ctx context.Context, vault, oldID, newContent, reason st
 	if err := batch.UpdateEngramState(ctx, wsPrefix, oldULID, storage.StateSoftDeleted); err != nil {
 		return storage.ULID{}, fmt.Errorf("evolve: batch update old state: %w", err)
 	}
+	for _, name := range carriedEntities {
+		if err := batch.WriteEntityEngramLink(ctx, wsPrefix, newULID, name); err != nil {
+			return storage.ULID{}, fmt.Errorf("evolve: batch carry entity link: %w", err)
+		}
+	}
+	for _, rec := range carriedRels {
+		if err := batch.WriteRelationshipRecord(ctx, wsPrefix, newULID, rec); err != nil {
+			return storage.ULID{}, fmt.Errorf("evolve: batch carry relationship: %w", err)
+		}
+	}
 	if err := batch.Commit(); err != nil {
 		return storage.ULID{}, fmt.Errorf("evolve: batch commit: %w", err)
+	}
+
+	// Fund the ledgers for the carried links. The mention-count and
+	// co-occurrence ledgers are one increment per link key created, one
+	// decrement per link key destroyed: DeleteEngram decrements both
+	// unconditionally for every link it removes, so a carried link with no
+	// matching increment becomes an unfunded liability that the pruner cashes
+	// in — hard-delete the predecessor and MentionCount under-reads while the
+	// successor is still linked, and a co-occurrence pair drops to zero and
+	// vanishes from ScanEntityClusters while both entities are still mentioned
+	// together. Done post-commit, mirroring DeleteEngram's post-commit
+	// decrements: a crash here leaves counts slightly low with the links
+	// intact, the mirror image of DeleteEngram's slightly-high stale case.
+	for _, name := range carriedEntities {
+		if err := e.store.IncrementEntityMentionCount(ctx, name); err != nil {
+			slog.Warn("engine: evolve: failed to increment mention count for carried entity", "entity", name, "engram", newULID.String(), "err", err)
+		}
+	}
+	// Cap mirrors DeleteEngram's maxCoOccurrenceEntities. Beyond the cap the
+	// ledger is best-effort on both sides — DeleteEngram's first-50 comes from
+	// map iteration, so the truncated subsets need not be identical — matching
+	// the pre-existing behaviour for >50-entity engrams rather than fixing it.
+	const maxCoOccurrenceEntities = 50
+	coNames := carriedEntities
+	if len(coNames) > maxCoOccurrenceEntities {
+		slog.Warn("engine: evolve: engram has unusually many carried entities, co-occurrence funding capped",
+			"engram", newULID.String(), "entity_count", len(carriedEntities), "cap", maxCoOccurrenceEntities)
+		coNames = coNames[:maxCoOccurrenceEntities]
+	}
+	for i := 0; i < len(coNames); i++ {
+		for j := i + 1; j < len(coNames); j++ {
+			if err := e.store.IncrementEntityCoOccurrence(ctx, wsPrefix, coNames[i], coNames[j]); err != nil {
+				slog.Warn("engine: evolve: failed to increment co-occurrence for carried pair", "a", coNames[i], "b", coNames[j], "engram", newULID.String(), "err", err)
+			}
+		}
+	}
+
+	// Mark entity extraction complete when the successor carries a curated set,
+	// so the retroactive enrichment provider does not re-extract over it.
+	if len(carriedEntities) > 0 {
+		existingFlags, _ := e.store.GetDigestFlags(ctx, plugin.ULID(newULID))
+		if err := e.store.SetDigestFlag(ctx, newULID, existingFlags|plugin.DigestEntities); err != nil {
+			slog.Warn("engine: evolve: failed to set DigestEntities flag", "id", newULID.String(), "err", err)
+		}
 	}
 
 	// ── Content-hash bookkeeping: delete old mapping, add new mapping ──
--- a/internal/engine/engine_evolve_entity_test.go
+++ b/internal/engine/engine_evolve_entity_test.go
@@ -1,0 +1,184 @@
+package engine
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/scrypster/muninndb/internal/plugin"
+	"github.com/scrypster/muninndb/internal/storage"
+	"github.com/scrypster/muninndb/internal/transport/mbp"
+)
+
+// engramEntities returns the entity names linked to id via the 0x20 forward index.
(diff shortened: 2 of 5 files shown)


scrypster (changes requested):
Thanks for this, and for the startup repair in particular. Recognising that existing vaults are already carrying stripped successors, and that fixing the write path alone leaves them broken forever, is the right instinct. The carry itself is sound. Reverting it turns `TestEvolve_CarriesEntityLinks` and `TestEvolve_CarriesRelationshipRecords` red, and reverting the repair turns both repair tests red, so the core of this is proven rather than asserted.

Three things block it, and one of them is not visible from your branch.

**1. The merge does not compile.**

#655 landed after you opened this and added two tests that call the 7-arg `Evolve`. You change `Evolve` to 8 args. Neither side conflicts textually, so git merges them into broken code:

```
vet: internal/engine/engine_evolve_test.go:183:72: not enough arguments in call to eng.Evolve
```

Your branch alone is clean, and the green check on this PR is from 21 July, six days before #655 merged, so it is a statement about a base that no longer exists. Rebase and add the `nil` entities argument at `engine_evolve_test.go:183` and `:219`.

Trap: `go build ./...` passes in the broken state because it does not compile test files. Gate on `go vet ./...` or `go test`.

**2. The startup repair is not crash-safe, and a partial repair becomes permanent.**

Each carried link is a separate store-level call, and each of those commits its own `NoSync` Pebble batch. The skip test is "does the successor have any entity link at all", early-exiting on the first hit. So a successor stranded with six entities that gets killed after link two is skipped forever on every later boot, and links three through six are never recovered. That is the same silent loss of reachability the issue is about, reintroduced by its own fix. A rolling restart or an OOM is enough, no unusual conditions needed.

The fix is already in this PR and unused. You add `pebbleStoreBatch.WriteEntityEngramLink` and `WriteRelationshipRecord` exactly so this can be atomic, then the repair calls the store-level singletons instead. Repair one successor per `StoreBatch`, covering all its links, its relationship records, and the digest flag, committed once.

Trap: `pebbleStoreBatch.Commit()` does not call `replicateBatch`, whereas the store-level methods do. Switching to the batch silently drops replication of the repair writes. Make that a deliberate choice rather than a side effect.

**3. MentionCount, and the same hole in co-occurrence.**

I need to reverse what I told you in the issue thread. I said not to increment on carry because it would double-count. That was wrong, and I am sorry for sending you down that path.

The ledger is already one increment per link key created and one decrement per link key destroyed. `DeleteEngram` decrements unconditionally, once per link it removes. A carried link with no matching increment is an unfunded liability, and the pruner cashes it in: write an engram mentioning Ada, evolve it, let retention hard-delete the predecessor, and `MentionCount` reads 0 while the successor is still linked to Ada. It floors at 0 and the entity record survives, so `find_by_entity` still works, but the number a user sees is simply wrong. The direction matters too, since 0 reads as "unused, safe to ignore".

That is the recommended path, not an edge case. The guide tells agents to use evolve rather than forget and remember, and any vault with `RetentionDays` or `MaxEngrams` prunes.

Co-occurrence has the identical hole and it is worse. `DeleteEngram` decrements every entity pair, and `DecrementEntityCoOccurrence` deletes the 0x24 key at zero. So after the predecessor is pruned, the Ada to Grace edge is gone while the successor still mentions both. `ScanEntityClusters` reads those keys with a minimum count of 2, so even a drop from 2 to 1 removes the pair from clustering. That is a missing graph edge, not a display number.

Please increment both on carry.

**Required changes**

1. Rebase on develop and fix the two `Evolve` call sites. Gate on `go vet`, not `go build`.
2. Make the per-successor repair atomic in one `StoreBatch`. Decide the replication question deliberately.
3. Take `casLocks.For(succID)` around that batch. The repair currently races `DeleteEngram`, which holds that lock while removing the same 0x20/0x23/0x21/0x26 keys the repair writes, so an in-flight repair write can land after a hard delete and leave keys pointing at a deleted engram.
4. Increment `MentionCount` and co-occurrence for carried entities and pairs. Delete `TestEvolve_DoesNotDoubleCountMentions` and replace it with a conservation test: write gives 1, evolve with carry gives 2, hard-deleting the predecessor gives 1 with the pair key still present. The comment block explaining the no-increment reasoning needs inverting too.
5. Add a per-vault repair watermark, or an env opt-in, so a healed vault does not pay a full soft-deleted scan on every boot forever. Add startup delay and jitter to match `runPruneWorker`, which waits 60s plus jitter rather than competing with HNSW load and first traffic.
6. Add a test that exercises `runEvolveEntityLinkRepair` through `NewEngine` and `Stop`. Both current repair tests call `repairEvolveEntityLinksOnce` directly, so deleting the spawn line leaves everything green. For a backfill, the wiring is the part you least want unproven.
7. Strengthen `TestEvolve_InlineEntitiesReplaceCarry`. It currently passes with the carry disabled, because with no carry there is nothing to replace.
8. Document the "successor with at least one link is skipped" limitation in the log line and the body, since "repaired=N" will read as complete.

**On scope: please split this.**

Around ten of the eighteen files are one-line call-site updates from the signature change. The real content is three separable pieces. The carry is tight, correct and ready once rebased. Inline entities on evolve is an additive feature. The startup repair is 138 new lines, a new scan primitive, and a one-way backfill over existing user data, and it carries every finding above except the first.

Landing the first two now stops the bleeding, which is most of the value since the damage accrues per evolve going forward. The repair can then get the crash-safety and watermark work it needs without holding the fix hostage.

That last piece is Tier 3 by our own rubric, since it is a migration over existing records, so it needs a second independent pass before it merges. I would rather that pass look at 138 focused lines than at all three changes at once. I will arrange it when you resubmit.

isaac-ranger:
All eight items landed, across the split you asked for. Map:

- **Split** — this PR is now the carry alone (force-pushed as a fresh single commit on develop). Inline entities → #680, startup repair → #681, both stacked on this branch. #681 is the Tier-3 piece awaiting your second reviewer.
- **1 (broken merge)** — rebased on d5f36e3; the two #655 call sites compile with the carry in place. The signature change itself moved to #680, so this PR no longer touches `Evolve`'s arity at all. Gated on `go vet`, not `go build` — confirmed the trap: `go build ./...` stays green with the test files broken.
- **4 (the reversal)** — carried links now fund both ledgers, and thank you for the correction: your unfunded-liability framing is in the comments verbatim-adjacent, and `TestEvolve_DoesNotDoubleCountMentions` is replaced by your conservation sequence (1 → 2 → 1 with the pair key surviving the predecessor's hard delete). One honest caveat found while mirroring the decrement: beyond the 50-pair cap the two sides already truncate *different* subsets on develop (DeleteEngram's names come from map iteration), so the comment claims best-effort parity there, not exactness.
- **2, 3, 5, 6, 8** — in #681: per-successor atomic batch (your unused-batch-methods observation was the fix, as you said), stripe lock against DeleteEngram, watermark + 60s/jitter delay, a NewEngine/Stop lifecycle test that goes red if the spawn line is deleted, and an unconditional sweep log that names the skip limitation. The replication question is decided deliberately (not replicated — every node self-repairs; reasoning in the method comment).
- **7** — in #680: the replace-carry test now contains a carry control leg on the same predecessor shape, so it fails with the carry disabled instead of passing vacuously.

One addition beyond the review: an adversarial pass on the repair before resubmitting found that exhausting the fixpoint round cap while rounds were still repairing would have watermarked a still-damaged vault — the same stranding class at the loop level. Fixed (only a converged pass watermarks; the vault resumes next boot) and pinned with a 40-hop descending-ULID chain test, the one shape where the cap binds.

scrypster (approved):
This is a clean one, Cairn, and it's the right root fix for #622.

We cloned the branch and ran it with `-race` (storage and engine both green), then RED-sanity checked it the real way: reverted just your engine.go change, kept the new tests, and all three of `TestEvolve_CarriesEntityLinks`, `_MentionCountConservation`, and `_CarriesRelationshipRecords` go red (missing links, mention count 1 instead of 2, zero relationship records). Put engine.go back and they pass. So it's a genuine fix, not a test that passes both ways. We also confirmed the "+351/-0, pure additive" framing holds literally: nothing on a non-evolve write path moved, the new batch and entity helpers are net-new functions, and the diff outside evolve is empty.

The conservation test is the part worth calling out. write=1, carry=2, hard-delete the predecessor, back to 1, with the co-occurrence pair required to survive. That's exactly the "pruner cashing in an unfunded liability" case your commit message worries about, asserted end to end instead of just checking the increment fired. Same measure-first discipline you've brought since the recall-events work in #574 and the HNSW integrity pass in #471, and it's why your evolve series keeps landing clean.

One thing to carry forward, not a blocker: `pebbleStoreBatch.Commit()` (the abstraction Evolve uses) never calls `replicateBatch`, so in a cluster the whole evolve write, entity links included, never reaches followers. That predates this PR (it came in with the atomic-batch evolve refactor), and your funding calls faithfully mirror the existing unreplicated decrement side, so you're symmetric with what's already there rather than adding a new gap. We're noting it under the #596 replication class rather than blocking on it here. It matters for you specifically because #680 and #681 stack on this: if either adds a new cluster-facing write through the same batch path, that's where "pre-existing, not mine" stops holding.

Merging now. Thanks for taking #622 all the way to the ledger funding instead of stopping once the links carried.

PR #651, opened 2026-07-23
Add exact-concept lookup and newest-first capped entity reads

koalaoptics:
## Summary

- add a vault-scoped exact-concept reverse index at the next free registered prefix (`0x2B`)
- backfill that derived index with an idempotent migration and maintain it atomically on writes and hard deletes
- expose case-sensitive exact-concept lookup through the engine and MCP surface
- preserve upstream fuzzy entity resolution while making capped entity reads select the newest live records
- retain chronological presentation for the newest capped timeline window

## Compatibility and safety

- `0x29` remains `RecallEvent` and `0x2A` remains `Lease`; this change does not reuse either prefix
- hash candidates are hydrated and compared against the complete concept string
- soft-deleted and archived records are excluded
- vault clearing includes `0x2B`
- no deployment, image publication, production migration, resolver change, or production-data operation is included

## Verification

- `go test ./...`
- `go vet ./...`
- `go test -race ./internal/storage ./internal/storage/migrate ./internal/engine ./internal/mcp`
- standard and `localassets` builds
- each sequential commit independently passed full test/vet and the repository methodology gate
- deterministic synthetic comparison: Koala `v0.4.15-alpha` exact/newest behavior vs stock upstream `v0.9.0` vs this candidate; candidate restored both compatibility behaviors with no application errors
- final GitNexus scope: 33 files, 96 symbols, 10 affected execution flows, HIGH risk; write, migration, lifecycle, MCP, and capped-read paths are covered by the checks above

## Review note

Automated independent review attempts did not return a usable report and therefore carry no approval weight. Deterministic tests and the synthetic compatibility receipt are the evidence for this source-only candidate.


koalaoptics:
Verification update:

- published tip `baa7a41631878be379cecf9bd88a83ad4e79f7bb` matches the locally verified tip
- upstream `main` (`3eed40c928450b93a237a1b15f847363ac6e3a0c`) is an ancestor of the branch
- published PR scope is exactly the 33 qualified source/test/doc files
- no CI checks are configured or reported for this PR

The remaining gate is an independent upstream review; the author account cannot self-approve. No deployment or production migration is part of this PR.

koalaoptics:
Closing because this compatibility work is Koala-fork-specific and was mistakenly opened against upstream. We are retaining it only on the koala-optics fork; no upstream contribution is intended.

PR #655, opened 2026-07-25
fix(engine): inherit MemoryType and TypeLabel across evolve

### LINKED ISSUE #653: Evolve drops Summary and silently resets MemoryType to the zero value
### Summary

`Engine.Evolve` explicitly inherits `Concept` and `Tags` from the predecessor but never sets `Summary`, `MemoryType`, or `TypeLabel` on the successor. The successor is therefore born with an empty summary and its type silently reset to the zero value, `Fact`.

This is the same constructor as #622, and a sibling defect: that issue covers the dropped entity links (PR #646), this one covers the dropped scalar fields. #646 does not address these.

### Root cause

`internal/engine/engine.go`, `Engine.Evolve`:

```go
if concept == "" {
    concept = oldEng.Concept
}
newEng := &storage.Engram{
    ID:         newULID,
    Concept:    concept,
    Content:    newContent,
    Tags:       oldEng.Tags,
    Confidence: 1.0,
    Stability:  30.0,
    State:      storage.StateActive,
    CreatedAt:  now,
    UpdatedAt:  now,
    LastAccess: now,
    Embedding:  embedding,
    Trust:      storage.TrustInferred,
}
```

`storage.Engram` carries `Summary`, `MemoryType`, and `TypeLabel`, and all three are absent from this literal. `Summary` becomes `""`; `MemoryType` becomes `0`, which is `Fact`.

The failure is by omission — the constructor inherits some predecessor fields and drops others, with no signal at either the API or the log that anything was lost.

### Reproduction

Against a v0.9.0 daemon (this run: v0.9.0 plus local patches, `244310a`), over MCP:

1. `muninn_remember` with an explicit `summary` and `type: "procedure"`.
2. `muninn_evolve` with only `id`, `new_content`, `reason` — no `concept`, no `entities`, so everything unspecified should inherit.
3. `muninn_read` the successor.

Observed on the successor:

| Field | Predecessor | Successor | |
|---|---|---|---|
| `concept` | set | inherited | ok |
| `tags` | set | inherited | ok |
| `entities` | set | inherited | ok (#646 behavior) |
| `summary` | sentinel string | **absent** | dropped |
| `type` | `procedure` | **`fact`** | silently reset |

### Impact

The type reset is unconditional and is the more serious half: a `decision` or `procedure` becomes a `fact` on evolve, with no error. Any consumer filtering or routing on memory type sees the successor as a different kind of thing than the predecessor, and a supersede chain degrades every hop.

The summary half is deployment-dependent. On a vault with no enrich provider configured the summary simply stays empty. On a vault with enrichment it may be regenerated asynchronously — different text rather than empty — so the visible severity varies.

### Suggested fix

Inherit all three in the constructor, matching the existing `Concept`/`Tags` treatment, with caller-supplied values taking precedence where the API exposes them:

```go
Summary:    oldEng.Summary,
MemoryType: oldEng.MemoryType,
TypeLabel:  oldEng.TypeLabel,
```

Worth considering more broadly whether this constructor should start from a copy of the predecessor and overwrite the fields evolve genuinely changes, rather than enumerating fields to carry — the enumerate-what-to-carry shape is what produced both this and #622, and will produce the next one when a field is added to `storage.Engram`.

Happy to send a PR if the approach looks right, and equally happy to fold it into #646 if you would rather keep the evolve field-carry work in one place.

isaac-ranger:
Fixes #653.

Sent ahead of a maintainer reply on the issue since the change is small and self-contained — happy to close it if you would rather shape the approach differently.

**Scope note:** an earlier revision of this PR also inherited `Summary`. That was wrong and has been removed — see the [correction comment](https://github.com/scrypster/muninndb/issues/653#issuecomment-5080967748) on the issue for the trace. The short version: `Summary` is content-derived, `Evolve` replaces the content, and a non-empty `Summary` reads as caller-authoritative to both `engramHasSummary` and the retroactive processor — so inheriting it would pin stale text permanently *and* leave `KeyPoints` empty forever, since the summarize stage is their only producer.

### The change

`Engine.Evolve` built its successor with a struct literal that explicitly inherited `Concept` and `Tags` and never set `MemoryType` or `TypeLabel`. Evolve exposes no parameter for either, so they took the Go zero value — and `MemoryType`'s zero value is `Fact`. A stored `decision` or `procedure` was not left "unclassified", it was silently relabelled a fact, with no error and nothing in the log.

Two lines, inheriting alongside the fields already carried:

```go
MemoryType: oldEng.MemoryType,
TypeLabel:  oldEng.TypeLabel,
```

Inheriting the type does suppress LLM reclassification, via `engramHasClassification`. That is deliberate and is the difference from the summary case: a memory's type is not invalidated by editing its content — an updated `decision` is still a `decision` — so the inherited value is true of the successor, and suppressing the stage also saves an LLM call per evolve. A summary describes specific text; replace the text and the description is false.

### Tests

- `TestEvolve_InheritsMemoryTypeAndTypeLabel` — single evolve. Asserts the predecessor's own fields first, so a failure indicts `Evolve` rather than `Write`. Fails without the fix: `0x1` vs `0x0`, `"architectural_decision"` vs `""`.
- `TestEvolve_TypeSurvivesSupersedeChain` — three hops, failing at every hop without the fix. A single-hop assertion would still pass if inheritance ever read from a default-filled intermediate, and the compounding case is the one that destroys a long-lived memory.

The assertion that the successor's `Summary` is empty is a **regression guard against re-introducing the inheritance — it does not fail without the fix.** Flagging that explicitly so it isn't mistaken for a teeth-bearing test.

`go build ./...`, `go vet`, `gofmt -l` clean; `./internal/engine/...` green.

### One design note, deliberately not acted on here

This constructor enumerates which predecessor fields to carry, and that shape is what produced both this bug and #622 — it will produce the next one when a field is added to `storage.Engram`. Starting from a copy of the predecessor and overwriting only what evolve genuinely changes would make the failure mode unrepresentable rather than policy-checked. Larger change, more review surface, so this PR stays minimal and matches the existing style. Glad to do it the other way if you prefer.

Two adjacent losses found while checking and deliberately left alone: `Consolidate` writes its merged engram with only a concept and content, dropping all source types, summaries and tags — same class, but many-to-one and out of this issue's scope. `CreatedBy` and `Classification` are dropped too but appear to be dead fields with no writer anywhere in the tree.

### Relationship to #646

That PR carries entity links and relationships across evolve — same constructor, different fields. Independent diffs; whichever lands first, the other should rebase cleanly.

### DIFF
--- a/internal/engine/engine.go
+++ b/internal/engine/engine.go
@@ -2993,11 +2993,26 @@ func (e *Engine) Evolve(ctx context.Context, vault, oldID, newContent, reason st
 	if concept == "" {
 		concept = oldEng.Concept
 	}
+	// MemoryType and TypeLabel are inherited alongside Concept and Tags. Evolve
+	// exposes no parameter for either, so leaving them unset reset the type to
+	// the MemoryType zero value, Fact — a decision or procedure was silently
+	// relabelled a fact on every evolve, and again at each hop down a supersede
+	// chain (issue #653).
+	//
+	// Summary is deliberately NOT inherited. It is content-derived (see the
+	// field doc on storage.Engram), and Evolve replaces the content, so the
+	// predecessor's summary is false of the successor. Leaving it empty is also
+	// what lets the summarize stage regenerate it: engramHasSummary and the
+	// retroactive processor both treat a non-empty Summary as caller-authoritative
+	// and skip the stage, which would pin the stale text permanently and leave
+	// KeyPoints — produced only by that stage — empty forever.
 	newEng := &storage.Engram{
 		ID:         newULID,
 		Concept:    concept,
 		Content:    newContent,
 		Tags:       oldEng.Tags,
+		MemoryType: oldEng.MemoryType,
+		TypeLabel:  oldEng.TypeLabel,
 		Confidence: 1.0,
 		Stability:  30.0,
 		State:      storage.StateActive,
--- a/internal/engine/engine_evolve_test.go
+++ b/internal/engine/engine_evolve_test.go
@@ -135,3 +135,96 @@ func TestEvolve_ConceptRenameWhenProvided(t *testing.T) {
 	assert.Equal(t, "answer owed to Alice on #895", oldEng.Concept,
 		"predecessor concept must not be mutated")
 }
+
+// TestEvolve_InheritsMemoryTypeAndTypeLabel verifies that MemoryType and
+// TypeLabel are carried to the successor, alongside Concept and Tags.
+//
+// Evolve exposes no parameter for either, so before #653 they were simply
+// absent from the successor's struct literal and MemoryType fell back to its
+// zero value, Fact. A stored decision or procedure silently became a fact, and
+// a caller filtering or routing on memory type saw the successor as a different
+// kind of thing than the predecessor.
+//
+// Summary is asserted to be EMPTY on the successor, not inherited: it is
+// content-derived, Evolve replaces the content, and a non-empty Summary would
+// suppress the summarize stage that regenerates both it and KeyPoints.
+func TestEvolve_InheritsMemoryTypeAndTypeLabel(t *testing.T) {
+	eng, cleanup := testEnv(t)
+	defer cleanup()
+	ctx := context.Background()
+
+	const (
+		originalSummary   = "one-line summary describing the predecessor content"
+		originalTypeLabel = "architectural_decision"
+	)
+
+	resp, err := eng.Write(ctx, &mbp.WriteRequest{
+		Vault:      "test",
+		Concept:    "Original",
+		Content:    "v1",
+		Summary:    originalSummary,
+		MemoryType: uint8(storage.TypeDecision),
+		TypeLabel:  originalTypeLabel,
+	})
+	require.NoError(t, err)
+
+	ws := eng.store.ResolveVaultPrefix("test")
+
+	// Guard: the predecessor really carries what the test claims it does, so a
+	// failure below indicts Evolve rather than Write.
+	oldULID, err := storage.ParseULID(resp.ID)
+	require.NoError(t, err)
+	pred, err := eng.store.GetEngram(ctx, ws, oldULID)
+	require.NoError(t, err)
+	require.NotNil(t, pred)
+	require.Equal(t, originalSummary, pred.Summary, "precondition: predecessor summary")
+	require.Equal(t, storage.TypeDecision, pred.MemoryType, "precondition: predecessor type")
+
+	newID, err := eng.Evolve(ctx, "test", resp.ID, "v2", "update", nil, "")
+	require.NoError(t, err)
+
+	succ, err := eng.store.GetEngram(ctx, ws, newID)
+	require.NoError(t, err)
+	require.NotNil(t, succ)
+
+	assert.Empty(t, succ.Summary,
+		"summary must NOT be inherited: it describes the predecessor's content, and a "+
+			"non-empty value suppresses the summarize stage that regenerates it and KeyPoints")
+	assert.Equal(t, storage.TypeDecision, succ.MemoryType,
+		"memory type must be inherited, not reset to the zero value (Fact)")
+	assert.Equal(t, originalTypeLabel, succ.TypeLabel,
+		"free-form type label must be inherited")
+}
+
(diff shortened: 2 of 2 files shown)


isaac-ranger:
Correcting myself on the summary half of this issue — I think I got it wrong, and the PR needs to shrink.

### The summary half is not a defect

Evolve replaces the engram's **content**. A summary inherited from the predecessor describes text that no longer exists on the successor. Leaving `Summary` empty is not a loss — it is the accurate state of a memory whose content just changed and has not been re-summarized yet.

Worse, inheriting it actively prevents the correct summary from ever being produced. `RetroactiveProcessor.processEngram` gates on:

```go
hasSummary := eng.Summary != "" || (flags&DigestSummarized != 0)
```

and when that is true it discards whatever the enrich plugin produced:

```go
if hasSummary {
    result.Summary = eng.Summary
    ...
}
```

An evolved successor carries digest flags `0` (only `DigestEmbed` is stamped, after the batch), so it lacks `DigestEnrich` and **is** selected by `CountWithoutFlag`. Today it gets swept, summarized, and ends up with a summary describing its own content. If `Summary` were inherited, the non-empty string would satisfy `hasSummary`, the generated summary would be thrown away, and the predecessor's stale text would be pinned to the successor permanently.

So on any deployment with an enrich provider configured, inheriting `Summary` trades an empty field for a permanently wrong one. That is a worse failure than the one I reported.

I should have caught this before opening the PR rather than after. My own issue text even noted that enrichment might regenerate the summary asynchronously, and I then proposed a change that removes exactly that regeneration.

### What still stands

`MemoryType` and `TypeLabel` are a different case and the report holds for them.

To be precise about the mechanism, since it is the *same* mechanism: classification is field-gated too, symmetrically with summarization —

```go
// enrich/pipeline.go
if p.stageEnabled("classification") && !engramHasClassification(eng) { ... }

func engramHasClassification(eng *storage.Engram) bool {
    return eng.MemoryType != storage.TypeFact || eng.TypeLabel != ""
}
```

So inheriting the type also suppresses re-classification. The difference is not structural, it is whether the inherited value is *true of the successor*:

- A memory's type is not invalidated by editing its content. A `decision` that gets updated is still a `decision`, so the inherited value is correct and suppressing re-classification is desirable — it also saves an LLM call per evolve.
- A summary is a description of specific text. Replace the text and the description is false, so suppressing re-summarization pins a wrong value.

The zero value being `Fact` is what makes the current behavior harmful rather than merely absent: the successor is not "unclassified", it is "silently relabelled as a fact", and on a deployment with no enrich provider that relabelling is permanent and compounds at every hop of a supersede chain.

### What I have done

PR #655 is amended to inherit `MemoryType` and `TypeLabel` only. The summary assertions are removed from the tests; what remains is the type inheritance across a single evolve and across a three-hop chain, both still failing without the fix.

### Open question for you

If you do want a summary to survive an evolve on deployments with no enrich provider — where an empty summary stays empty forever — that needs a way to distinguish "inherited, please refresh" from "caller-supplied, leave alone." The digest flags look like the natural place, but it changes the meaning of `hasSummary` and I did not want to make that call unilaterally inside a bugfix. Happy to take it up separately if it is worth having.

scrypster (approved):
This is really good work, and the correction on `Summary` is the part I want to call out. You caught it yourself, traced why inheriting it would be actively worse than leaving it empty, and removed it before I ever looked. I verified that reasoning and it holds. `engramHasSummary` is a plain field check, so an inherited summary would suppress the only stage that produces `KeyPoints`, and the retroactive processor would then overwrite a fresh summary with the stale one. You were right to pull it.

What I checked. Reverting only `engine.go` and keeping your tests turns both red, and the chain test fails at every hop rather than just the first, so it genuinely pins the compounding case. Restoring turns them green. The engine tree passes with `-race`. I also merged current `develop` in and re-ran, since your branch is behind and `develop` has moved in `internal/engine`, and that is clean too. No cognition invariant is touched, and no cross-surface obligation fires.

Also worth saying: you flagged that `assert.Empty(succ.Summary)` passes with or without the fix. That is accurate, and pointing it out unprompted rather than letting it read as evidence is the right instinct.

Your instinct in the design note is right too. Enumerating fields to carry has now produced #622 and #653 and it will produce another. Building the successor from a copy and explicitly clearing what evolve should not keep would make the whole class unrepresentable. Correct to keep it out of a bugfix though.

This is a nice follow-on to the memory-type work you did in #616. Merging.

PR #656, opened 2026-07-25
fix(engine): apply meta filters and trust exclusion to entity-boost injected results

### LINKED ISSUE #654: Post-pipeline entity boost bypasses meta filters: tags_all/tags_any and created_after/before are not applied to injected results
### Summary

The post-pipeline entity-boost pass injects engrams that the activation pipeline never retrieved, and those injected results are never checked against the request's meta filters. A recall constrained by `tags_all`/`tags_any` or by `created_after`/`created_before` returns rows that satisfy none of those constraints.

This is not the flooding behavior in #569, and it is **not fixed by #570** — the daemon these results come from already runs #570's rarity weighting, accumulation cap, and threshold gate, and still leaks. #570 controls *how much* gets injected; this issue is about injected rows escaping the filter gate entirely.

### Root cause

`passesMetaFilter` is the correctness gate, and every call site is inside `ActivationEngine.Run` (`internal/engine/activation/engine.go`, four sites in the retrieval phases). The comment on `seedTagCandidates` states the intent directly:

> The tag index stores Hash(tag) (4-byte), so a seeded ID can be a hash-collision false positive; passesMetaFilter in phase 6 remains the correctness gate.

`applyEntityBoost` runs *after* `Run` returns:

```go
result, err := e.activation.Run(ctx, actReq)
...
result.Activations = e.applyEntityBoost(ctx, wsPrefix, result.Activations, actReq.Threshold)
```

and its signature is

```go
func (e *Engine) applyEntityBoost(ctx context.Context, ws [8]byte, results []activation.ScoredEngram, threshold float64) []activation.ScoredEngram
```

It receives no filters. It cannot honor them — the information never reaches it.

Only the pass-2b injection path leaks. Pass-2a boosts engrams already present in the result set, and those already passed phase 6. Pass-2b fetches engrams the pipeline never returned and appends them after a hand-rolled lifecycle check:

```go
if eng.State == storage.StateSoftDeleted || eng.State == storage.StateArchived {
    continue
}
```

That check is the only predicate applied. `tags_all`, `tags_any`, `tag_prefix`, `created_after`, `created_before`, and explicit `state` filters are all skipped.

### Reproduction

Live daemon, v0.9.0 + local patches (`244310a`, includes #570). Vault contains memories spanning several weeks.

```
muninn_recall  context=["Trane thermostat outage reseat"]
               since=2026-07-25T00:00:00Z
               threshold=0.1  limit=15
```

15 rows returned. Six were created 2026-07-13/14, well outside the `since` bound. Each of the six carries `score == entity_boost == 0.221076` — no content score at all, i.e. pure pass-2b injection:

```
01KXF7FRJV...  created 2026-07-14  score 0.221076  entity_boost 0.221076
01KXF78GRX...  created 2026-07-14  score 0.221076  entity_boost 0.221076
01KXESMKEH...  created 2026-07-13  score 0.221076  entity_boost 0.221076
01KXH3WVXF...  created 2026-07-14  score 0.221076  entity_boost 0.221076
01KXER1QNG...  created 2026-07-13  score 0.221076  entity_boost 0.221076
01KXH3J59H...  created 2026-07-14  score 0.221076  entity_boost 0.221076
```

The same bypass is reproducible with `tags_all` on a tag none of the injected rows carry.

### The bug is intermittent, and here is why

`applyEntityBoost` returns early when the pipeline produced nothing:

```go
if len(results) == 0 {
    return results
}
```

Injection seeds from the top-N surviving results, so a filter restrictive enough to empty the result set also suppresses the leak. A first reproduction attempt returned a clean zero for exactly this reason. **The bypass only manifests when the filtered query still returns at least one result to seed from** — which is worth knowing before anyone tries to reproduce it with a maximally narrow filter and concludes it does not exist.

### Secondary observation, same mechanism

The response above reported `"total": 8` while returning 15 rows. 8 is the phase-6-filtered count; the 7 pass-2b injections are appended afterward and never counted. Whatever the fix for the filter bypass, `total` and the returned rows should agree.

### Suggested fix

Thread the request filters into `applyEntityBoost` and apply `passesMetaFilter` to every pass-2b candidate before appending, next to the existing lifecycle check. Pass-2a needs no change.

Doing the filter check before `GetEngram` is not possible since the predicate needs the full engram, but the existing threshold gate already bounds how many candidates reach that point.

A structural alternative, if you would rather not have a second gate to keep in sync: move the entity-boost pass inside `Run` so phase 6 remains the single correctness gate as the `seedTagCandidates` comment intends. That is a larger change and I have not attempted it.

Glad to send a PR for the threaded-filters version on your word.

isaac-ranger:
Fixes #654.

Sent ahead of a maintainer reply on the issue — happy to close it if you would rather take a different approach, particularly the structural alternative at the bottom.

### The change

`applyEntityBoost` runs after `activation.Run` returns, and therefore after phase 6. Its injection path fetched engrams the pipeline had never retrieved and appended them behind a hand-rolled lifecycle check:

```go
if eng.State == storage.StateSoftDeleted || eng.State == storage.StateArchived {
    return nil
}
```

That was the only predicate applied. `tags_all`, `tags_any`, `tag_prefix`, `created_after`, `created_before`, explicit `state` filters and `ExcludeUntrusted` were all skipped. The pass received none of those inputs — it could not have honored them.

This threads the request's filters and `ExcludeUntrusted` through, and gates injected candidates on `activation.PassesMetaFilter`, a thin exported wrapper over the same predicate phase 6 uses so the two cannot drift.

The trust check mirrors phase 6 exactly: only `TrustUntrusted` is excluded, and `TrustUnset` passes through as the backward-compat alias for `TrustInferred`. There is an assertion pinning that, so the fix cannot quietly become an over-exclusion.

**Scope widened from the filed issue, deliberately.** The issue reported the meta-filter bypass; this also fixes `ExcludeUntrusted`. It is a vault operator's standing posture rather than a per-call convenience — `plasticity.go` documents untrusted engrams as filtered from ACTIVATE results, and the MCP tool surface advertises the guarantee to end users — which places it above the meta filters in the contract hierarchy, not below. Fixing the weakest of these contracts while shipping a commit message naming the strongest as still-broken seemed like the wrong trade. Say the word and I will split it back out.

Engrams **already in the result set are deliberately not re-checked.** Every path into `result.Activations` goes through one of phase 6's four gated scoring loops (RRF, CGDN, ACT-R, legacy), so they have already passed these predicates and re-checking would be redundant.

### Tests

All three fail without the fix:

- `TestEntityBoost_InjectedResultsRespectMetaFilters` — `tags_all`. Also asserts a filter-*satisfying* neighbour is still injected and the seed survives, so a fix that merely suppresses injection would not pass.
- `TestEntityBoost_InjectedResultsRespectTimeFilters` — `created_after`, the form originally observed on a live daemon, where a recall bounded to the current day returned entity-linked engrams from two weeks earlier scored purely by entity boost with no content score.
- `TestEntityBoost_InjectedResultsRespectExcludeUntrusted` — trust, plus the `TrustUnset` pin and a with-the-posture-off case.

`go build ./...`, `go vet ./internal/...`, `gofmt -l` clean. `./internal/engine/...`, `./internal/mcp/...`, `./internal/query/...` green, plus `-race` on the entity-boost tests.

### Known residuals — still bypassed, named rather than fixed

Phase 6 applies two further exclusions the injection path still ignores, both pre-existing and neither introduced here:

1. **`StructuredFilter`** — MQL `WHERE` clauses arrive via `ActivateWithStructuredFilter` and are applied inside phase 6 only, so injected engrams bypass the predicate entirely. Needs the structured predicate threaded the same way the meta filters now are.
2. **Foreign live-lease hiding (#548)** — leased-out engrams are injected despite the visibility contract. Needs `CallerOwner` and `IncludeLeased` threaded plus a per-candidate lease read (an extra Pebble read in the hub fan-out, or batching). #576 records the lease as advisory under a cooperative trust model, which is the severity argument for deferring it.

Separately, injected entries carry a flat `entityBoostFactor` and so are not subject to the caller's `threshold`. Mechanically that is the same class, with two differences worth stating: the violation is **visible** in the reported score, unlike a tag or trust violation; and enforcing it is a product decision rather than a patch, since injections score `entityBoostFactor` by construction and enforcing a caller threshold above it would disable injection entirely for those callers.

Happy to take any of the three in a follow-up.

### Relationship to #569 / #570

Distinct. #570 governs *how much* gets injected — rarity weighting, accumulation cap, threshold gate. This is about injected rows escaping the exclusion gates regardless of how many there are. The bypass reproduces on a daemon already running #570, and the diffs touch different lines; whichever lands first, the other should rebase cleanly.

### A structural alternative

Threading exclusions into a post-pipeline pass means there are now two places that must stay in sync with "what phase 6 excludes" — and the residual list above is evidence that this is a real maintenance hazard, not a hypothetical one. The `seedTagCandidates` comment states the intent that phase 6 is the single correctness gate, and moving the entity-boost pass inside `Run` would restore that property instead of adding a second gate that already lags by two exclusions. Substantially larger change, not attempted here — but if you would rather have the invariant than the patch, say so and I will look at it.

### DIFF
--- a/internal/engine/activation/engine.go
+++ b/internal/engine/activation/engine.go
@@ -1910,6 +1910,18 @@ func computeGatedActivation(vectorScore, normalizedFTS, decayFactor, hebbianBoos
 	return contentRelevance * gate
 }
 
+// PassesMetaFilter evaluates the request's filter predicates against a full
+// engram, for callers outside this package.
+//
+// Phase 6 is the correctness gate for everything the pipeline itself retrieves,
+// but passes that run after Run returns — the entity-boost spread activation in
+// particular — can introduce engrams the pipeline never saw. Those candidates
+// have never been filtered, so they must be checked here before being surfaced
+// (issue #654).
+func PassesMetaFilter(eng *storage.Engram, filters []Filter) bool {
+	return passesMetaFilter(eng, filters)
+}
+
 // passesMetaFilter evaluates filter predicates against a full engram.
 // Accepts *storage.Engram directly — avoids a separate GetMetadata call in phase6.
 func passesMetaFilter(eng *storage.Engram, filters []Filter) bool {
--- a/internal/engine/engine.go
+++ b/internal/engine/engine.go
@@ -2133,7 +2133,7 @@ func (e *Engine) activateCore(ctx context.Context, req *mbp.ActivateRequest, str
 	// After BFS produces a scored set, any engram sharing a named entity with a
 	// top-N result receives a small boost. This surfaces entity-linked engrams
 	// that have no direct association edge to the query-matching engrams.
-	result.Activations = e.applyEntityBoost(ctx, wsPrefix, result.Activations)
+	result.Activations = e.applyEntityBoost(ctx, wsPrefix, result.Activations, actReq.Filters, actReq.ExcludeUntrusted)
 	// Re-apply MaxResults: entity boost may have appended engrams beyond the limit.
 	// applyEntityBoost re-sorts by score descending, so truncation preserves top-K.
 	if actReq.MaxResults > 0 && len(result.Activations) > actReq.MaxResults {
--- a/internal/engine/engine_entity_boost.go
+++ b/internal/engine/engine_entity_boost.go
@@ -50,6 +50,13 @@ func (e *Engine) applyEntityBoost(ctx context.Context, ws [8]byte, results []act
 		seenInResults[r.Engram.ID] = i
 	}
 
+	// Candidates rejected once stay rejected: the fetch and every exclusion
+	// check depend only on the candidate, never on which seed reached it. A
+	// candidate sharing entities with several seeds would otherwise be
+	// re-fetched and re-checked per seed — common under restrictive filters,
+	// where most of a dense entity neighbourhood is rejected.
+	rejected := make(map[storage.ULID]struct{})
+
 	// For each seed engram, iterate its entity links (0x20 forward index).
 	for _, topEng := range seeds {
 		_ = e.store.ScanEngramEntities(ctx, ws, topEng.Engram.ID, func(entityName string) error {
@@ -67,12 +74,17 @@ func (e *Engine) applyEntityBoost(ctx context.Context, ws [8]byte, results []act
 					// Boost existing result.
 					results[idx].Score += entityBoostFactor
 				} else {
+					if _, wasRejected := rejected[engramID]; wasRejected {
+						return nil
+					}
 					// Fetch engram and add as a new entity-boosted result.
 					eng, err := e.store.GetEngram(ctx, ws, engramID)
 					if err != nil || eng == nil {
+						rejected[engramID] = struct{}{}
 						return nil
 					}
 					if eng.State == storage.StateSoftDeleted || eng.State == storage.StateArchived {
+						rejected[engramID] = struct{}{}
 						return nil
 					}
 					// This engram was never in the pipeline's candidate pool,
@@ -84,9 +96,11 @@ func (e *Engine) applyEntityBoost(ctx context.Context, ws [8]byte, results []act
 					// backward-compat alias for TrustInferred and passes
 					// through, as it does in the pipeline.
 					if excludeUntrusted && eng.Trust == storage.TrustUntrusted {
+						rejected[engramID] = struct{}{}
 						return nil
 					}
 					if !activation.PassesMetaFilter(eng, filters) {
+						rejected[engramID] = struct{}{}
 						return nil
 					}
 					results = append(results, activation.ScoredEngram{
--- a/internal/engine/engine_entity_boost.go
+++ b/internal/engine/engine_entity_boost.go
@@ -26,7 +26,13 @@ const (
 // the 0x23 reverse index. Each such engram receives a score boost of
 // entityBoostFactor (or is added to the result set with that score if it was
 // not already returned by BFS). Results are re-sorted by score descending.
-func (e *Engine) applyEntityBoost(ctx context.Context, ws [8]byte, results []activation.ScoredEngram) []activation.ScoredEngram {
+//
+// filters and excludeUntrusted carry the request's exclusion rules. They must be
+// applied to engrams injected here, because this pass runs after activation.Run
+// has returned and therefore after phase 6, the pipeline's correctness gate.
+// Engrams already in results need no re-check — they passed phase 6 on the way
+// in — so only the injection path consults them (issue #654).
+func (e *Engine) applyEntityBoost(ctx context.Context, ws [8]byte, results []activation.ScoredEngram, filters []activation.Filter, excludeUntrusted bool) []activation.ScoredEngram {
 	if len(results) == 0 {
 		return results
 	}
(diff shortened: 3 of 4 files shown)


isaac-ranger:
The three residuals named above are now tracked as #657, so they are not left named-but-unfiled.

scrypster (changes requested):
Nice work, and the delegation to `passesMetaFilter` instead of a copied predicate is the right call. That is the thing that stops the two paths drifting apart in six months.

I verified the claims rather than taking the description. The trust check does mirror phase 6 exactly, `TrustUnset` passes on both paths, and the ordering matches so a `state` filter behaves the same either way. COG-18 is fine: the boost factor and TopN are untouched and this can only remove candidates, so the flooding bound from #581 gets tighter rather than looser. Reverting just the two new guards turns all three tests red, and they assert the satisfying neighbour still gets injected, so they are load-bearing in both directions.

One thing blocks merge. All three tests call `applyEntityBoost` directly with filters they build themselves. That proves the function filters correctly, but not that `activateCore` actually passes the request's filters into it, and that second half is what fixes the daemon. I checked how much that matters by un-wiring the call site to `nil, false` and running the whole engine package. It stayed green. So a later refactor of `activateCore` can silently reopen #654 and CI will say nothing.

1. Add one test that drives real `Activate` with `Filters: [{created_after, ...}]` and asserts the old entity-linked engram does not come back. `TestEntityBoost_MaxResultsRespectedAfterBoost` already has the shape to copy.
2. Trap on that test: the request-level filter is `mbp.Filter` and the conversion copies `Value` verbatim, so `created_after` needs a real `time.Time`, not an RFC3339 string, or the type assertion fails open and the test passes for the wrong reason. Same for `state`, which must be a `storage.LifecycleState`. Assert the positive case as well so a silently ignored filter cannot produce a false green.
3. Please keep the merge linked to `Fixes #654` only. I checked #657 and all three of its items are still live after this, exactly as you said in the body.

Separately, one thing worth folding in while you are here. Rejected candidates are not memoized, so each one gets re-fetched for every seed that shares an entity with it. That path used to be rare because only soft-deleted engrams hit it. Your change makes it the common case for any restrictive recall. A small rejected-set would bound it to one read each.

The structural argument in your body is right and I would rather see it as its own RFC than a fourth rule threaded through. This is your third fix in this area after #616 and #655 and the diagnosis keeps being better than the individual patch, which is a good sign for the RFC.

isaac-ranger:
Both asks landed in 042b796, plus the fold-in:

1. **The wiring test** — `TestEntityBoost_ActivateWiresRequestFiltersIntoBoost` drives real `Activate` with a `created_after` filter and asserts both directions: the out-of-window entity neighbour stays out, and the in-window neighbour still arrives via the boost path, so a silently ignored filter (or a boost pass that never ran) cannot produce a false green. Reproduced your probe before writing it: un-wiring the call site to `(nil, false)` left the package green; with the new test it fails exactly there and nowhere else.
2. **The trap** — the filter value is a real `time.Time` through the verbatim conversion, per your note; the positive assertions guard the fail-open side.
3. Merge link unchanged — `Fixes #654` only; #657's three residuals are untouched by this branch.

The rejected-set memo is folded in: rejection is candidate-only (fetch, state, trust, filter — none depend on the seed), so each rejected candidate now costs one read per pass instead of one per seed sharing an entity with it.

On the RFC: agreed, and I'd rather write it than thread a fourth rule through too — the entity-injection contract (what may enter results after phase 6, and under whose predicates) will come as its own RFC rather than piecemeal. It will fold in #657's residuals as the open questions.

scrypster:
This holds up on the merits, Cairn. We RED-proved it twice (neuter the two exclusion checks and the three filter/trust tests go red, neuter the wiring and the Activate-request test goes red), and it delegates to phase 6's own `passesMetaFilter` so it can't drift from what phase 6 actually applied.

Two logistics rather than code changes. We'd like #570 to land first, since both of you rework `applyEntityBoost` and it's the larger foundational change, so we're holding this one for a small rebase on top rather than merging now and forcing #570 to rebase against you. And reflex just merged as #685, which changed `applyEntityBoost`'s signature (added a threshold param), so the rebase will touch that line too. Both are small and nothing changes in the logic here. Your #657 already tracks the StructuredFilter and lease residuals, so those are covered.

isaac-ranger:
Status note for anyone reading the flags: this PR is in a deliberate holding pattern, not awaiting work. Per the review above, #570 lands first (both PRs rework `applyEntityBoost` and it's the larger change), then this branch gets its small rebase on top — the conflict badge is that sequencing made visible, and the changes-requested status predates the comment accepting the rework on the merits. #570's round-3 push now carries the full injection contract (filters, trust, lease, validity), so the rebase here should shrink further: this PR's two gates delegate to the same predicates that branch already threads through. Rebase lands promptly once #570 merges.

scrypster:
Closing this since #570 folds it in and takes it further. The meta-filter and trust exclusion on injected results that you added here are in #570, and there they also compose with the validity and supersession gates from the stack that just landed. Thanks for splitting the work out clearly enough that it was easy to consolidate.

PR #659, opened 2026-07-27
feat: UPSERT write mode for Write/BatchWrite (#556)

### LINKED ISSUE #556: feat(grpc,engine): UPSERT write mode for Write and BatchWrite
## Summary

`Write` and `BatchWrite` are CREATE-only today. There is no way to update an existing engram without resetting its cognitive state. This is a blocking gap for any bulk ingestion workflow that needs crash-safe restarts or incremental re-sync.

## Context

Raised via community integration work (go-rag bridge, https://github.com/madeinoz67/go-rag), but this is a general write-path gap that affects any high-throughput or re-ingestion use case.

## Current behavior

- Content-hash dedup exists: identical content returns the existing ID and bumps `AccessCount`
- `idempotent_id` field exists on `WriteRequest` (MBP type) but is **not wired through the engine in gRPC or REST paths** — only MCP (`handleRemember`) uses it. This is effectively dead code outside MCP and should be fixed as part of this work.
- There is no merge path: a caller cannot update `Concept`, `Content`, or `Tags` on an existing engram without a full re-write that creates a new ULID (via `Evolve`)

## Proposed change

Add `upsert_mode bool` (or `WriteMode enum {CREATE, UPSERT}`) to `WriteRequest` in both the MBP types and proto schema.

When `upsert_mode = true` and `idempotent_id` is set:
1. Check idempotency key (before content-hash dedup — order matters)
2. If found: `GetEngram` → merge → re-encode and write back
3. **Must preserve**: `CreatedAt`, `Confidence`, `Stability`, `AccessCount`, `LastAccess`
4. **May update**: `Concept`, `Content`, `Tags`, `TypeLabel`, `Embedding`, `Summary`
5. Update content-hash index if content changed (stale hash → new hash)
6. Associations are untouched automatically (separate `0x0A` key space)

Default `upsert_mode = false` preserves all existing behavior.

## Files

- `internal/transport/mbp/types.go` — add field to `WriteRequest`
- `proto/muninn/v1/service.proto` — add field to `WriteRequest`
- `internal/transport/grpc/engine_adapter.go` — propagate field
- `internal/engine/engine.go` — `Write()` and `WriteBatch()` merge logic, wire `IdempotentID` check

## Notes

- The idempotency check must run **before** the content-hash check in `Write()`. Content may change on upsert — the old hash mapping needs to be updated, not treated as a dedup hit.
- Intra-batch dedup limitation (noted in existing code comment at line ~1299) should be documented in the PR.

madeinoz67:
## Summary

Implements `upsert_mode` for `Write` / `BatchWrite` — a crash-safe, incremental re-ingest that pins an engram to `sha256(idempotent_id)` on a single ULID (create on first use, merge-in-place on subsequent writes). Closes #556. Unblocks content-addressed sync for RAG-bridge consumers.

Shipped as three reviewable increments + two adversarial review passes (32-agent RedTeam + the project's `code-reviewer` agent) whose findings are all addressed + two merges with the latest `develop` (prefix collisions resolved).

- **Inc 1** `e0f5192` — keyspace: `UpsertKey = 0x2E` prefix + `UpsertKeyKey` (mirrors `ContentHashKey`). **Relocated twice** during review: `0x2B→0x2D` (upstream #681/#570 took 0x2B/0x2C) → `0x2D→0x2E` (upstream #694 took 0x2D for ProspectiveIntent).
- **Inc 2** `65ff6ac` — storage `UpsertEngram` (compound batch: engram + content-hash + upsert-key co-committed) + engine `writeUpsert` + `upsertKeyLock` + `WriteBatch` routing.
- **Inc 3** `4187e02` — transports: proto `bool upsert_mode = 12`, gRPC adapter, MCP `muninn_remember`, CLI `--upsert-mode` (REST is free via the mbp alias).
- **Parity tests** `a933618` / `a74488e` — MCP + gRPC cross-surface.
- **RedTeam fixes** `3f34e74` / `9b68a22` / `46c692b` — 3 criticals + 4 highs (see below).
- **Merge `develop`** `2f4ec65` / `86f2d53` — resolved two prefix collisions; all CI checks green.
- **Code-reviewer fixes** `a73309c` — Trust gating, Importance, OpenAPI, export/clone lists, stale comment cleanup (see below).

## Changes

- `Write`/`BatchWrite` gain `upsert_mode` (default `false` — fully backward-compatible; the default path is untouched, pinned by `TestWrite_UpsertMode_DefaultUnchanged`).
- New Pebble prefix `UpsertKey = 0x2E` + `UpsertKeyKey` (registered, scope-tested, in `clearVaultDataPrefixes` + `vaultScopedExportPrefixes` + `vaultScopedSwapPrefixes`).
- `storage.UpsertEngram`: create-on-miss / merge-in-place-on-hit / recreate-on-stale-pointer (soft-deleted, hard-deleted, or ClearVault-orphaned). Holds `casLocks.For(id)` across read→commit (invariant STO-2) and co-commits engram + `0x28` content-hash + `0x2E` upsert-key in one Pebble batch.
- `engine.writeUpsert`: acquires `upsertKeyLock` + `contentHashLock`, calls `resolveTrust` + `importanceFromRequest` (mirrors the default Write path), calls `UpsertEngram`, runs the search-critical hooks (HNSW/FTS/triggers), sweeps stale FTS postings on a content-change merge (`fts.DeleteEngram`, mirroring `Forget`).
- Cross-surface: proto/gRPC/REST(alias)/MCP/CLI all thread `upsert_mode`. OpenAPI spec updated.

## Flagged for maintainer review

1. **The `0x2E` durable forward index.** The issue body says "use `idempotent_id` as the key"; we do, but back it with a durable per-vault index because `0x19` is a TTL-purged receipt and would lose the mapping. Originally allocated at `0x2B`; relocated to `0x2E` after upstream `#681`/`#570`/`#694` allocated `0x2B`/`0x2C`/`0x2D` during review.
2. **Single-writer scope** is stated in the docstring (server-enforced dedup on the key string; no per-caller namespacing), not enforced. Cross-tenant key collision is possible if two callers share an `idempotent_id` — documented as the trust model.
3. **`AccessCount` preserved-not-bumped** under upsert (differs from content-hash dedup) — confirm this matches your intent.
4. **RESET vs ACCUMULATE.** This PR implements **ACCUMULATE** (preserve cognitive state on merge) per the issue body. A prior 32-agent RedTeam hardened the opposite (RESET, 23/32: re-promotion is transport replay, not a cognitive access, so preserving `AccessCount`/Hebbian/Ebbinghaus can forge the learning signal). Contributor posture: built to spec, surfacing the concern here at its strongest. If you prefer RESET, the change is localized to the merge path.

## Adversarial review — two passes, all findings addressed

**Pass 1: 32-agent RedTeam** — found 3 criticals + 4 highs:
- **C1** `upsertMerge` didn't hold `casLocks.For(id)` → #594 resurrection/lost-update race. Fixed + RED-sanity-tested.
- **C2** StateActive guard was dead code for hard-delete → re-ingest errored forever. Fixed (`errors.Is(err, ErrNotFound)` → recreate).
- **C3** `0x2E` missing from `clearVaultDataPrefixes` → ClearVault orphaned entries. Added.
- **H** MemoryType dropped on merge; `replicateBatch` skipped; content-hash lock mismatch; FTS/trigram stale on merge. All fixed.

**Pass 2: project `code-reviewer` agent** — found 5 additional findings:
- **Trust**: `writeUpsert` hardcoded `TrustInferred`, bypassing `resolveTrust` (S8 verified-credential gate). Fixed — now calls `resolveTrust`.
- **Importance**: `writeUpsert` omitted `Importance` on create (silent 0, missed COG-20 prune floor). Fixed — now calls `importanceFromRequest`.
- **OpenAPI**: `openapi.yaml` missing `upsert_mode` (REST alias silently propagated it). Fixed.
- **Export/clone lists**: `vaultScopedExportPrefixes`/`vaultScopedSwapPrefixes` missing `UpsertKey`. Fixed.
- **Stale comments**: pervasive `0x2B`/`0x2D` references updated to `0x2E`.

## Release Checklist

- [x] ~~`CHANGELOG.md`~~ — maintainers
- [x] OpenAPI spec (`internal/transport/rest/openapi.yaml`) — **updated** (`upsert_mode` added)
- [ ] SDK types (Python/Node/PHP) — **not updated** (Inc 4 follow-up; the field is additive + optional, so existing SDKs keep working — confirmed by the CI Node + Python SDK integration tests passing)
- [x] `docs/internals/keyspace-registry.md` — updated (`0x2E` row + free-bytes to `0x2F+`)
- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./...` passes (incl. `-race` on the upsert concurrency tests; full engine + storage suites green)
- [x] **CI 10/10 green** (Go build&test, Windows, Playwright E2E, CLI integration, Python SDK, Node SDK, Web console vitest, Vulnerability scan, API spec validation, Shellcheck)

The v1 hook scope (`writeUpsert`) runs HNSW/FTS/triggers/counters on create + FTS-sweep on merge, but skips the caller-enrichment/analytics hooks (inline entities, contradiction worker, novelty, auto-association) — upsert targets content-addressed re-ingest, which doesn't carry those. Documented; full hook parity is a follow-up if a caller needs it.


scrypster (changes requested):
Thank you for the patience on this one. Four keyspace relocations through no fault of yours is real friction, and we know it. Before anything else: the engineering here held up under a hard look. We removed your casLocks serialization and watched the lost-update test fail with 71 of 200 engrams resurrected, so that fix is genuinely load-bearing rather than defensive. The compound single-batch commit, the transport threading, the prefix-list registrations and the parity tests are all careful work, and the majority of this PR carries over to what we are about to suggest.

We put the design question to two independent reviewers with instructions to argue both sides hard. They came back with the same answer, having rejected both options we were choosing between, and we think they are right.

The requirement your bridge actually has is stable external addressability, not a frozen ULID. Frozen-ULID is the one implementation of that requirement that forecloses history. So: keep the 0x2E index, but let it map the key to the CURRENT HEAD of that key's chain rather than to a fixed engram, and re-point it on each evolve. Write path becomes: key miss means Write and pin, key hit means load the head and compare content hash, identical means no-op, different means EvolveAt and re-point. The key is the stable identity, the ULID is free to move, and your bridge never has to hold a ULID at all if we add a read-by-key resolution surface, which we would like.

That gives you every declared edge, every as_of destination, every provenance record and the version-chain redirect for free, and none of the roughly thirty cognition invariants gains an "unless this is a projection" clause. It also deletes the merge function rather than fixing it, which matters because of what we found.

Three live findings in the merge path, given as evidence rather than as a scolding:

1. Link with relation supersedes stamps ValidUntil without soft-deleting, so your State != StateActive guard does not catch it. Tonight's resync then merges fresh document text into an engram whose validity window is closed, the write returns success, and default recall can never return that content again. Silent, permanent, and reachable with two ordinary tool calls.
2. A content-changing merge without a caller-supplied embedding leaves the old vector in HNSW under the same ULID. FTS gets a stale-posting sweep, the vector index does not, so retrieval runs on last week's evidence against this week's text.
3. CreatedAt is preserved by design, and RetentionDays prunes by CreatedAt. So a document refreshed last night gets hard-deleted for being ninety days old, after which the next sync mints a new ULID and the stable-ULID property the feature exists to provide is gone anyway.

Worth saying plainly why we are changing the shape rather than asking for a fourth round of fixes: three adversarial passes have now produced thirteen findings on this path, and they are all the same bug. A path that mutates an existing engram in place has to hand-re-derive every invariant the create path establishes, and each pass finds a different one it forgot. That is the shape talking, not you.

Three of your four flagged questions dissolve under this design. RESET versus ACCUMULATE stops being ill-posed, because the successor is a distinct engram and EvolveAt already defines inheritance, which means your red team's instinct that re-promotion is transport replay rather than a cognitive access was right. AccessCount preserved-not-bumped becomes moot for the same reason. And a cross-tenant key collision degrades from a silent content overwrite to a visible duplicated chain. Your fourth question, whether the durable 0x2E forward index was the right call over the TTL-purged receipt index, was right and stays right. It just points at the head.

Traps to pre-scope so you do not find them in review: re-point 0x2E inside the same batch as the evolve, or a crash strands the key on the predecessor. Chain intermediates stay in HNSW and FTS, so measure index growth at your bridge's real churn rate and treat RetentionDays as the bound. And a high-churn document will exceed the chain depth cap, which is survivable because the head is directly retrievable, but the walk currently reports no-current-version instead of truncated in that case. That last one is our bug, not yours, and we are filing it separately.

One genuine question, and it does not gate anything: does the bridge want the documents in MuninnDB, or the claims derived from them? If it is the raw documents, we would rather tell you now that the write path may be the wrong home for them than let you build further on it. Your answer shapes the docs rather than the decision, since key-to-head is the right answer either way.

Your five merged PRs have all been in this ingest lane and the use case is real. We want this in.

madeinoz67:
Thank you for the careful review — the "that is the shape talking, not you" framing was the right call, and the evolve-based design is cleaner than anything the merge path could have been patched into.

**Implemented as requested.** The merge function is deleted entirely. `writeUpsert` is now an orchestrator:

- **Key miss / stale pointer** → delegates to `e.Write` (default create path — all hooks, enrichment, resolveTrust, importanceFromRequest) + `PutUpsertKey`.
- **Key hit, identical content** → strict no-op (`upsert-identical`).
- **Key hit, changed content** → `EvolveAt` + re-point `0x2F` (the actual byte; see below) to the successor in the **same atomic batch**.

The `0x2F` index maps the key to the **current head** of the chain, re-pointed on each evolve. A read-by-key resolution surface would let the bridge resolve without holding a ULID — I'd welcome that as a follow-up.

**The three findings are structurally gone** — confirmed by the verify pass:

1. `supersedes` / `ValidUntil` → the head's `State != Active` (superseded by `EvolveAt`'s `SupersedeEngram`) triggers the stale-pointer recreate. No merge into a closed-validity engram is possible.
2. HNSW stale vector → the successor gets a fresh ULID + fresh HNSW insert (or the retroactive processor embeds it). The predecessor is soft-deleted and hidden from recall.
3. `CreatedAt` / `RetentionDays` → the successor gets `CreatedAt: now`. A refreshed document can't be pruned for being old.

**The atomicity trap** — `StoreBatch.RepointUpsertKey` queues the `0x2F` Set into `EvolveAt`'s batch, called before `Commit()`. The re-point commits or rolls back with the successor.

**Prefix note:** the index is at `0x2F`, not `0x2E` — upstream landed `AssocWeightRepairMark` (`0x2E`, #756) since your review. Your "keep the 0x2E index" is honored as the design intent; the const comment documents the relocation history (`0x2B → 0x2D → 0x2E → 0x2F`, each from an upstream collision caught by the registry's disjointness test).

**Residual:** `EvolveAt` sets `Trust = TrustInferred` on the successor. A content-change upsert of a `verified` memory drops to `inferred` — the create branch doesn't have this (it goes through `e.Write → resolveTrust`). If you want `verified` preserved across evolves, that's an `EvolveAt` change, not a `writeUpsert` change. I left it as-is since it matches the existing `EvolveAt` behavior.

**Your question — documents or derived claims?**

Documents. The go-rag bridge chunks source documents and stores the chunk text + a pre-computed embedding for RAG retrieval. MuninnDB's enrichment pipeline extracts entities/claims from those chunks; the bridge doesn't feed derived claims — it feeds raw chunk text and lets MuninnDB's cognition do the extraction. The write path is the right home: the bridge needs server-enforced keyed dedup (so a re-sync of an unchanged document is a no-op, and a changed document replaces the old version), which is exactly what this PR provides.

scrypster (approved):
Redesign received, and it's the right one. Merging.

You turned a full architectural rewrite around in under a day, deleted your own 193-line compound rather than defending it, and the result is smaller than what it replaced. That is not a common thing to watch someone do.

**Before the substance: the week of silence before our review was ours, not yours.** You opened this on 27 July with three clean increments and two adversarial passes already run, and it sat. Then five keyspace relocations, because `0x2B`, `0x2C`, `0x2D` and `0x2E` all got taken out from under you while you waited. Make that six — `0x2F` went to the replication keyspace in #726, so the byte is now **`0x30`**. We relocated it ourselves in the merge rather than asking you to do it a fifth time.

## What decided the PR

The supersession question. #856 landed yesterday and made it law that every content-replacing operation must **declare** supersession — a `RelSupersedes` edge plus a closed `ValidUntil` — because consolidation was the one verb that didn't, and it left a permanent recall hole.

**Your v1 merge path would have had exactly that defect.** The evolve-based redesign has it structurally right, and we confirmed it rather than reading it: predecessor `ValidUntil` closed, declared edge present, successor Active. Your test pinned the edge but not the closed stamp, so we added that assertion — COG-28 reads the pair, and half a signature isn't a pin.

We reverted your two load-bearing mechanisms to check they carry weight. Without `RepointUpsertKey` inside the evolve batch the pointer strands on the predecessor; without the `upsertKeyLock` stripe, 19 of 20 concurrent upserts orphan off the chain under `-race`. Both real, both discriminating.

A full adversarial pass ran against the redesign — 10-hop chains, 24 concurrent same-key upserts, every stale-pointer shape including cross-vault and zero-ULID, clone, ClearVault, corrupt pointer values, append-mode across every transport. **The atomicity story held under all of it.**

## Three things we found and fixed on our side

**A dedup collision, and it was going to bite you quietly.** The create branch delegates to `Write`, which runs content-hash dedup — but upsert identity is the *key*, never the content. Two **different** upsert keys whose chunks carried byte-identical text collapsed onto one engram and then shared fate: revising document B soft-deleted document A out from under its own key, and the predecessor carried a false claim that A had been superseded by B's revision. Duplicate chunk text across documents — boilerplate, licence blocks, repeated headings — is ordinary in a RAG corpus.

**And the one I'd most want you to know about**, because it's the subtler half of the same idea: `muninn_forget(id, not_true_since=...)` is valid-time invalidation. It deliberately leaves `State=Active` and stamps `ValidUntil`. `writeUpsert` was checking `State != Active` alone, so an invalidated head hit the identical-content branch — and the document **could never return to recall unless its text happened to change.** `muninn_guide` actively recommends `not_true_since` for exactly this case, so a sync tool would have re-uploaded the unchanged document forever and been told `upsert-identical` every time. The sibling path at `engine.go:1401` already got this right; upsert now matches it.

Both fixed with RED-verified pins, plus one comment correction: the create-path crash window self-heals the *pointer*, but leaves an orphaned Active duplicate nothing will ever supersede. Your comment said it self-heals; it heals half.

## Two honest limitations, neither blocking

On the evolve branch, tags and confidence come from the previous version and the ones you send are dropped — that's `EvolveAt`'s inheritance model, and #730 recently made evolve *reject* tags outright in favour of `muninn_update_tags`. Documented rather than changed; use `muninn_update_tags` if the bridge needs to retag. Your note about `EvolveAt` forcing `TrustInferred` is correct and you were right to leave it alone — it affects every evolve, not just yours.

We also deferred the gRPC `hint` field: `WriteResponse` has none at all today, and adding it needs a proto regen we couldn't do safely here.

## One thing for next time, and it's on us for not making it visible

Your three upstream merges resolved `CLAUDE.md` by taking your branch's side, which reverted about 85 lines. Among them is a rule you've never had a chance to read: **this repo is public, and measurement corpora are referred to as "a production vault" — numbers kept, name dropped.** A vault name reached develop once, and a public repo's history is public forever. Nothing in your PR trips it; we scanned. You simply wouldn't have known it existed.

Separately, an adversarial pass on your branch surfaced a data race in `pebbleStoreBatch.mutateEngram` that reproduces on clean develop with no upsert involved — filed as #858. Not yours, but worth knowing: your work makes automated same-head evolves routine, which raises its hit rate.

Your five merged PRs have all been in this lane and you keep landing the hard ones. The read-by-key resolution surface is next, so the bridge never has to hold a ULID at all — we're filing it, and it's yours if you want it.

Thanks for the redesign, and for the flagged-questions section. Building to spec while arguing the other side at its strongest is the right way to hand a maintainer a decision.

PR #680, opened 2026-07-27
feat(mcp,engine): optional inline entities on evolve, replacing the carried set

isaac-ranger:
Second piece of the #646 split (stacked on #646 — contains its commit; the diff collapses to one commit once #646 merges). Additive feature, no bug fix.

`Evolve` gains an optional `entities` parameter with remember's shape and normalization (20-cap, NFKC trim, `normalizeEntityType`, empty-name drop), surfaced on `muninn_evolve`. When present it REPLACES the links otherwise carried forward from the predecessor — the update changed what the memory is about — and takes remember's inline path: record upsert (which funds the mention count), links, co-occurrence and `co_occurs_with` relationship records, since these are genuinely new mentions. REST and gRPC continue to pass nil (carry-only); surfacing the parameter there is a follow-up.

Per the review, `TestEvolve_InlineEntitiesReplaceCarry` no longer passes with the carry disabled: it proves both arms on the same predecessor shape. A control evolve without inline entities asserts the links carry forward — with the carry neutralized this leg fails, so the replace assertions cannot pass vacuously — and the inline evolve asserts the replacement set only, not a merge. RED-verified in that configuration.

No new MCP tool, so the registry-parity list and mutating/read-only classification are unchanged; the `cmd/muninn` integration smoke passes. `go vet -tags localassets ./...` clean, full `./internal/...` green, mcp under `-race`.


scrypster (commented):
Nice, this one's basically ready. We ran engine and mcp with the new test and it's green, and we checked the important thing by hand instead of trusting the description: your evolve inline-entity path is line for line the same as remember's caller-entity path, same 20-cap, NFKC trim, normalizeEntityType, same UpsertEntityRecord and co-occurrence funding. So the parity you claim actually holds. TestEvolve_InlineEntitiesReplaceCarry is nicely built too, the control arm asserts the carry is live before the replace arm asserts the swap, so it can't pass vacuously.

One required change before we merge:

1. The new `entities.type` in the muninn_evolve schema hand-writes a 7-type description instead of using `entityTypeEnum`. Every other entity-type field in tools.go (remember, remember_batch, correct_entity) uses `"enum": entityTypeEnum`. The normalization is correct so it's a schema-advertising bug not a data one, but a schema-driven client would be told 7 valid types (technology, project, tool, database, service, framework, language) don't exist. `entityTypeEnum` is already in scope at the top of allToolDefinitions(), so it's a one-line swap.

Two non-blocking notes for later: evolve's response has no Hint field so it can't report malformed-entity counts like remember does (just an asymmetry), and `entities: []` behaves the same as omitting it, so there's no way to explicitly clear an evolved memory's entities right now. If that's intentional a line in the description would save someone a puzzle. Stacked on #646 which just merged, so it'll rebase clean. Send the enum fix and we'll land it.

scrypster:
Heads up: reflex (S0-S8) just merged as #685. It added `trust`, `read_only`, and `exclude_type_labels` params to the same `handlers.go` and `tools.go`, different tool blocks than your evolve `entities`, so you'll get a textual conflict on those two files when you rebase. It's keep-both, not a redesign. The `entityTypeEnum` one-liner from the review above is still the only change we need before this lands.

scrypster (approved):
Nice, this is a clean feature. The guard so inline entities suppress the carry instead of doubling it is exactly right, and it composes with the valid-time stamping and the append guard on evolve without any of them stepping on each other.

Heads up, the stack moved under you a couple of times while this sat (the reflex stack, then the importance dimension, both touched the evolve signature), so rather than bounce it back we resolved it onto current develop ourselves. The combined EvolveAt now carries entities, importance, and effective_at together, and we kept your entities-replace-carry intact. We checked the thing that worried us most, that the empty-entities path still carries the predecessor entities and relationships exactly like #646 does, and it does, byte for byte. Both the replace path and the importance override redden a test when disabled, so neither is a vacuous pass.

Ran the full suite with -race on the evolve and ledger path, clean. Merging once CI is green. Thanks for keeping the inline path a faithful mirror of what remember does.

PR #681, opened 2026-07-27
fix(engine,storage): startup repair for evolve-stripped successors — atomic, funded, watermarked

isaac-ranger:
Third piece of the #646 split (stacked on #646; Tier 3 by the rubric — a one-way backfill over existing user data — so this expects the independent second pass before merge). Addresses review items 2, 3, 5, 6 and 8.

One-shot startup pass healing the damage the pre-fix Evolve left on existing vaults: supersede-successors born with zero entity links even though their predecessor has them. Soft-delete preserves the predecessor's 0x20/0x23/0x21 entries, so links are copied forward rather than reconstructed, driven from the soft-deleted side via a new index-only `ScanEngramsByState`. Chains stripped at every hop heal by fixpoint rounds.

**Crash safety (item 2).** Each successor is repaired in ONE Pebble batch — links, relationship records, digest flag — in `storage.RepairEvolveEntityCarry`, using the batch methods #646 added. A kill mid-repair can never strand a partial link set for the has-any-link skip test to mistake for intact: a successor is repaired completely or not at all.

**Replication, decided deliberately (your trap).** The batch is NOT replicated, on purpose: the repair is a deterministic local backfill derived from data every replica already has, and every node running this build performs its own pass at startup. Replicating the writes would double-apply them on followers that also repair. Stated in the method comment.

**DeleteEngram race (item 3).** The engram's stripe lock (`casLocks.For(succID)`) is held across the existence re-check, the has-links re-check, and the batch commit. `DeleteEngram` removes the same key families under the same stripe, so a repair write cannot land after a hard delete. No lock inversion: nothing inside the critical section takes another stripe, and DeleteEngram releases the stripe before taking entity locks.

**Ledger funding.** Repaired links fund mention counts and co-occurrence post-commit, same conservation as #646's write-path carry (proven by the same write-1/carry-2/delete-1 sequence through the repair path). Funding failures warn and continue, matching Evolve's own funding: once links are committed the repair cannot be retried, so aborting the sweep would cost the rest of the vault without recovering the counts.

**Watermark + startup cost (item 5).** A vault that completes a CONVERGED pass — a round that found nothing left to repair — is watermarked (new 0x2B prefix: registry, collision guards, `ClearVault`, keyspace doc all updated) and skipped on later boots. Exhausting the round cap while rounds are still repairing does NOT watermark; the vault resumes on the next boot until it converges (`TestEvolveEntityLinkRepair_RoundCapExhaustionIsNotClean` pins this with a 40-hop descending-ULID chain — the one shape that makes the cap bind, since the 0x0B scan heals ascending chains in one round and cycles self-terminate via the has-links skip). The pass also waits out 60s + jitter at startup, matching `runPruneWorker`; `EngineConfig.EvolveRepairDelay` overrides for tests. Documented residual: a rollback to a pre-fix binary after a vault is watermarked would strand new damage; recovery is bumping `evolveRepairVersion`.

**Wiring test (item 6).** `TestEvolveEntityLinkRepair_EngineLifecycleAndWatermark` drives real `NewEngine`/`Stop` across three boots (damage → heal+mark → skip). Deleting the spawn line turns it red while the direct-call tests stay green — RED-verified in exactly that configuration.

**Log honesty (item 8).** The sweep log is unconditional — a zero is a signed claim, not silence — and carries `clean`; the limitation (repaired counts successors healed from zero links; successors with any existing link are assumed intact and skipped) is documented at the log site and above.

`go vet -tags localassets ./...` clean, full `./internal/...` green, storage and engine under `-race`.


scrypster (changes requested):
This is a lot of careful engineering, Cairn, and it mostly holds. We gave it the full Tier-3 treatment, a primary pass plus an independent adversarial pass whose whole job was to make it corrupt user data. It couldn't. Watermark tear, ledger double-count, repair-versus-delete races, keyspace resurrection on vault reuse, malformed records, all defended, and both idempotency and interrupted-then-resume are RED-proven. The 0x2B watermark is in prefix.All() and clearVaultDataPrefixes, so a cleared vault won't resurrect a stale mark.

The adversarial pass found one real residual, and it's a concurrency one your own docstring rules out:

1. The atomicity comment says the stripe lock means a repair can't land after a concurrent writer. True for casLocks-respecting writers (DeleteEngram, CompareAndSet). Not true for ReplayEnrichment, which writes entity links and bumps MentionCount with no casLock, targeting exactly the StateActive, digest-unset, stripped successors the repair targets. An operator upgrading to fix #622 might also run replay-enrichment inside the repair window, and the two double-fund the ledgers. It's accounting drift not content loss (and self-limiting, higher counts get boosted less), so not a corruption blocker, but the docstring has to stop claiming it's impossible. Cheapest close: you already read the digest flags at the top of RepairEvolveEntityCarry, so skip when DigestEntities is set. A TOCTOU sliver remains so soften the comment too. Don't reach for a coarser lock, the asymmetry is that ReplayEnrichment takes no casLock at all.
2. The "every replica already has this data" comment rests on an assumption that isn't independently true, pebbleStoreBatch.Commit() never replicates, so Evolve's writes don't reach followers via the incremental log (same gap we noted on #646). Soften it, and let's track the batch replication gap as its own issue in the #596 class.

Everything else checked out under real execution. Make those two docstrings honest and add the digest-flag skip, and it's ready. Genuinely strong work on a hard problem, this is the kind of migration most people get subtly wrong.

scrypster:
Good news on the keyspace side, Cairn: reflex just merged as #685 and we deliberately took 0x2C for its tag-range index, leaving 0x2B free for your evolve-repair watermark. So there's no collision to sort out, the byte is yours. You'll just want a rebase on develop for the textual overlap on engine.go and storage. The DigestEntities skip and the two docstring softenings from the review above are still what's needed.

scrypster (approved):
Okay, this is the kind of migration most people get subtly wrong, and you did not. We picked up the two things from the last round ourselves rather than send it back to you again: the digest-set skip so an enrichment pass and the repair cannot both fund the same successor, and the two docstrings that were overclaiming (the atomicity one versus ReplayEnrichment, and the replicas-already-have-it one). Both are in now.

We resolved it onto current develop for you too, since the reflex stack, the importance dimension, and the inline-entities work all landed under you and moved the same files. Then we ran a fresh independent adversarial pass whose only job was to make the repair corrupt or skip real data, and it could not. The digest skip does not mask a real repair (a genuinely stripped successor has no digest entry, so raw reads zero), and it actually drops a false-positive carry onto entity-less successors as a bonus. The 0x2B watermark does not collide with S1 taking 0x2C, and a cleared vault drops the mark and re-scans. There was one duplicated carry loop the merge left behind, harmless but dead, so we deleted it.

We added a red-proven test for the skip so it cannot quietly rot. Merging once CI is green. Genuinely strong work, Cairn.

PR #687, opened 2026-07-28
feat(auth): append-only credential mode (read + create-new; no evolve/forget)

scrypster:
Adds an **append-only credential mode** (`auth.ModeAppend`): a key that can recall and remember/remember_batch but is refused any operation that modifies or deletes an existing engram (evolve, forget, trust, merge, link, ...).

## Why
Automated capture (the session-end "flush" reflex) writes to a vault driven by untrusted transcript content. A full or even write-mode key could evolve/forget existing memories. An append-only credential makes that structurally impossible: capture can add and read, never destroy or overwrite — the same posture SEC-7 gives cross-vault, now for destroy-in-vault. An adversarial review of the flush pipeline flagged exactly this: destroy-in-vault rested on a client-side tool allowlist, not the credential.

## Enforcement — two layers
- **MCP dispatch gate** (`isAdditiveTool`): append keys reach read tools + the four additive tools (`remember`, `remember_batch`, `remember_tree`, `add_child`) only; destructive mutating tools return `forbidden`.
- **Engine backstop** (`auth.AppendFromContext` in `Evolve`/`Forget` → `ErrAppendForbidden`): transport-agnostic, mirroring SEC-14's verified-trust gate, so the guarantee holds on every surface, not just MCP.

Accepted by `keys_store`/`capability_store` validation, the CLI (`api-key create --mode append`), and the REST admin key-create handler.

## Verification
- RED-proven: neuter the engine gate and Evolve-under-append succeeds; restore and it's refused.
- `-race` clean across auth/mcp/engine.
- End-to-end on a labs binary over real MCP: an append key **created** a memory, was **refused** `forget`/`evolve`/`trust` ("append-mode key cannot call this tool"), could `recall`, while a full key could `forget` (control).

Tier 3 (touches `internal/auth/`). Expects the independent adversarial pass before merge.


scrypster:
Closing: this landed already. The append-only credential mode shipped via `fb77813` (the reflex-stack integration, #688) — `auth.ModeAppend`, `AppendFromContext`, `Engine.refuseAppend`/`ErrAppendForbidden` across the destructive ops, the `isAdditiveTool` MCP dispatch gate, the CLI `--mode append` flag, and the REST/OpenAPI enum are all on develop. SEC-15 in `docs/internals/invariants.md` names this PR explicitly, and `guardedOps` now covers 31 ops — a superset of the 30 here, with Restore, Link and PruneVault added as the later-found stragglers. Nothing left to salvage; the PR just never got closed.

Issue #700, opened 2026-07-29
annotate:true fills SupersededBy/ConflictsWith from an ungated reverse-edge lookup, bypassing the caller's visibility contract


isaac-ranger:
Found by our adversarial review of the visibility-gate PR (supersession increment); pre-existing, not introduced there — but that PR's gate makes the composition visible.

**The path.** \`muninn_recall\` with \`annotate:true\` calls \`GetAnnotations\` per row (\`internal/mcp/handlers.go\` ~539-551). \`GetAnnotations\` (\`internal/engine/annotation.go\` ~51-62) returns the first reverse \`RelSupersedes\` edge raw: no state check, no lease visibility, no trust/meta/structured filters, no as-of view. \`augmentAnnotations\` (~2079) then fills \`SupersededBy\` from that lookup exactly when the ranking phase left it empty — which, with the gated supersession phase, is exactly when the gate refused to name the successor.

**Concrete leak.** Unstamped chain A←B, B under a live foreign lease, caller ≠ owner, \`annotate:true\`: the ranking phase abstains atomically (no injection, no demotion, no annotation), then the annotate fallback stamps lease-hidden B's ULID onto row A anyway — the #548 class in metadata form, one layer above the gate. Same composition surfaces a soft-deleted superseder's ID (deletion-means-deletion broken in metadata) and, under \`as_of\`, a view-future head's ID. \`ConflictsWith\` has the same ungated shape.

**Fix trap worth pre-naming.** The handler cannot distinguish "the gate abstained" from "the row sat below the \`MaxResults+supersessionMargin\` examine horizon and was never walked" — simply deleting the fallback silently un-annotates below-horizon rows. A real fix probably routes the annotate lookup through the same visibility gate (\`visibilityGate.Nameable\`) with the request in hand.

COG-21 (added in the gate PR) scopes itself around this surface and cites this issue as the tracked deferral.

PR #701, opened 2026-07-29
fix(engine): shared visibility gate for post-pipeline injections; supersession substitutes atomically under the caller's view

isaac-ranger:
Follow-through on the #570 rounds and the class they exposed: three review cycles found the same structural defect in three different fields (#654 meta filters; your Tier-3 trust pass; our lease+validity pass), because each post-pipeline injector carried its own fragment of the visibility contract. This PR factors the contract into one place and routes both injectors through it — and it closes the supersession-side holes that were still live.

## What it does

**`visibilityGate`** (new, `internal/engine/visibility_gate.go`): holds the resolved `ActivateRequest` and renders the caller's eligibility decision by delegating to phase 6's own predicates — six checks: state, meta filters, trust, structured (MQL WHERE) filter, foreign live lease (fail-closed on read error), valid-time. Two deliberate layers:
- **`Nameable`** — the existence half (state/filters/trust/structured/lease): an ID that fails it may not surface anywhere the supersession ranking phase emits, result rows and `SupersededBy`/`CurrentVersion` annotations alike. For these predicates the ID itself is the secret (#548).
- **`ValidForView`** — the valid-time half (COG-19): governs only what may be presented as *current*. Split because supersession stamps `ValidUntil` on every superseded engram — chain intermediates are expired **by construction** and must remain nameable as lineage while never counting as current.

**Entity boost** (pass 2b): pure extraction of the #570-round inline checks into `gate.Admits` — no behavior change except one: injections now also honor `StructuredFilter`, which phase 6 applies inside `Run()` before both injectors (an MQL `WHERE` could previously receive boost-injected rows violating its predicate). Pinned by `TestEntityBoost_InjectionRespectsStructuredFilter`.

**Supersession** (`applySupersession` + `resolveSupersessionHead`): previously took no request at all — injected heads bypassed filters, trust, lease, and as-of, while the post-hoc COG-19 cut produced the worst half-state (stale demoted and annotated with a head that was then dropped). Now the chain resolves **under the caller's view**: hidden nodes are traversable but unnameable; the effective head is the deepest nameable node valid for the view; the annotation names the nearest nameable, view-known successor; under `as_of`, a node postdating the instant may not even be named (pinned by a backdated-successor test). A chain with no such head **abstains whole** — no injection, no demotion, no annotation: demote-without-inject silently truncates the topic, and an annotation naming a hidden node is the #548 leak in metadata form. Admitted injections count in `TotalFound` (closing the follow-up scoped in the #570 review), one shared clock covers admission through the final COG-19 cut, and COG-21 in `docs/internals/invariants.md` pins the contract (COG-9/COG-18 anchors updated).

## Honest scoping

- **Default-view reach:** a `Link`-stamped predecessor is validity-expired and dies inside `Run()`, so in default view this phase serves **unstamped legacy chains** (every pre-valid-time store) — plus all `as_of`/`include_invalid` views. The differential `TotalFound` test documents this.
- **Deferred, filed as #700:** the `annotate:true` fallback (`GetAnnotations` → fill-when-empty) is a pre-existing ungated surface that can re-leak an ID the gate refused to name; fixing it needs the gate plumbed into the annotate path without silently un-annotating rows below the examine horizon. COG-21 cites it.
- **Known residuals, documented in-code:** boost runs on its own clock a few microseconds before `injectorNow` (overcount sliver, swept by the final cut); the truncated-then-reinjected double-count applies to both injectors; the depth cap truncates the walk rather than abstaining (pre-existing; doc now says so). Offered follow-up if you want an issue: the `Run()` backing-array aliasing with the async activation-log drainer, which all in-place post-pipeline phases share.

## Verification

Both substantive commits went through a full adversarial review pass against their final diff (tier 3 by your rubric — lease keywords); the second pass's two findings (ViewFuture unpinned; COG-21 overclaiming the annotate surface) are resolved in the third commit. Every guard is RED-verified against its bypassed check: five per-predicate abstention guards, the hidden-intermediate naming leak, the view-future naming leak, view-validity (forced true), the as-of pair, and the `TotalFound` call-site wire (fails 3/3 with the `+=` dropped, drift-controlled against score movement between runs). `go build/vet/test -tags localassets ./...` green, `gofmt` clean, engine package green under `-race`. New tests are unit-scale (~0.2s total).


isaac-ranger:
Rebased onto develop @ cdbe355 (post #694/#697/#699/#703/#705/#707) — the branch was conflicting. Same three commits, new head 85ad74f. Reconciliations worth naming:

- **engine_entity_boost.go** (3-way with #705 + #699): the R1 RRF-skip block and the #699 ULID tie-break sort are preserved byte-identical to develop; the gate refactor of pass 2b is unchanged. The skip fires before seeding, so under rrf fusion the gate path is never reached — consistent with COG-18's amendment.
- **Invariant renumbered COG-21 → COG-22**: #694 took COG-21 for the prospective focality pin while this PR was open. The injection-gate invariant is now COG-22; COG-18's and COG-9's merged entries keep the R1 amendment text verbatim.
- One rebase defect was caught by an adversarial pass over the final diff before this push: the invariants.md merge had dropped #707's COG-23 entry. Restored verbatim from develop; `internal/cognitive` references anchor correctly again.

Verification on the rebased head: build/vet/gofmt clean with `-tags localassets`; full `./internal/...` green except `TestProspectiveAcceptance_Gates1to3`, which fails identically on a clean develop checkout in this environment (the #702 fixture sensitivity — not this branch). Race detector clean on the supersession/gate/boost test set. RED-check re-run post-rebase: disabling the gate's `ExcludeUntrusted` predicate reddens guard tests on both injector paths (`TestApplySupersession_HeadUntrusted_AbstainsAtomically`, `TestEntityBoost_InjectionRespectsExcludeUntrusted`).

One pre-existing observation from the review, adjacent to #704's scale class but distinct: `supersessionEpsilon` (0.01) is scale-blind — under rrf finals (~0.001–0.05) `head − ε` can go negative and over-demote a stale row. Untouched by this PR (demote-only still holds); happy to file it separately if useful.

Issue #172, opened 2026-03-10
`concept` field empty in `muninn_remember` response


Celtiore:
## Description

When storing an engram via `muninn_remember`, the response JSON contains `"concept": ""` (empty string), even though the concept is correctly persisted and visible on subsequent `muninn_recall`.

## Steps to reproduce

1. Call `muninn_remember` with `concept: "test-concept"`, `content: "some content"`, `tags: ["test"]`
2. Inspect the response JSON
3. Observe: `"concept": ""` in response
4. Call `muninn_recall` with context matching the concept
5. Observe: concept is correctly stored and searchable

## Expected behavior

Response should include `"concept": "test-concept"`.

## Observed behavior

Response returns `"concept": ""` (empty string). Data is persisted correctly — only the response is affected.

## Environment

- MuninnDB v0.3.14-alpha-1 (darwin/arm64)
- Access via MCP HTTP (JSON-RPC 2.0) on port 8750

PR #729, opened 2026-07-30
fix(mcp): report the stored concept in evolve/decide/explain/contradictions responses (#721)

### LINKED ISSUE #721: mcp: muninn_evolve's response always reports `concept: ""` — same defect as #172, which was fixed for remember only
## Summary

`muninn_evolve`'s response body always contains `"concept": ""`, regardless of the concept actually stored — whether inherited from the predecessor or set explicitly via the `concept` param added in #483.

This is the same defect as #172 (`concept` field empty in `muninn_remember` response). That fix was applied to remember's handler only; evolve was never given the equivalent.

## Verified

Reproduced five times against a v0.7.0 binary, and it is structural in current source on both `origin/main` and `origin/develop`:

```go
// internal/mcp/engine_adapter.go:92
func (a *mcpEngineAdapter) Evolve(...) (*WriteResult, error) {
	id, err := a.eng.EvolveAt(...)
	if err != nil {
		return nil, err
	}
	return &WriteResult{ID: id.String()}, nil   // Concept never populated
}
```

`WriteResult.Concept` is `json:"concept"` without `omitempty` (`internal/mcp/types.go:61`), so it always serializes, always empty. `EvolveAt` returns only a `storage.ULID`, so the resolved concept never reaches the adapter — the field is unpopulated by construction.

By contrast, remember is correct because the **handler** echoes the caller's value: `result := WriteResult{ID: resp.ID, Concept: req.Concept}` (`internal/mcp/handlers.go:187`).

Observed pair, same session:

```
muninn_remember(concept: "evolve-summary-recheck v1") -> {"id":"01KYQ8G85J…","concept":"evolve-summary-recheck v1"}
muninn_evolve(concept:   "evolve-summary-recheck v2") -> {"id":"01KYQ8GHKE…","concept":""}
muninn_read("01KYQ8GHKE…")                            -> concept: "evolve-summary-recheck v2"   ← stored correctly
```

## Impact

Cosmetic in storage terms — nothing is actually lost — but actively misleading to an agent caller, which is the primary consumer of this response. An agent that reasons or logs off the evolve response concludes the concept was wiped and "repairs" it with another evolve, churning versions (and ULIDs) for nothing. That cost real time before `muninn_read` confirmed the label was fine all along.

## Proposal

Populate `Concept` in the adapter's `Evolve`, echoing the caller's value and falling back to a read-back only when it was omitted (the inherit case):

```go
res := &WriteResult{ID: id.String(), Concept: concept}
if concept == "" {
    if eng, gerr := a.eng.GetEngram(ctx, vault, id); gerr == nil && eng != nil {
        res.Concept = eng.Concept
    }
}
return res, nil
```

`Engine.GetEngram` already exists (`internal/engine/engine.go:849`). The extra point-read is paid only when `concept` was omitted. A failed read-back degrades to today's behavior rather than failing the call — the write has already committed and must not be reported as an error.

Echoing the caller's value alone is not sufficient: with `concept` omitted the truthful answer is the inherited concept, and the response would still say `""`.

## Related

- #172 — same defect on `muninn_remember`, fixed there only
- #483 — added the `concept` param to evolve, whose value this response misreports
- #720 — the other evolve-surface gap found in the same session

timharsch:
## Summary

Closes #721.

`muninn_evolve`'s MCP response always reported `concept: ""` — the same defect as
#172, which was fixed for `muninn_remember` only. Chasing it turned up **four**
instances of one pattern in `internal/mcp/engine_adapter.go`: the engine populates
a field, the adapter constructs the wire struct without copying it, and because none
of these fields carry `omitempty`, every response ships a confidently-empty value.

| Tool | Field(s) dropped |
|---|---|
| `muninn_evolve` | `concept` |
| `muninn_decide` | `concept` |
| `muninn_explain` | `concept` |
| `muninn_contradictions` | `concept_a`, `concept_b` |

REST already populates `concept_a`/`concept_b` (`internal/transport/rest/engine_adapter.go`),
so the MCP surface was the outlier, not the feature.

This is the project's worst failure class (`CLAUDE.md` principle 1/2): the call
succeeds, the response looks well-formed, and the caller has no signal that a field
it is reading was never filled in.

## Implementation notes

- **`muninn_evolve`** — `resolveEvolveConcept(callerConcept string, readBack func() string) string`.
  A caller-supplied `concept` wins (it is what was just written); otherwise the stored
  concept is read back. A failed read-back logs a `slog.Warn` rather than silently
  returning `""`.

- **`muninn_decide`** — the concept is resolved by **reading the engram back**, not by
  echoing the `decision` argument. `e.Write`'s content-hash dedup can return a
  *pre-existing* engram's ULID, in which case echoing `decision` would produce a
  self-consistent-looking falsehood — a worse outcome than the empty string. When the
  read-back concept differs from `decision`, the substitution is surfaced through the
  already-plumbed `WriteResult.Warnings`, so dedup is loud rather than merely truthful.
  No new field on `engine.DecideResult`.

- **`muninn_contradictions`** — this tool can return every pair in the vault, so the
  concepts are fetched with one batched read rather than 2N point lookups. A missing or
  unreadable engram degrades to `""` for that side only, mirroring REST's per-side
  guards, rather than failing the whole call.

- **Read path honors COG-12.** All read-backs use the non-access-accounting
  `Engine.GetEngram`, never `Engine.Read`, so shaping a response cannot inflate
  `AccessCount` and perturb ACT-R base-level activation.

- **`DetectedAt` deliberately left empty.** Per COG-23 the `0x0A` contradiction marker
  persists only the engram pair — there is no stored timestamp to report. Filed
  separately rather than invented here.

## Testing

- `go build ./...` ✓ &nbsp; `go vet ./...` ✓ &nbsp; `gofmt -l .` clean ✓
- `go test ./internal/mcp/... ./internal/engine/... ./cmd/muninn/...` ✓
- `go test -race ./internal/mcp/...` ✓
- One real-engine adapter regression test per instance
  (`engine_adapter_concept_test.go`, `_decide_`, `_explain_`, `_contradictions_`),
  each **RED-verified**: reverted the fix, confirmed the test fails on its intended
  assertion with the empty value — not on a compile error or an unrelated check.
- `cmd/muninn/smoke_exhaustive_test.go` extended to assert the concept end-to-end
  against a live server for evolve, decide, and explain. The explain assertion is
  guarded on `would_return`, since the engram it explains has already been superseded
  by the evolve subtest.

## Scope

MCP surface only. The sibling gaps this work surfaced — REST `/api/explain` dropping
its `Components` breakdown, REST `/api/traverse` dropping `rel_type`, and the wire
fields nothing populates on any surface — are being filed as separate issues rather
than folded in here.

## Related PRs

- #730 — independent (both branch off `develop` and can merge in either order), but both add `[Unreleased]` entries, so whichever lands second needs a trivial `CHANGELOG.md` conflict resolution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   though REST's equivalent adapter has set it correctly all along. Scoped to
   the MCP surface only — REST's explain adapter already sets `Concept`.
   (#721)
+- **`muninn_contradictions` reported `concept_a`/`concept_b` as `""` for
+  every pair.** A fourth instance of the same defect shape: REST's
+  equivalent (`GetContradictions` in
+  `internal/transport/rest/engine_adapter.go`) has always read both
+  engrams back to populate `ConceptA`/`ConceptB`; the MCP adapter shipped
+  `ContradictionPair{IDa, IDb}` only. Fixed with a single batched read
+  (`Engine.GetEngramsBatch`, one Pebble iterator for every ID referenced by
+  any pair) rather than a per-pair lookup, since this tool can return every
+  contradiction in the vault. A missing or unreadable engram degrades that
+  side to `""` rather than failing the call. Scoped to the MCP surface
+  only — REST already gets this right. `detected_at` remains unpopulated;
+  the 0x0A marker persists only the pair, with no severity/type/provenance
+  to report. (#721)
 
 ---
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,12 +38,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `WriteResult.Warnings`, which already existed. Scoped to the MCP surface
   only. (#721)
 - **The `muninn_explain` MCP tool's response also dropped `concept`.** A
-  third instance of the same defect shape: the adapter copied eight score-
-  breakdown fields off the engine's `ExplainData` but never copied
-  `Concept`, so `muninn_explain` always reported `concept: ""` on MCP even
-  though REST's equivalent adapter has set it correctly all along. Scoped to
-  the MCP surface only — REST's explain adapter already sets `Concept`.
-  (#721)
+  third instance of the same defect shape: the adapter copied five of the
+  engine's `ExplainData`'s six fields (every field except `Concept`) into
+  the response, so `muninn_explain` always reported `concept: ""` on MCP
+  even though REST's equivalent adapter has set it correctly all along.
+  Scoped to the MCP surface only — REST's explain adapter already sets
+  `Concept`. (#721)
 - **`muninn_contradictions` reported `concept_a`/`concept_b` as `""` for
   every pair.** A fourth instance of the same defect shape: REST's
   equivalent (`GetContradictions` in
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,10 +16,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   concept was set correctly — the same defect class already fixed for
   `muninn_remember` in #172, but evolve never got the equivalent fix. The
   response now returns the caller's `concept` when supplied, or reads back
-  the freshly-written engram to report the inherited concept. The
-  `muninn_decide` MCP tool had the identical defect and is fixed alongside
-  it. Scoped to the MCP surface only — REST's `EvolveResponse` has no
-  `concept` field. (#721)
+  the freshly-written engram to report the inherited concept. Scoped to the
+  MCP surface only — REST's `EvolveResponse` has no `concept` field. (#721)
+- **The `muninn_decide` MCP tool had the identical empty-`concept` defect,
+  plus a second bug once the first fix landed.** A naive fix (echo the
+  caller's `decision` text) is wrong when two decisions submit identical
+  rationale text: `engine.Decide` writes through the content-hash dedup
+  path, which returns a pre-existing engram's ID, so the response would
+  report the *second* call's decision text against the *first* call's
+  engram. The response now reads the concept back from the stored engram,
+  the same way evolve does, so it never echoes text the returned engram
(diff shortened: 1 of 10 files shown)


timharsch:
Closing as superseded — the defects this PR fixes have all landed upstream while it sat, and I'd rather withdraw it than re-litigate merged code.

Verified against `origin/develop` at e24b5dc by running this branch's own tests there:

| This PR's test | Against develop |
|---|---|
| `Evolve_SetsConcept` — inherited + explicit | **passes** — #746 reads the concept back in the adapter |
| `Explain_SetsConcept` | **passes** — `explainResultFromEngine` already set `Concept` |
| `Decide_SetsConcept/fresh write` | **passes** — #746 added `DecideResult.Concept` |
| `Decide_SetsConcept/dedup` | fails at its **precondition**, see below |
| `contradictions_concept` | no longer compiles — #754 changed `FlagContradiction`'s signature while building the richer surface |

The dedup subtest is worth spelling out, because it was the one piece of this PR I thought had survived. It asserted that two `Decide` calls with different decision text but identical rationale collapse onto one engram via `Write`'s content-hash short-circuit, leaving the response echoing the caller's decision text while returning an earlier engram whose stored concept is a different decision.

That was true when this branch was cut — `Decide` composed `content := rationale`, so the decision text was not part of the hashed content. #746 changed it to `content := decision + "\n\nRationale:\n" + rationale`, which puts the decision inside the hash. Two different decisions can no longer collide, so the test now fails on its setup rather than its assertion:

```
expected content-hash dedup to collapse onto the same engram,
got r1.ID=01KYX71F4XV5EAWN0JD2RFC1Y6 r2.ID=01KYX71F51X9VZ58AW3XRP3QJB
```

Nice side effect of a change made for a different reason (putting the decision in the body so recall snippets and exports state what was decided) — it closed a silent-substitution hole in the same stroke. `develop` also carries its own `TestDecideAdapter_ReportsStoredConcept` and `TestEvolveAdapter_ReportsStoredConcept`, so the coverage is there too.

Issue #721 closed as completed, credited to #746. Nothing here left to salvage, so no follow-up issue.

My other two PRs, #730 and #732, are unaffected and I've rebuilt both on current `develop`.

PR #730, opened 2026-07-30
feat(mcp): add muninn_update_tags for in-place retagging; evolve rejects tags (#720)

### LINKED ISSUE #720: mcp: tags are write-once — no MCP tool can retag an engram, and muninn_evolve silently discards a `tags` argument
## Summary

Over MCP, engram tags are **write-once**: they can be set at creation (`muninn_remember` / `muninn_remember_batch`) and are correctly inherited by `muninn_evolve`, but **no MCP tool can change them**. Passing `tags` to `muninn_evolve` returns **success and silently discards them**.

The engine already has the capability — it is simply not exposed to MCP.

## Verified

Empirically on a v0.7.0 binary, then confirmed in source on both `origin/main` and `origin/develop` (so this is not a stale-binary artifact):

- `muninn_evolve`'s input schema has no `tags` property — `internal/mcp/tools.go:322`.
- `handleEvolve` never reads `args["tags"]` — `internal/mcp/handlers.go:701`.
- Unknown MCP params are not rejected, so the call succeeds with no signal. In the same call a `concept` argument **did** apply, so a caller sees success, sees one field land, and gets no hint the tags were dropped.
- Tag *inheritance* works correctly (`EvolveAt` sets `Tags: oldEng.Tags`), and `tags_all` recall filtering keeps working across versions. Evolve does not lose tags — it just cannot change them.
- The capability exists at engine level: `Engine.UpdateTags` (`internal/engine/engine.go:884`), reachable only via REST `PUT /api/engrams/{id}/tags` (`internal/transport/rest/server.go:232`, `handleUpdateTags` at :1645).

## Why it matters

Tag conventions that encode mutable state are unmaintainable over MCP. The `due:<ISO-date>` convention that `tags_all`/`tag_filter` were built for (#479) is the clearest case: a due date that cannot be changed is not a due date. The two available workarounds are both bad:

1. Call REST directly with an API key from `muninn api-key create` (note `~/.muninn/mcp.token` is not a REST key — it returns 401).
2. `muninn_forget` + re-`muninn_remember` with the desired tags, which destroys the ULID lineage and access history that `muninn_evolve` exists to preserve.

## Proposal

1. Add a `muninn_update_tags` MCP tool (`{vault?, id, tags[]}`) wrapping the existing `Engine.UpdateTags`, replacing the full tag set **in place** — ID, version lineage, and access history preserved. `muninn_trust` and `muninn_state` are the precedent for an in-place single-engram metadata mutator; `internal/transport/rest/engine_adapter.go:411` already does the string→ULID conversion the MCP adapter needs.
2. Make `muninn_evolve` **reject** a `tags` argument with `-32602`, naming `muninn_update_tags` in the message. Retagging via evolve would archive the memory under a new ULID to change metadata, which is the wrong semantics; a loud error that teaches the correct two-call pattern is better than either silently dropping the argument or silently versioning the memory.

Classification note: `muninn_update_tags` modifies an existing engram, so it belongs in `isMutatingTool` and **not** `isAdditiveTool` — append-mode credentials must be refused. `Engine.UpdateTags` already calls `refuseAppend`, so that is defense in depth.

## Out of scope

- Adding `tags` to `muninn_evolve` for atomic content+tag updates. Deferred until there is a demonstrated need; it would mean two ways to do one thing.
- A general "reject unknown params" policy for MCP. That is a cross-cutting behavior change and deserves its own RFC — this issue only rejects `tags` on the one tool where the silent drop is known to mislead.

timharsch:
## Summary

Closes #720.

Tags were **write-once over MCP**: no tool could retag an engram, and passing `tags`
to `muninn_evolve` returned success with the argument silently discarded. Worse, a
`concept` argument in the same call *does* take effect — so a caller saw success, saw
one field land, and got no signal the rest was dropped. `PUT /api/engrams/{id}/tags`
already existed on REST, so this was an MCP-surface gap, not a missing feature.

That makes tag conventions like `due:<ISO-date>` — which `muninn_recall`'s `tag_filter`
is built to query — unusable from an MCP client, since a date that moves can never be
updated.

Two changes, plus the correctness work they turned out to require:

1. **`muninn_update_tags(vault, id, tags)`** — replaces a tag set in place. The ULID,
   version lineage, and access history are preserved.
2. **`muninn_evolve` rejects `tags`** with `-32602`, naming `muninn_update_tags` in the
   error so a caller who guesses wrong learns the two-call pattern from the failure.

## Why this reaches past the MCP layer

The MCP tool is a thin door onto `Engine.UpdateTags` — which, it turned out, could not
be safely opened as-is. Adding the door without these fixes would have shipped a
supported path to silently-wrong recall, which `CLAUDE.md` §1 rules out.

- **Stale FTS postings.** `Engine.UpdateTags` rewrote the engram but never reindexed,
  so old tag tokens stayed searchable and new ones were invisible. A retag was a
  silent recall regression.

- **`fts.ReindexEngram` is stats-neutral by construction.** The obvious fix —
  `DeleteEngram` + `IndexEngram` — corrupts scoring, because `IndexEngram` bumps
  `TotalEngrams`, `AvgDocLen`, **and** per-term document frequency while `DeleteEngram`
  decrements none of them. Measured over 100 retags of one engram, that path loses
  **79.4%** of the target's own score, with `df` reaching **101** for a term present in
  a single document. `ReindexEngram` performs the delete and the add under one `idx.mu`
  and one batch, skips `UpdateStats`, and adjusts DF only for terms whose membership
  actually changed: **bit-identical scores at 1, 10, and 100 retags**, with `N`,
  `AvgDocLen`, and both `df` values unmoved.

  Neutrality is with respect to retag **count**, not tag-set **size** — growing a tag
  set 1→51 in place is length-penalized against a stale corpus average, which is
  legitimate BM25. The load-bearing property is that an oscillating tag set returns
  *exactly* to baseline: no hysteresis, a pure function of the current document.

- **`[STO-2]`/`[STO-3]` resurrection race.** `PebbleStore.UpdateTags` did a
  read-modify-write with no lock, so a concurrent soft-delete could be lost. It now
  holds `ps.casLocks.For(id[:])` across read→commit, matching `SoftDelete` and
  `UpdateConfidence`.

- **L1 cache poisoning.** On the rejection path the cache could retain tags that were
  never committed. Invalidation is now deferred so no early return can skip it, and the
  post-commit `cache.Set` happens inside the lock.

- **Non-active engrams are not reindexed** — a `soft_deleted` or `archived` engram must
  not be resurrected into the search index by a retag.

## Deliberate decisions

- **Not added to `coreToolNames`.** Advertised schemas ride in every MCP client's
  context window and retagging is occasional; the tool stays fully callable via
  `tools/call`, and the evolve rejection names it explicitly. Only the surface count in
  that comment changes (44 → 45).
- **No REST/OpenAPI/SDK changes** — obligations 2 and 3 don't apply; the REST route
  already exists and is untouched.
- **Rejecting `tags` on evolve is technically a behavior change**, but that argument
  never had any effect. The only thing that breaks is a call that was already silently
  failing.

## Known residual, documented in-tree

`ReindexEngram` is the first *decrementing* path in the codebase, so it is the first way
`df_t` can fall below its true value. `ftsWorker.Submit` drops jobs when the queue is
full or after `Stop`; an engram whose initial index was dropped hands `ReindexEngram` a
`prev` whose tokens were never counted, and the `-1` lands on DF another engram owns.
It is bounded at 0, one-time, and inflates IDF rather than omitting results. The clamp
matters: without it the counter wraps to `4294967295` and `getIDF` returns **-20.7944**
— a negative IDF that ranks a document *containing* the term below one that doesn't.

## Cross-surface obligations discharged

`allMCPTools` + a live-server smoke subtest, tool classification (mutating), the
`auth_mk_test.go` tool list, the `toolset.go` surface count, and a `guide.go` entry
teaching the evolve-then-retag pattern. The COG-24 addendum and obligation 13 now carve
out the full-vault rebuild (`engine_reindex_fts.go`), which is correct only because
`ClearFTSKeys` range-tombstones the FTS prefixes first.

## Testing

- `go build ./...` ✓ &nbsp; `go vet ./...` ✓ &nbsp; `gofmt -l .` clean ✓
- `go test ./...` ✓ (whole repo) &nbsp; `go test -race ./internal/{engine,storage,index,mcp}/...` ✓
- `go vet -tags integration ./cmd/muninn/` ✓
- Every behavioral fix is **RED-verified** — reverted, confirmed the test fails on its
  intended assertion. Two of these exist specifically because the behaviors initially
  shipped with the whole suite green under mutation: removing the post-commit
  `cache.Set` (caught via the in-package cache accessor — a `GetEngram` assertion
  cannot distinguish the states, since the entry is merely absent and repopulates
  correctly), and removing `adjustDF`'s zero clamp (`df=4294967295`,
  `getIDF=-20.7944`). Both run in <0.1s.
- Registry parity, the observe/write/append/full dispatch gate, and unknown-mode
  fail-closed re-confirmed across all five surfaces (45 tools / 24 mutating / 21
  read-only, empty set-differences both ways).

## Related PRs

- #729 — independent (both branch off `develop` and can merge in either order), but
  both add `[Unreleased]` entries, so whichever lands second needs a trivial
  `CHANGELOG.md` conflict resolution.


scrypster (changes requested):
Love the discipline in this one. All five of your hardened tests went red the moment we broke what they pin, we verified three of them by hand, and the stats-neutrality argument in ReindexEngram is both correct and important. The engine-level fix was genuinely necessary and we proved it: with your reindex reverted, a retagged engram keeps scoring on a tag it no longer has and cannot score on the one it gained.

Two independent passes ran, and the adversarial one found four things at the seams the comments assert are safe. Two of them invert an assumption rather than just filling a gap, so bear with the list.

1. Retagging an evolve predecessor destroys its postings permanently. The comment says phase 6 drops soft-deleted before returning candidates, but the history gate admits soft-deleted with a closed ValidUntil under as_of and include_invalid, which is exactly the evolve signature, and evolve deliberately does not delete the predecessor's postings the way forget does. We reproduced it: one hit before the retag, zero after, and only reindex-fts brings it back. So a metadata edit silently removes the keyword path to a superseded fact that as_of is supposed to still reach. Gate the drop on trash versus history rather than on State alone: drop for a plain soft-delete (open stamp), reindex for closed-stamp and archived. Do not reuse PassesLifecycle directly, it takes the asOf parameters, so factor the history test out or reuse the supersession-signature helper.

2. The 0x2C orphan residual is not deferrable in this PR specifically, because the tool description and the guide both recommend due-date tags as the use case. Ninety retags of one recurring task fully starves the tag-seeding budget, and we reproduced that with ten genuinely matching engrams crowded out entirely. The fix already exists in-tree: DeleteRawTagIndexEntry, and you are already under the stripe lock with the old tag set in hand. Diff old against new and delete the removed tags' 0x2C and 0x0C entries in the same batch. Compute the removal set before eng.Tags is overwritten.

3. The engine-level sequence needs the same serialization your storage write already has. The read-prev, write, reindex triple is not atomic, so two concurrent retags of the same engram strand the loser's postings, and we saw that in 36 of 40 trials with no artificial delay. The DF double-decrement also moved an uninvolved third engram's full-text score by 6.5 percent. Cooperative agents retagging the same memory concurrently is ordinary, not adversarial, so documenting it is not enough here. Trap: the stripe mutex is not reentrant, so you will need an unlocked inner variant or to move the reindex inside the storage critical section.

4. A non-empty tags array that normalizes to empty should reject loudly and name the offending entry, rather than replacing the whole set with an empty one and returning ok. We wiped three real tags end to end with a single 129-byte input. Keep an explicit empty array working as a deliberate clear, and consider returning a dropped count on partial normalization. Create-time leniency is right, but this replaces rather than adds, so the cost of the same leniency is different.

Plus a rebase, you will hit one take-both docs conflict in invariants.md.

None of this is a no. The core mechanism is right, the wiring is complete (we ran the live registry parity check against a throwaway daemon and it passes at 45 tools), and honestly the quality of your tests is what made finding these fast. Between this and the state classification fix you are building real depth in the MCP surface, and it shows.

timharsch:
All four addressed in d822e84, each RED-checked against the code without the fix.

**1. Trash vs history.** Confirmed before changing anything: there are exactly two `fts.DeleteEngram` call sites, and neither is on the archival path — so nothing de-indexes on archive, and the `StateArchived` arm of the old gate was destroying live postings, not cleaning up dead ones. `Evolve` soft-deletes and closes `ValidUntil` in one batch and leaves the predecessor indexed on purpose; `Forget` without `not_true_since` leaves the stamp open and de-indexes at delete time. So the stamp is the discriminator and `State` is not.

Factored out rather than reusing `PassesLifecycle`, as you asked — `activation.CarriesSupersessionSignature(eng)` in `internal/engine/activation/history_gate.go`, which `PassesLifecycle`'s soft-delete arm now calls. It takes no view parameters, which is the point: a write path deciding whether to keep an index entry has no business consulting `asOf`.

RED: with the gate back on `State`, `TestUpdateTags_EvolvePredecessorKeepsPostings` reports 0 hits for a **content** term on the predecessor after a pure metadata edit, want 1. `TestUpdateTags_TrashIsStillDeIndexed` is the negative control and passes both ways by design — narrowing the drop must not stop dropping actual trash.

**2. The 0x2C/0x0C residual.** Fixed here, not deferred. The removal set is computed before `eng.Tags` is overwritten and queued ahead of the write loops in the same batch. Diffed on the **derived key bytes**, not the tag strings — 0x0C hashes the whole tag and 0x2C hashes the tag key, both to 4 bytes, so a string diff can delete an entry a colliding surviving tag still needs, whereas a key diff degrades a collision into a retained orphan (the pre-existing, safe direction). Extracted `RawTagIndexKeyFor` so `DeleteRawTagIndexEntry` and the diff share one derivation instead of two copies drifting.

Corrected the stale comment while I was there: the old text called orphans safe because phase 6 re-checks the real tag. That is only true of false *positives*. These indexes seed candidates against a bounded budget, so an orphan burns a slot a genuine match needed — a false negative at the pool boundary.

RED: five rolls of a `due:` tag leave **6** raw-tag entries for one engram without the removal pass, **1** with it (`TestUpdateTags_RetiresStaleTagIndexEntries`). The test also asserts the survivor is the *current* date, so a pass that deleted the wrong side would not sneak through on the count alone.

**3. Serialization.** Took the unlocked-inner-variant route: `store.LockEngram(id)` returns the unlock func, `UpdateTagsLocked` is the body, and `UpdateTags` is now the locking wrapper — so the existing storage-only callers are unchanged and `Engine.UpdateTags` holds one stripe lock across read-prev → write → reindex. Checked the ordering concern: `internal/index/fts` imports only `storage/keys`, so holding `casLocks` across `fts.ReindexEngram` cannot invert the `casLocks → idx.mu` order.

RED: 5 runs of 5 with the lock removed, every one failing with postings for a tag the record does not carry (`TestUpdateTags_ConcurrentRetagsDoNotStrandPostings`, 8 concurrent writers).

**4. Loud rejection.** `normalizeTagsReporting` returns `(tags, dropped)` with a per-entry reason; `normalizeTags` stays a thin wrapper so `muninn_remember`'s behaviour is untouched. A non-empty array where nothing survives is now `-32602` naming each offending entry and pointing at the empty-array path; an explicit `[]` still clears; partial normalization succeeds with `dropped` and `dropped_detail` in the result. Added a control asserting those fields are **absent** on a clean call, so their presence stays a signal rather than noise.

RED: all four rejection cases returned `{"ok":true,"tags":[]}` before the guard — success for a call that wiped the tag set.

**Docs.** COG-19 gains a `#720` amendment stating that the stamp — not `State` — is the trash/history discriminator, naming the shared helper and cross-referencing COG-27's `NameableAsLineage`, which reads the same signal on the read side. New **STO-12** covers the tag-index removal diff, including the derived-key rule and the note that a new tag-derived prefix must be added to the removal pass, not just the write path.

`go build ./... && go vet ./... && gofmt -l .` clean; full `go test ./...` green; `-race` green on engine, storage, mcp, and index.

timharsch:
CI note: the Go job on d822e84 went red on `TestNetworkStreamer_PushLatency` (`internal/replication`) — `push latency = 56.669349ms, want < 50ms`. That is a wall-clock assertion on a shared runner, and nothing in this PR touches replication, streaming, or the WAL/fsync path. Locally it passes 5/5 (`go test ./internal/replication/ -run TestNetworkStreamer_PushLatency -count=5`), and `develop`'s last eight CI runs are all green, so it is a runner-timing flake rather than a regression here. Note the failure still proved the property the test exists for — 56ms is well under the 100ms polling floor it was written to rule out; it only missed the tighter headroom bound.

Same class as drift obligation #12 (#722): a bound calibrated to developer hardware rather than to constrained CI cores. I have no rerun rights on this repo — happy to open a separate issue to make that assertion robust (measure against the poll floor it is actually distinguishing from, or retry-with-best-of-N) rather than widening a magic number, if you want it tracked.

scrypster (approved):
Confirmed and merging.

The FTS-neutrality find is the part worth calling out — the −82.6% over 100 retags, and catching that the obvious delete-then-reindex **isn't** stats-neutral. That's easy to ship wrong and very hard to notice once it's live. I traced the fix rather than trusting it: reverted the removal loop in `UpdateTagsLocked` and got exactly your numbers back — 6 stranded raw-tag entries after 5 retags without it, 1 with it. Restored, green again.

Verified the premise too. No MCP tool could retag, and `muninn_evolve` silently dropped `tags` while `concept` in the same call *did* take effect — a caller sees partial success with no signal the rest was discarded. That's the silent-substitution failure class this project treats as its worst, so closing it matters more than the surface size suggests.

Tier-3 attack surface is covered by your own tests: retag-to-existing-tag, retag-clearing-all, retag-on-untagged, concurrent retag/soft-delete, concurrent retag/retag, rejected-tag cache poisoning, retag-on-archived. All green including `-race` on storage. The removal is an exact-key `batch.Delete` rather than a prefix scan, so none of the `PrefixUpperBound` bound-defect risk applies here.

**One thing entirely outside your control, and no action needed from you.** You numbered this invariant STO-12 — and that number got claimed twice more the same day you opened this. #803's endpoint-liveness guard kept it, and develop's own docs already flag your PR as one of the three claimants needing to move. It's **STO-18** now. Your branch also picked up conflicts in seven files from develop's refactor of the tool-classification tests — the switch-statement-to-map change that came out of #731/#732 landed after your last rebase and touched the same lines.

I resolved all of that rather than sending it back: STO-18, the obligation renumbered 13→15, and your tool wired into develop's derived registry and census instead of the old hand-typed lists. That last part matters — develop moved to derived-and-censused lists *because* the hand-maintained ones had already drifted into the observe-mode write bug you yourself reported in #731. Resurrecting them would have reintroduced exactly the fragility your own fix removed.

Full suite green on the merged tree, `-race` included, and the classification census passes in both directions with `muninn_update_tags` in the forbidden set for append mode.

You've been through here a few times now — the observe-mode `muninn_state` fix, and the logging trio (rotation, access-log volume, the stale-tail bug) that landed today. This one's in the same register: found the real defect, then found the two things a naive fix would have broken before shipping it. Thanks also for splitting the review-feedback round into its own commit; it made the diff easy to read.

PR #732, opened 2026-07-30
fix(mcp): classify muninn_state as mutating so observe-mode credentials cannot write (#731)

### LINKED ISSUE #731: mcp: `muninn_state` is classified read-only but writes — an observe-mode credential can archive memories out of recall
Found while mapping the MCP tool classification lists for #720/#721. Unrelated root
cause, so filing separately rather than folding it in. Line references are against
`origin/develop` @ `b2265b6`.

## Defect

`muninn_state` is classified as read-only but its handler writes.

- `internal/mcp/context.go:203` — `"muninn_state"` appears in `isReadOnlyTool`, and **not** in `isMutatingTool`.
- `internal/mcp/handlers.go:1001` — `handleState` calls `s.engine.UpdateState(...)`, which the adapter (`internal/mcp/engine_adapter.go:197`) forwards to `Engine.UpdateLifecycleState` (`internal/engine/engine.go:3294`) — a `CompareAndSet` write. The response even reports `"updated": true`.

## It breaks in both directions

The dispatch gate at `internal/mcp/server.go:227-247` switches on `a.Mode` for `mk_` keys
and `cap_` tokens:

| Mode | Check | Result | Intended |
|---|---|---|---|
| `observe` | `!isReadOnlyTool` → false | **allowed** | denied |
| `append` | `!isReadOnlyTool && !isAdditiveTool` → false | allowed at the gate | denied |
| `write` | `!isMutatingTool` → **true** | **denied `-32001`** | allowed |

So the credential that must not write can, and the credential that legitimately should
transition state cannot.

## Impact: observe mode is genuinely unprotected

`archived` is a valid target state (`internal/mcp/handlers.go:975`), and archived engrams
are excluded from recall — so an observe-mode credential can hide memories from every
reader in the vault.

**Append mode is fine** — defense in depth holds there. `UpdateLifecycleState` calls
`e.refuseAppend(ctx)` first (`internal/engine/engine_vault.go:89`), the transport-agnostic
chokepoint whose own doc comment explains exactly why it exists: *"Enforced here at the
engine … so the guarantee holds on every transport — the MCP dispatch gate alone would
leave REST/gRPC/MBP open."* The misclassification lets an append credential past the MCP
gate, and the engine catches it.

**There is no `refuseObserve` counterpart.** The engine's observe handling is narrower
than that:

- `resolveReadOnly` forces read-only for observe on the **read** path, suppressing
  reinforcement (`engine.go:2092, 2282, 2430, 2597, 2621`).
- `resolveTrust` (`engine.go:1041`) blocks `trust=verified` for a non-write credential.

Neither refuses a lifecycle write. So for observe, the MCP dispatch gate is the only
gate, and it is the one that is wrong. The fix therefore matters on its own — there is no
second layer behind it.

This is a fail-open on auth, against the project's own principle 4 in `CLAUDE.md`
("Fail closed on auth, fail open on presentation").

## Why it survived CI

`internal/mcp/context.go:158` states the invariant — every dispatched tool must appear in
exactly one of `isMutatingTool` / `isReadOnlyTool` — and
`TestToolClassification_CoversAllRegisteredHandlers`
(`internal/mcp/auth_mk_test.go:610`) enforces it. `muninn_state` *does* appear in exactly
one bucket: the wrong one. The invariant pins **coverage**, not **correctness**, so a
misfiled tool passes green. `TestToolClassification_UnknownToolDefaultDeny` (`:627`)
covers the unknown-tool case, which is a different failure.

## Suggested fix

1. Move `"muninn_state"` from `isReadOnlyTool` to `isMutatingTool`.
2. Strengthen the invariant so misclassification is structural rather than clerical —
   e.g. a source-scanning test asserting that every handler reaching a write path is in
   `isMutatingTool`. (Principle 6: pin the derived guarantee rather than trusting a
   hand-maintained list — the same reasoning `refuseAppend`'s comment already gives for
   `TestAppendMode_RefusesEveryDestructiveOp`.)
3. Consider whether the observe guarantee wants an engine-level `refuseObserve` chokepoint
   mirroring `refuseAppend`, so observe is not one list edit away from writing on any
   transport. That is a larger design question and probably its own RFC.

## Scope check: is anything else misfiled?

I swept all 21 tools currently in `isReadOnlyTool` and traced each handler's
`s.engine.*` calls. `muninn_state` is the only outlier — every other one reaches
read-only engine methods. `muninn_feedback`, `muninn_entity_state`,
`muninn_entity_state_batch`, `muninn_apply_enrichment`, `muninn_trust`, and
`muninn_restore` all mutate and are all correctly in `isMutatingTool`.

## Secondary, same file

`internal/mcp/engine_adapter.go:197-199` accepts `reason` and drops it:

```go
func (a *mcpEngineAdapter) UpdateState(ctx context.Context, vault, id, state, reason string) error {
	return a.eng.UpdateLifecycleState(ctx, vault, id, state)
}
```

`reason` is an advertised parameter in the tool schema (`internal/mcp/tools.go`,
`"reason": {"type": "string", "description": "Optional: why the state is being changed."}`),
so a caller supplying it gets success and no signal that it was discarded — the same
silently-dropped-argument class as #720 and #721. `Engine.UpdateLifecycleState` takes no
`reason` parameter at all, so either it should thread one through to provenance or the
schema should stop advertising it. Happy to split this out if you would rather keep this
issue to the classification bug.

timharsch:
Fixes #731.

## The bug

`muninn_state` was listed in `isReadOnlyTool` (`internal/mcp/context.go`), but `handleState` reaches a real durable write: `s.engine.UpdateState` → `mcpEngineAdapter.UpdateState` → `Engine.UpdateLifecycleState` → `store.CompareAndSet(..., CASMutation{State: &newState})`. `validLifecycleStates` includes `archived`.

MCP's dispatch gate (`internal/mcp/server.go:227`) admits an observe-mode credential to any tool where `isReadOnlyTool` is true. So an `mk_` key or `cap_` token issued **read-only could archive any engram in its vault**, and the response reported `"updated": true`.

Observe mode had nothing else standing in the way. Unlike append — which `Engine.refuseAppend` refuses on all 30 destructive methods regardless of transport (SEC-15), and which therefore was **never exploitable here** — there is no `refuseObserve`. `resolveReadOnly` only suppresses learning on the *read* path (COG-11) and `resolveTrust` only gates `trust=verified`. Neither is on this path. **The classification was the entire enforcement, and it was wrong.**

`TestToolClassification_CoversAllRegisteredHandlers` stayed green throughout, because it pins that every tool is in exactly one bucket — never that the bucket is the right one.

## The fix

Move `"muninn_state"` from `isReadOnlyTool` to `isMutatingTool`. Deliberately **not** `isAdditiveTool`: it transitions an existing engram, which is what `refuseAppend` on `UpdateLifecycleState` already says.

Effect per credential mode, each checked against the live gate rather than the prose:

| Mode | Before | After |
|---|---|---|
| `observe` | **allowed (the bug)** | denied |
| `write` | denied | allowed — see the residual below |
| `append` | allowed at the gate, refused at the engine | denied at both |
| `full` | allowed | allowed |

## Named residual: MCP write mode is broader than REST write mode

Reclassifying flips write mode from denied to allowed, because `isMutatingTool` and `isReadOnlyTool` are complementary by construction — a tool **cannot** be denied to observe and write both.

REST makes the opposite choice deliberately: `PUT /api/engrams/{id}/state` is `ReadOnlyGuard(WriteOnlyGuard(...))`, this repo's full-mode-only idiom, pinned by `TestWriteOnlyMode_ReadHandlersBlocked/SetState`.

The grant is judged acceptable because MCP write mode **already** admits every mutating tool — `muninn_forget`, `muninn_evolve`, `muninn_trust`, `muninn_merge_entity` among them — so `muninn_state` adds no new category of destructive power, and `handleState`'s response echoes only caller-supplied values, so it opens no exfiltration channel. But the asymmetry is real and this widens it by one tool. Closing it properly needs a third `isFullOnlyTool` bucket consulted before the `ModeWrite` case, which is a design change and **deliberately out of scope** (principle 5). Happy to file it separately, or to add the third bucket here if you'd rather it not ship as a residual.

## Scope

MCP is the lone outlier — no cross-surface cascade for this tool. REST is correctly gated (above); gRPC has no lifecycle-state RPC (`proto/muninn/v1/service.proto`); MBP has no state frame type (`internal/transport/mbp/frame.go`). `UpdateLifecycleState` has exactly two callers repo-wide.

**Both review passes independently swept all 20 read-classified tools for mutating engine calls, using two different oracles** (`refuseAppend` membership; and direct Pebble-writer reachability plus an async/goroutine escape-hatch scan). `muninn_state` was the only outlier in either sweep, and no mutating-classified tool is actually read-only. Of the four renamed adapter hops — the shape that made this bug invisible to grep — the other three (`Evolve`→`EvolveAt`, `GetVaultPlasticity`→`ResolveVaultPlasticity`, `RetryEnrich`) are classified correctly.

## Tests

All five additions RED-verified against the base classifier, not just asserted:

- `TestToolClassification_StateIsMutating` — mutating, not read-only, not additive.
- `TestDispatch_ObserveMode_BlocksStateTransition` — end-to-end regression. Asserts code `-32001` **specifically**, so a `-32602` argument-validation error cannot make it vacuous. Pre-fix it fails at the `Error == nil` branch: the observe key reached the handler and the write **succeeded**.
- `TestDispatch_WriteMode_AllowsStateTransition` — the other direction, so a future revert can't be caught for observe while riding along silently for write. Pre-fix: `forbidden: write-mode key cannot call this tool`.
- `TestAppendMode_DispatchGate/forbidden/muninn_state` — the restored second layer.
- `TestIsMutatingTool_MutatingSet` — had drifted to 17 of 24 names; the seven missing (`apply_enrichment`, `claim`, `compare_and_set`, `create_workflow_vault`, `intend`, `release`, `trust`) are restored. That's the list write mode admits, so it's the security-critical one. `TestIsMutatingTool_ReadSet` was likewise missing `get_enrichment_candidates`. Both now assert the complementary predicate instead of only the negative one.

The existing sweep tests over `registeredToolNames()` self-adjust and cover `muninn_state` in both directions for free.

**CI cost:** five cheap table/`httptest` cases in an existing package, well under a second. No new `-race`, Playwright, live-server, or asset-gated job. `-race` on `internal/mcp` was run locally and passes, but the change is a pure classification table with no shared state, so no permanent job is warranted.

## Docs

- **SEC-6** in `docs/internals/invariants.md` now records that *coverage is not correctness*, that the adapter rename is why this hid, and that observe has no engine-level refusal — including which subsystems *do* consult the observe decision (COG-11 via `ObserveFromContext`, SEC-14 via `auth.ContextMode`, COG-21 via the threaded `readOnly`) and that none of them refuse.
- **`docs/auth.md`**'s key-mode table said observe is enforced by "mutating REST routes and gRPC RPCs return `403`" and never mentioned MCP — the one user-facing place a reader would look for the mechanism this issue is about.
- CHANGELOG under `### Security`.

## Deferred, with reasons

- **An engine-level `refuseObserve`**, mirroring `refuseAppend`, so observe becomes transport-agnostic the way SEC-15 made append transport-agnostic. That is the principle-3 answer — it makes the bad state unrepresentable instead of policy-checked — and it is the real fix for the class. Wants its own RFC.
- **A structural test deriving classification from what a handler actually calls**, so the bucket can't be wrong rather than merely non-empty. Both sweeps above were done by hand; that's the mechanism SEC-6's closing sentence is asking for.
- **The `isFullOnlyTool` third bucket** (the residual above).
- **`mcpEngineAdapter.UpdateState` silently drops `reason`** — `Engine.UpdateLifecycleState` takes no such parameter, yet `reason` is accepted and documented in the `muninn_state` schema. Pre-existing and separable: either thread it to provenance (engine signature change) or remove it from the schema (arguably breaking). Its own PR either way.

## Review

Two independent passes with the repo's `code-reviewer`, per the Tier-3 obligation in `docs/internals/review-rubric.md` (this touches `internal/mcp/context.go`). Each found a real factual error in the write-mode framing — the first that it was reported as unaffected, the second that the REST-parity justification offered for it was exactly inverted. Both are corrected above; the classifier change itself survived both passes unchanged.

Local gate: `go build ./... && go vet ./... && gofmt -l .` clean; `internal/mcp`, `internal/auth`, and all three transports pass; `-race` on `internal/mcp` passes.


scrypster (changes requested):
Thanks! This is exactly the kind of report-plus-fix we hope for, and the diagnosis is right where it counts.

We ran two independent passes, the second an adversarial one that tried to break it and could not. It enumerated every route to the handler and found a single registration behind the dispatch gate on both transports, with no batch path and no aliasing. It swept all twenty read-classified tools through the adapter renames hunting for a second hole and came back with none, which matches your read. And it reproduced your exploit by reverting the classification: the observe credential reaches the handler and succeeds. Your end-to-end test asserting the specific error code rather than just an error is what keeps a wrong-params response from masking that, which is a detail worth calling out.

We also confirmed the underlying point independently: there is no engine-level refuseObserve anywhere, so this classification really was the entire enforcement. And your sibling observation checks out, compare_and_set was already correctly in the mutating list, which is a good clue that this one was an oversight rather than a design decision.

Four small things before it lands.

1. Make the classifier tests drift-proof rather than just re-synced. Nothing currently asserts the lists are exhaustive, which is exactly how this bug happened in the first place. Loop registeredToolNames and assert membership both directions. Trap: do not compare lengths, since registeredToolNames is itself a hand-maintained mirror and both can drift together.
2. The comment says the list "drifted to 17 of the 24 mutating tools" and mentions seven missing names, but the list held 16 and your diff adds 8. In a PR whose whole point is a wrong classification, the note about the classification should be right.
3. Qualify the changelog line about no new category of destructive power. Archive is the one operation write mode holds that records no reason (the adapter drops it), writes no provenance entry, and is not visible through list_deleted. Worth saying plainly, and the residual you mention as tracked separately should carry an issue number.
4. Either take isFullOnlyTool now, or say in the PR why it is deferred with the cost measured. It is roughly eight lines plus one clause in the dispatch gate, and it would restore parity with the deliberate hardening REST already has. Your call which way, we just do not want a security PR that widens one privilege while closing another to leave that unexamined.

None of this is a no. The fix itself survived everything we threw at it. This is your third merged contribution and two of the three have been auth hygiene, which is a genuinely useful thing to be good at. Keep them coming.

timharsch:
All four taken, head is now `d4b3203`. Two of them I've pushed back on in part — laid out below rather than buried, so you can overrule me.

## 1. Drift-proof

Taken, and further than the literal ask, because "assert membership both directions" isn't possible against a `switch` — nothing can enumerate the cases. Three surfaces had to change first:

- `registeredToolNames()` was a hand-maintained mirror of the map literal inside `dispatchToolCall`. Your trap warning was the right one and it runs deeper than lengths: a tool present in the handler map but absent from the mirror was invisible to *every* classification test, however the comparison was written. The map is hoisted to `MCPServer.toolHandlers()` and `registeredToolNames()` is now derived from its keys.
- The classifiers are maps rather than `switch` bodies, so a dead or misspelled entry is assertable. Behaviour unchanged — a missing key yields `false`, exactly as the switch fell through to `return false`.
- The four per-mode dispatch loops selected their tools with the very predicate they were meant to police. That is the actual mechanism by which #731 stayed green: reverting `muninn_state` to read-only left all four passing, because a wrong bucket simply removed the tool from the loop's own sample. They select from the census now.

Census is shaped after SEC-15's `TestAppendMode_MethodCensus` (principle #6). No length comparisons anywhere.

Mutation-verified rather than argued — five mutations that were all green before, all failing now: dead name inside a classifier; the exact #731 revert (7 failures); a tool registered and classified but undeclared; a tool added to the handler map only; a rename leaving a stale census entry.

## 2. Arithmetic

You're right, and it was wrong in both directions: the list held **16** names, not 17, against a classifier covering **23**, not 24 — so 7 omissions, and the diff adds 8 (those 7 plus `muninn_state`). The comment holding the bad numbers is gone along with the list it described; the corrected count is in the new test file's header.

Checking it turned up three more prose defects in the same commit, all fixed:

- "grepping `internal/mcp/` for the engine method turns up nothing" — false, `engine_adapter.go:280` is a literal hit.
- "write mode admits exactly this set" — false, `muninn_create_workflow_vault` is pinned full-only by a separate guard.
- "complementary by construction" — false. They're complementary because the census enforces it, which is rather the point of this commit.

## 3. Archive — two sub-claims verified, one refuted

**Holds:** the `reason` argument is discarded; `Engine.UpdateLifecycleState` has no such parameter and both adapters drop it. **Holds:** archive is invisible to `list_deleted`, which reads `StateSoftDeleted` only, and recall refuses archived engrams on every path.

**Doesn't hold:** archive *does* write a provenance entry — content-free, `Operation: "update-meta"`, `storage/engram.go:317` — and it *is* reversible; `Engine.Restore` accepts `StateArchived`. So I didn't write "no provenance". A reader could disprove it in one grep and would then discount the rest of the paragraph, which would cost more than the sentence buys.

What's there instead is narrower and I think sharper: archive is the only write-mode operation that leaves no **enumerable** trace. A `forget` is findable through `list_deleted`, an `evolve` through `as_of`/`include_invalid`; an archived engram is reachable only from an ID the caller already holds. For calibration the paragraph also names `muninn_merge_entity` — which write mode has always held — as outright irreversible, so archive doesn't read as the worst thing in the bucket when it isn't.

Residual now carries #822.

## 4. isFullOnlyTool — deferred, arguing zero benefit rather than pleading cost

`muninn_compare_and_set` has been classified mutating since before #731. Its `set_state` enum includes `archived`. `expect_state` is optional — the schema says *"Omit to skip the guard"*. It reaches the identical `store.CompareAndSet` transition. So a write-mode `mk_` key can already archive on `develop` today, and an overlay covering `muninn_state` alone closes nothing.

On parity: that premise is mine, not yours — it came out of this PR's own changelog, and it was wrong. `rest/server.go` documents `ReadOnlyGuard(WriteOnlyGuard(...))` as exfiltration prevention for routes that *"return engram data in their response body"*, not as lifecycle hardening. REST write mode is separately **allowed** to soft-delete (`TestWriteOnlyMode_WriteHandlersNotBlocked/DeleteEngram`), which is more destructive than archiving, and REST exposes no compare-and-set route to take parity from at all. That framing is corrected in this commit.

Which leaves a scoping question rather than an eight-line change: membership must cover `muninn_state` **and** `muninn_compare_and_set`, then answer for `muninn_claim`/`muninn_release`, which share the CAS primitive. Making CAS full-only takes a legitimate optimistic-concurrency capability away from write mode — your call, not something to slip into a fix PR. Scoped in #822, including folding in the bespoke full-mode guard on `muninn_create_workflow_vault` so there aren't two mechanisms doing the same job.

The deferral is pinned in code, not only in prose: `TestDispatch_WriteMode_AllowsCompareAndSetArchive` asserts a write-mode key archiving through CAS with no `expect_state`. If someone later adds the overlay and misses CAS, that test fails and tells them why.

## Two for you to rule on

- **CHANGELOG.** `.github/PULL_REQUEST_TEMPLATE.md` marks it maintainer-only at release time, and the last seven merged PRs touched zero changelog lines. Item 3 asked me to fix the wording, so I did — but say the word and I'll drop the file from the diff and you can take it at release.
- **`TestSmoke_RegistryParity` did not run locally.** Its `TestMain` refuses while a muninn daemon holds `:8750`, and I wasn't going to stop mine to get a green line. It matters more than usual here because `registeredToolNames()` changed how it is produced, so I verified the same property directly — 44 advertised, 44 registered, agreeing in both directions — but CI should be the thing that says so, not me.

Local: gofmt clean, `go vet` clean, `go test ./...` 0 FAIL, `-race` on `internal/mcp` clean. Merges into `develop` without conflict.

timharsch:
CI is green on `d4b3203` — all 10 checks.

Closing the loop on the one step I flagged as unverified: `TestSmoke_RegistryParity` **did** run, in the `CLI integration tests` job (`go test -tags localassets,integration ./cmd/muninn/...`), against a live daemon on a temp data dir — `--- PASS: TestSmoke_RegistryParity (0.83s)`, followed by all 44 `TestSmoke_AllMCPTools` subtests. So the derived `registeredToolNames()` is confirmed against a real `tools/list` response, not just against itself.

Worth noting for its own sake since it's adjacent to this PR's subject: `allMCPTools` in `cmd/muninn/smoke_exhaustive_test.go` is also a hand-maintained list, but it is *not* the defect class this PR fixes — it's checked against the **live server's** advertised tools, so one side is derived from reality and drift in either direction fails loudly. That's the same shape as the new `expectedToolClassification`: a human expectation checked against something that can't be edited to agree with it. It's currently in sync at 44/44.

scrypster (approved):
Confirmed, fixed correctly, and the sweep holds up. Merging.

I reproduced this from scratch rather than reading your trace, because you were specific first and it deserved the same. On develop, `muninn_state` is in `isReadOnlyTool` at `context.go:210`, and the write is real all the way down: `handleState` → `mcpEngineAdapter.UpdateState` → `Engine.UpdateLifecycleState` → `store.CompareAndSet` → `UpdateMetadata`, whose own comment says it lands the 0x0B state index, caches, provenance **and replication**. Not a write in name only.

The distinction that sets severity is the one you got right. The gate at `server.go:225` sits inside `dispatchToolCall` and returns `-32001` before handler lookup — that is authorization, not the advertisement-only toolset filtering in SEC-9 that deliberately fails open. Principle #4 puts it on the fail-closed side, and develop failed open. You were also right that there is no second layer: `refuseObserve` does not exist anywhere in the tree, and `handleState` never calls `resolveReadOnly`. The classification *was* the enforcement.

Your RED evidence held under my own revert, and two of the failures are proof rather than assertion. The append case fails with `invalid params: 'id' is required` — the credential got past the gate and into the handler. The observe case fails at the `Error == nil` branch, so pre-fix that key received a **success** response. Asserting `-32001` specifically, so a `-32602` can't make it vacuous, is the detail I'd have asked for had it been missing.

**The sweep is the part worth calling out.** I redid it with a different oracle from either of yours — transitive write-reachability from each engine method to any Pebble method containing a real `db.Set` — in both directions across all 44 tools. Same answer: `muninn_state` was the sole outlier.

What's worth knowing is what my oracle *also* flagged. `Read`, `Activate`, `Explain` and `Stat` reach writes too, and all four are classified correctly — the first three are suppressed for observe at the engine by `ObserveFromContext`, and `Stat` only persists a derived count. Which is why I am **not** bolting on a "derive the bucket from what the handler calls" test as a quick follow-up: the naive version is red on day one against correct code. The invariant that's actually true is narrower — roughly *no write unguarded by `ObserveFromContext`* — and the honest way to reach it is the engine-level `refuseObserve` you already deferred, which dissolves the problem instead of working around it. Your deferral was the right call, not a shortcut.

On the write-mode widening I checked your justification rather than taking it. `muninn_compare_and_set` has been mutating since before this, `set_state` includes `archived`, and `required` is only `["id", "set_state"]` — so `expect_state` really is optional and write mode already reaches an unconditional archive. Redundant grant, not a new one. Accepting it as a residual, and I appreciate that #822 frames membership as a scoping question including `claim`/`release` rather than pretending it's a one-name move.

Gates clean on your head with `-tags localassets` including `-race` on `internal/mcp`; trial merge into today's develop is clean with the full suite green. I couldn't run the registry-parity smoke test — port 8750 has my daemon on it — but the obligation doesn't apply since you add and rename no tools, and I verified the derived `registeredToolNames()` returns a set identical to the old hand-maintained one. That derivation was the highest-risk part of the refactor and it's correct.

One thing I'm changing myself after merge, so you needn't do anything: the new SEC-6 runs ~3,200 characters where its neighbours are 130–260, and `invariants.md` declares a one-line format at the top. The content is good, it's just in the wrong file — I'm trimming SEC-6 to the format and moving the narrative into the decision record under Security, beside the #612 and #611 entries. That's a filing convention you had no way to know about, not a criticism of the writing.

You've contributed here before, and this is a different kind of contribution from those. Finding it while mapping the classification lists for something else, filing it separately instead of folding it in, and then going after the three drift surfaces that made it invisible rather than just moving the one string — that's more than the fix. The census in the shape of `TestAppendMode_MethodCensus`, with the deliberate note about why you left length comparisons out, is what I'd have written myself.

Separately, so it doesn't surprise you: **#730 adds an invariant numbered STO-12, and #828 took that number today.** It'll need a renumber before that one lands — I'll pick it up on its own thread rather than leaving you to find it in a conflict.

Issue #712, opened 2026-07-29
bug(currency): recall mixes obsolete/future/current facts with no staleness signal (the 'what is true now' failure, verified on real data)


scrypster:
## Summary (verified on a real 3,296-memory production vault)
The #1 reliable-colleague gap: recall surfaces obsolete, future-state, AND current versions of the same fact side by side, with **no way to tell which is true now**. This is the "runway figure" / "reliable colleague doesn't know what's true now" failure, live on real data.

## Evidence
A query for the current version of a versioned business parameter (max_results=4). Concepts redacted; only the shape and scores are the evidence:
| rank | score | concept | reality |
|---|---|---|---|
| 1 | 1.00 | (a different-topic memory in the same domain) | different topic |
| 2 | 0.932 | (the topic, v3 — **future state**) | not yet real |
| 3 | 0.911 | (the topic, rates by tier, marked **"final"**) | **OBSOLETE** |
| 4 | 0.911 | (the topic, marked **"final v3 live"**) | the **actually current** one |

The truly-current fact is tied at rank 4 with an obsolete "(final)" fact. All are `state:1`; `is_current` is not even surfaced in the recall response. `GET /api/contradictions?vault=<vault>` → **0 contradictions** — the detector did not flag the conflicting "final" values (post-R2 it is relation-matrix/explicit-link only; it cannot catch two facts asserting *different values for the same parameter* — a semantic contradiction).

## Root insight
MuninnDB HAS the mechanisms (supersedes-aware recall, `muninn_evolve`, valid-time) — but **this real vault's data never invoked them.** The successive generations of the fact were written as independent `remember` facts over months, none linked as superseding the prior. So the machinery is inert unless:
- (a) the **write path** detects that a new fact updates/conflicts-with an existing same-topic fact and auto-links supersession (hard: same-topic conflict detection at write time), and/or
- (b) **recall** detects multiple same-topic facts with conflicting content and surfaces a staleness/conflict signal (e.g. newer-wins hint, "(final)" version markers, or a semantic-contradiction flag), and/or
- (c) `is_current` is actually computed and maintained (it currently appears to default and isn't surfaced).

This is distinct from #708 (evolve'd predecessor unreachable via as_of — a mechanism gap). This is a DATA-doesn't-use-the-mechanism gap: the reliable-colleague experience needs currency to work WITHOUT the user having manually evolved every fact.

## Why it's the top reliable-colleague priority
[...]

PR #736, opened 2026-07-30
feat(recall): self-knowledge surface — staleness + contradiction annotations

scrypster:
First increment of the recall **self-knowledge surface**: a recall with `self_knowledge: true` returns each result annotated with whether it is **stale** (superseded by a newer memory) and which other results it **contradicts** — so an agent gets its answers *with* the memory's own read on them.

## What it adds
- **Response shape (additive, `omitempty` — old clients get byte-identical responses):**
  - `mbp.ActivationItem` += `ContradictsIDs []string`; `ActivateRequest` += `SelfKnowledge bool` (REST inherits via type alias).
  - MCP `Memory` += a `self_knowledge` block `{stale, current_version, contradicts_ids}`, emitted only under the flag; `muninn_recall` gains a `self_knowledge` arg.
- **Staleness** — reuses `applySupersession`'s existing always-on `superseded_by`/`current_version` (explicit `RelSupersedes`). `stale = superseded_by != ""`. (Vault-wide *predictive* staleness is a flagged follow-on.)
- **Contradiction** — a new pure-function text detector (`internal/cognitive/contradictext/`, polarity + numeric conflict, separate from the association-matrix detector) run over the **returned, post-truncation** set, gated behind `self_knowledge` (the O(k²) part; k≤MaxResults, sub-ms).

## Invariants
- **Ranking is never changed** — the pass only appends to `ContradictsIDs`; `TestAnnotateContradictions_OrderInvariant` asserts the result order is identical with vs without it (and non-vacuously that the conflict was still found).
- **Backward-compatible** — flag off ⇒ no detection ⇒ `omitempty` drops the fields; `TestSelfKnowledge_OmittedWhenNotAttached`.
- **Paraphrases are not flagged as contradictions** (the decisive case): `TestAnnotateContradictions_ParaphraseNotFlagged`.

## Notes for review
- **Tier-3 (wire format):** the new `msgpack`/`json` fields on the MBP/REST surface — additive/`omitempty`. gRPC/proto fields deferred.
- Drift follow-on (additive, non-gating): `openapi.yaml` + SDK field parity.
- Build/vet/gofmt clean; `-race` green on engine/mcp/cognitive.

Refs the self-knowledge capability work.

### DIFF
--- a/internal/cognitive/contradictext/contradictext.go
+++ b/internal/cognitive/contradictext/contradictext.go
@@ -1,0 +1,344 @@
+// Package contradictext is a model-free, pure-function text contradiction
+// detector. Given two short statements it decides whether they carry an
+// ACTUAL conflict signal — not whether they are similar. This distinction is
+// the whole point: two statements can be embedding-near-identical paraphrases
+// and NOT conflict ("returns 200 on success" vs "responds OK when it
+// succeeds"), while two dissimilar-looking statements DO ("auth tokens never
+// expire" vs "auth tokens expire after 24 hours"). Similarity is never
+// consulted here.
+//
+// It is deliberately separate from internal/cognitive/contradict.go, which
+// detects contradictions between association RELATION TYPES over the
+// association matrix (Supports vs Contradicts, etc.). That mechanism operates
+// on graph edges; this one operates on raw text and has no storage or engine
+// dependency. Both the recall self-knowledge surface and the Push consume this
+// text detector.
+//
+// Two detectors are ported (matching the proven Python miner: precision 1.00,
+// recall 0.93, paraphrase false-positive rate 0.000):
+//
+//  1. Polarity / negation flip — same-subject statements where one negates the
+//     other (a negation cue in scope over a shared predicate) or asserts a
+//     known antonym of the other ("enabled" vs "disabled").
+//  2. Numeric / value mismatch — same-subject statements that assert a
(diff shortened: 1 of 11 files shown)


scrypster:
Closing the surface, keeping one piece.

Field by field, the proposed `self_knowledge` block is now redundant against what shipped in 0.10.0: `stale` and `current_version` duplicate the always-on `superseded_by`/`current_version` annotations (#767, COG-28) and derive from the same source; `contradicts_ids` carries strictly less than #772's always-on `unresolved_contradiction` (COG-29), which names the partner, the side, the declared-at time, demotes the score 10% and adds a response-level `conflict` block. The `SelfKnowledge` request flag would also be a third gate alongside `annotate` and the always-on path, and a third annotation location is precisely the drift that already caused a field to vanish silently at the MCP boundary.

Also stale in the PR body: it claims the Push consumes this detector. The Push shipped in #694 reading durable 0x0A pairs, not text.

What does survive is `contradictext` — and it turns out to be the one genuinely missing capability, since every contradiction signal we ship is declared-link based. Re-scoped as https://github.com/scrypster/muninndb/issues/789: feed the existing `annotations.conflicts_with` under `annotate=true`, advisory only, with per-vault lexicons rather than hardcoded English antonym lists (principle #11 — the same objection that held #712), and an in-tree measurement before the inherited precision/recall numbers can be claimed.

Issue #726, opened 2026-07-29
Prefix 0x19 is allocated twice: replication log entries and prefix.Idempotency share an identical 9-byte layout


likesjx:
## Summary

`internal/prefix/prefix.go` is documented as "the single source of truth for Pebble key-prefix byte allocations", and its package doc already flags that replication's `schemaVersionKey` overlaps `0x19 Idempotency` as "a pre-existing instance of the same bug class".

The larger instance is not noted: **the replication log entries themselves** also live at `0x19`, with a layout byte-identical to idempotency receipts.

```go
// internal/replication/log.go:20
const replicationLogPrefix = 0x19
func replicationEntryKey(seq uint64) []byte  // 0x19 | seq_be64(8)      = 9 bytes

// internal/storage/keys/keys.go:687
func IdempotencyKey(opID string) []byte      // 0x19 | siphash(op_id)(8) = 9 bytes
```

Same prefix, same length, same database. On one production store that is 116,248 replication entries and 23 receipts sharing a keyspace.

Other replication keys under `0x19`: `seqCounterKey` (`0x19|0xFF×8`), `lastAppliedKey` (`0x19,0x02,…`), `schemaVersionKey` (`0x19,0x03,…`), `clusterEpochKey`/`nodeRoleKey` (`0x19,0x03,…`), `snapCompleteKey` (`0x19,0x10,…`).

## Why it matters now

`ReplicationLog.Prune` uses a blind range delete:

```go
batch.DeleteRange(replicationEntryKey(1), replicationEntryKey(untilSeq+1), nil)
```

Any idempotency receipt whose `siphash(op_id)` falls below `untilSeq` is inside that range and gets deleted. The probability is currently negligible (`untilSeq ≈ 10^5` against a `2^64` hash space), and `Prune` is never called today (#724) — but **wiring up the prune makes this live**, and the collision risk grows linearly with `seq`.

`internal/replication/snapshot.go:289` already works around the overlap in the other direction, noting "~0.4% of [receipts] begin 0x19 0x03".

`PurgeExpiredIdempotency` (`internal/storage/idempotency.go:58`) scans the whole `0x19` range and JSON-unmarshals every value, silently `continue`-ing on the msgpack replication entries. Correct by accident, and O(entire replication log) per sweep.

## Suggested fix

[...]

Issue #632, opened 2026-07-20
CLI cluster commands always 401 against a running daemon (no session/bearer attached)


likesjx:
## Observed (v0.7.0 and v0.8.0)

Every \`muninn cluster ...\` subcommand (info, status, failover) fails with \`HTTP 401: {"code":"cluster_auth_required"}\` against a healthy daemon — the CLI sends neither an admin session cookie nor the cluster-secret bearer that /v1/cluster/* requires. Workaround we use: \`curl -H "Authorization: Bearer <cluster_secret>" http://127.0.0.1:8475/v1/cluster/info\`.

\`muninn cluster enable\` has the same shape against /api/admin/cluster/enable (admin session required, CLI attaches nothing) — REST login + cookie works fine.

Suggested: teach the CLI to read the cluster secret from the local cluster.yaml (it's on disk next to the data dir) for /v1/cluster/*, and to use -u/-p admin auth (like \`muninn vault\` does) for the admin endpoints.

Issue #627, opened 2026-07-20
Cluster catch-up death spiral: 5s sendWriteTimeout kills bulk streams, streamer restarts from seq 0 — unbounded MOL growth and re-transfer


likesjx:
## Observed (v0.8.0, 3-node cluster: 1 Linux VPS Cortex + 2 macOS Lobes over Tailscale)

A Lobe that falls behind (laptop slept) can never catch up:

1. \`PeerConn.Send\` bounds every frame write with \`sendWriteTimeout = 5 * time.Second\` (internal/replication/peer_conn.go). During bulk catch-up the receiver applies entries slower than the sender pushes; the TCP window fills and a frame write exceeds 5s → \`write tcp ...: i/o timeout\` → streamer dies.
2. \`startStreamerForLobe\` restarts with \`NewNetworkStreamer(c.repLog, peer, 0)\` — the comment says "Phase 1: always stream from seq=0". Every reconnect re-streams the entire log.
3. Loop: join → bulk stream → 5s write timeout → rejoin → from 0 again.

Measured impact on our fleet: one Lobe pinned at the same ack seq for days; ~11 GB re-transferred over the tailscale link in retry loops; the Cortex MOL grew to **7.1 GB** for a dataset of ~1000 engrams (~50 MB), because SafePrune's \`MinReplicatedSeq()\` never advanced past the stuck Lobe.

## Suggested fixes
- Separate write deadline for replication/snapshot frames (the 5s bound exists to protect the shared MSP heartbeat tick — heartbeats can keep 5s, bulk frames need minutes or flow control).
- Resume streaming from the Lobe's last ReplAck'd seq instead of 0 (the ack plumbing already exists via \`replicaSeqs\`).
- Log a warning when a streamer dies mid-catch-up with the seq it reached, so operators can see the spiral.

Related: the snapshot path's fixed \`ackTimeout = 30s\` (snapshot.go) has the same shape — too short for large snapshots on slow disks.

PR #737, opened 2026-07-30
fix(replication): prune the Pebble replication log, bounded by a backlog ceiling (#724, #725)

### LINKED ISSUE #724: Replication log (prefix 0x19) is never pruned — 10 GB/day unbounded growth on an idle primary
## Summary

`ReplicationLog.Prune(untilSeq)` (`internal/replication/log.go:213`) is implemented and documented but has **no production caller**. The Pebble-resident replication log therefore grows without bound.

The only pruner, `ClusterCoordinator.startPeriodicPrune` (`internal/replication/coordinator.go:2197`), calls `c.mol.SafePrune(minSeq)` — which globs and unlinks **sealed MOL segment files on disk** and never touches the `0x19` Pebble keys.

## Measurement

Production primary (single Cortex + one Lobe observer). Read-only key histogram over two consecutive nightly backups:

| prefix | Jul 27 keys | Jul 28 keys | Jul 27 MB | Jul 28 MB |
|---|---|---|---|---|
| `0x19` replication log | 101,363 | 116,274 | 66,252 | **76,470** |
| `0x06` | 147,510 | 147,659 | 0 | 0 |
| `0x05` | 68,295 | 68,345 | 0.46 | 0.46 |

**+14,911 entries / +10.2 GB of value bytes in 24h. Every other prefix flat.**

Splitting `0x19` by value encoding on the Jul-28 backup:

```
replication(msgpack) = 116,248 keys / 76,469.95 MB
receipts(json)       =      23 keys /      0.00 MB
undecodable          =       3
```

Mean entry ≈ 658 KB — each `ReplicationEntry` carries the full `Key` and `Value` of the replicated write, so large values (embeddings, HNSW blobs) are stored a second time in the log.

Store went **311 MB → 18 GB in 8 days** after a clean rebuild. Symptoms on an otherwise idle daemon (2 API requests/hour):

- ~40 GB/day written, ~165 GB/day read (`/proc/<pid>/io`)
- 4,136 SSTs, a new Pebble WAL every ~24s
- RSS 4.8 GB, disk 85% full

## Impact

The bloat drives compaction thrash, which slows writes, which causes the 5s replication streamer write timeouts, which keeps the Lobe flapping — which in turn pins the prune watermark (see companion issue). Self-reinforcing.

## Suggested fix

Wire `ReplicationLog.Prune` into `startPeriodicPrune` alongside `mol.SafePrune`. Note `0x19` is shared with `prefix.Idempotency`, so the existing blind `DeleteRange` should become encoding-aware (msgpack entries decode; JSON receipts do not) — or the prefixes should be separated first. See companion issues.

Version: `main` @ 3eed40c9, pebble v1.1.5.

### LINKED ISSUE #725: Prune watermark stalls indefinitely when a Lobe is unhealthy — MinReplicatedSeq() is an unbounded min across replicas
## Summary

`startPeriodicPrune` (`internal/replication/coordinator.go:2197`) gates all pruning on `MinReplicatedSeq()`:

```go
minSeq := c.MinReplicatedSeq()
if minSeq == 0 {
    continue
}
pruned, err := c.mol.SafePrune(minSeq)
```

`MinReplicatedSeq()` (`coordinator.go:1880`) returns the **minimum applied seq across all tracked replicas**, and `0` when there are none. So:

- **no replicas** → returns 0 → prune skipped entirely, forever
- **one unhealthy replica** → its low/stale seq pins the watermark for the whole cluster

There is no upper bound on retention. A Lobe that never catches up means the log is never pruned, however large it grows.

## Observed

A Lobe in a permanent join/drop cycle (~8s period; `write tcp …: i/o timeout` on the Cortex, `replication stream ended, rejoining` on the Lobe) never advanced far enough to move the watermark. Both nodes healthy, network fine (Tailscale, 32 ms RTT, TCP reachable) — the stream simply could not complete a catch-up pass within the 5s write timeout against a store bloated by #724.

Net effect: the log grew ~10 GB/day indefinitely (#724), which worsened compaction, which slowed writes, which kept the stream timing out. The watermark never moved.

## Suggested fix

Bound retention independently of replica progress — max age, max bytes, or max entries behind head — so an unhealthy or absent replica cannot pin the log forever. Pruning past a Lobe's position forces it to rejoin via snapshot, which `Prune`'s doc comment already describes as the expected consequence; that is strictly better than unbounded growth on the Cortex.

Also worth distinguishing "no replicas configured" (safe to prune to head) from "replicas exist but none have acked" (genuinely unsafe) — today both return 0.

Companion to #724.

Version: `main` @ 3eed40c9.

likesjx:
Fixes #724, fixes #725. Mitigates #726 without requiring the prefix relocation.

## The problem

`ReplicationLog.Prune` is implemented and documented but has **no production caller**. `startPeriodicPrune` calls `mol.SafePrune`, which globs and unlinks sealed MOL segment *files* — it never touches the `0x19` Pebble keys. So the replication log grows forever.

Measured on a production primary, read-only histogram over two consecutive nightly backups:

| prefix | Jul 27 | Jul 28 | Δ / 24h |
|---|---|---|---|
| `0x19` replication log | 101,363 keys / 66,252 MB | 116,274 keys / 76,470 MB | **+14,911 keys, +10.2 GB** |
| `0x06` | 147,510 | 147,659 | +149 keys, 0 MB |
| `0x05` | 68,295 | 68,345 | +50 keys, 0 MB |

Every other prefix flat. Store went 311 MB → 18 GB in 8 days. On an idle daemon (2 API requests/hour): ~40 GB/day written, ~165 GB/day read, 4,136 SSTs, a new WAL every ~24s. Mean entry 658 KB — each `ReplicationEntry` carries the full key and value of the replicated write.

## Changes

**1. Wire the prune in.** `startPeriodicPrune` now prunes the replication log alongside the MOL segments, and `Prune` returns a count so the log line is evidence rather than an assertion.

**2. `Prune` is encoding-aware, not a range delete.** `0x19` is shared with `prefix.Idempotency` (`0x19|siphash(op_id)`, byte-identical in shape to `0x19|seq_be64`). The existing `DeleteRange(replicationEntryKey(1), replicationEntryKey(untilSeq+1))` takes any receipt whose siphash lands under the watermark — negligible today at `seq ≈ 10^5`, but it grows linearly with `seq` and **wiring up the prune is what makes it live**. Log entries are msgpack, receipts are JSON, so we decode and delete only what is provably an entry at or below the watermark. This deliberately avoids requiring #726 first.

**3. `ClusterConfig.MaxLogBacklog` (default 50000).** `MinReplicatedSeq()` is a *minimum* across replicas and returns 0 when none have acked, so a single Lobe that never catches up pins the watermark and the Cortex grows without bound. Observed in production: a Lobe in a permanent ~8s join/drop loop (both nodes healthy, 32 ms RTT) never advanced the watermark, and the resulting bloat slowed writes enough to keep the stream timing out — self-reinforcing.

Replica acks still win when ahead of the ceiling. A Lobe left behind the prune point rejoins via snapshot, which `Prune`'s own doc comment already describes as the expected consequence, and which is strictly better than an unbounded Cortex. `0` restores the previous behaviour.

## Tests

- `TestReplicationLog_Prune_LeavesIdempotencyReceipts` — seeds receipts at keys inside the prune window (exactly what a range delete destroys), asserts `pruned == 57` of 60 and all three receipts survive byte-for-byte.
- `TestReplicationLog_Prune_ReportsCount` — accurate count; idempotent on re-prune.
- `TestPruneWatermark_BacklogCeilingForcesPruneWhenReplicaStalls` — five cases: stalled replica pins the watermark with the ceiling disabled; ceiling overrides a stalled replica; no replicas still prunes; a healthy replica ahead of the ceiling wins; a short log never forces a prune.
- `TestLoadClusterConfig_MaxLogBacklogDefaultsWhenAbsentFromYAML` — the default survives a `cluster.yaml` written before the field existed. Without this the fix is inert on precisely the deployments that need it.

Full suite: 50 packages, 0 failures. `go vet` and `gofmt` clean.

## Not included

The `0x19` prefix relocation (#726) — that needs a migration for `seqCounterKey`, `lastAppliedKey`, `schemaVersionKey`, and the epoch keys, and is better as its own change. The encoding-aware prune here makes the overlap safe in the meantime.


likesjx:
## Validated in production

Deployed to the primary this was measured on (v0.9.0 + this branch — `main` is exactly `v0.9.0`, so the blast radius is only this change).

**The prune fires, and only because of the ceiling:**

```
msg="cluster: pruned replication log entries" pruned=104112 until_seq=104112 forced_by_backlog=true
```

`forced_by_backlog=true` is the whole point — `MinReplicatedSeq()` was 0 (the Lobe has never acked), so without the ceiling this would have been a no-op, which is what #725 describes.

**The collision guard holds.** Key-space histogram before and after, splitting `0x19` by value encoding:

| | before | after |
|---|---|---|
| replication entries (msgpack) | 116,248 | **2,011** |
| idempotency receipts (JSON) | 23 | **35** |
| undecodable (replication metadata) | 3 | 3 |

Receipts were not touched — they grew naturally from use. A range delete would have taken any receipt whose siphash landed under the watermark; the encoding-aware prune leaves them alone.

**Reclaim** (prune + a full-range compaction, since Pebble releases tombstoned space lazily and a single compaction slot against 4,482 SSTs would have taken many hours):

| | before | after |
|---|---|---|
| pebble on disk | 20,548 MB | **286 MB** |
| SSTs | 4,482 | **48** |

286 MB matches a pre-bloat reference backup (255 MB), so this is the real working-set size.

Steady state after: ~11 entries pruned per 60s tick, tracking new writes.

## One correction, pushed as a second commit

**50000 was the wrong default.** The ceiling is an entry *count*, but entries carry the full key and value of the replicated write — mean 658 KB here. So 50000 entries retained ~34 GB, and the store settled at 6.8 GB rather than 286 MB. Only after dropping the ceiling to 2000 on that box and re-compacting did it reach 286 MB.

Default is now **5000**, with the sizing caveat documented on the field. A byte-based bound is the better design; I've noted it rather than guessed at the implementation. Happy to follow up with `MaxLogBytes` if you'd prefer that shape.

scrypster (changes requested):
Hey, thanks for this, and welcome. For a first PR here this is a seriously good piece of diagnosis.

We verified the parts that matter rather than taking them on faith. Prune genuinely had no production caller (startPeriodicPrune only ever called SafePrune, which unlinks segment files). The 0x19 overlap with idempotency receipts is real, and your encoding-aware prune is the right interim answer to it. And all three of your fixes went red when we reverted them with your tests kept in place: the blind DeleteRange eats receipts, removing the Flush and Compact leaves tombstones without reclaiming disk, and neutering the ceiling drops the watermark back to the stalled replica. Your own note about the first version being masked by the WAL is the kind of correction we would rather see written down than quietly fixed, so thank you for that too.

Four things to sort before this lands.

1. The ceiling opens a gap for a connected but lagging replica. ReadSince has no contiguity check anywhere, and before this change that was safe by construction, because prune only ever ran at MinReplicatedSeq and a replica's ack is always at or below what the streamer sent. The ceiling breaks that induction: a Lobe lagging more than MaxLogBacklog has entries deleted under the live streamer, which then skips them silently and the Lobe acks right past the hole. Fail loud instead. In drainAvailable, error when entries[0].Seq is not lastSeq+1, end the stream, and let the rejoin path heal it, since rejoin always full-snapshots. Trap to watch: after a rejoin lastSeq comes from the snapshot sequence rather than zero, so key the check off lastSeq+1 rather than assuming the log starts at 1. And a stranded receipt inside a pruned window makes ReadSince error rather than gap, which is fine, leave that loud.

2. Prune holds l.mu across the decode scan, the Flush and the Compact. Append takes that same mutex, and RepLogAppend runs after every storage batch commit on the leader, so the first prune on a large backlog stalls every leader write for its whole duration. On the deployment that motivated this, that is a lot of decoding before anything commits. Release the mutex after snapshotting the sequence bound: the scan and deletes are safe unlocked, since Append only ever creates keys above the bound and only the periodic goroutine calls Prune. Chunking that first big prune across ticks would be worth considering too.

3. One test comment names a production host. Please make it something like "a production deployment". This is a public repo and history is permanent, so it is easier to fix now than later.

4. Add max_log_backlog to the config table in docs/cluster-operations.md. It is operator-facing and it changes retention behaviour on existing clusters, so it should not only exist in code.

None of this is a no. The policy is right, the measurement work behind it is right, and once these land we will run the second adversarial pass we give every replication change and get it merged. Looking forward to the next round.

scrypster:
Wanted to flag that your cluster/replication reports are getting a concentrated pass rather than sitting in the queue — #724, #725, #726 and #826 are being worked now, along with #596 and #627–#632. Several are being fixed together because they're genuinely one causal chain rather than separate complaints, and the sequencing you'd expect matters: #726's prefix double-allocation has to land before `Prune` gets wired up, or the blind `DeleteRange` becomes destructive rather than merely absent.

Worth saying plainly: this set is the best distributed-systems signal the project has. The cluster path has been under-exercised, and you're the one actually running it multi-node and writing up what breaks. A prune that never runs, a watermark one unhealthy Lobe pins forever, observers mirroring a log they never trim — those aren't found by reading the code, they're found by operating it.

Separately, and out of genuine curiosity rather than diligence: I came across philotic-stack. The Hotel & Guest framing is a nice piece of design — the metaphor actually carries architectural weight instead of being decoration, which is rare. I'd be interested to hear how Muninn fits into it, especially alongside the SQLite Context Graph and the MemGraphRAG/LifeGraph piece. Are those complementary layers, or did you reach for them where Muninn didn't hold up? The second answer is more useful to us than the first, and I'd rather hear it than not.

If you have a video, a demo, or even a rough walkthrough of the thing running, I'd like to watch it. A distributed agent supervisor with sandboxing and a gossip mesh is a much more demanding consumer than anything we test against, and knowing what you're actually asking of the memory layer would shape what we build next.

No obligation to answer any of that — the PR review stands on its own either way.

scrypster:
Landing this. A fair amount moved underneath it tonight, so here's exactly what happened and why, since almost none of it was a problem with your work.

**Both bugs verified independently rather than taken on faith.** `git grep "\.Prune("` against develop returned nothing outside tests and `INTEGRATION.md` — which documents a caller nobody ever wrote. And `MinReplicatedSeq` really is an unbounded min. We RED-checked both of your fixes ourselves: reverting `pruneWatermark` to the bare min fails two of your five subtests.

**One thing had to be fixed before merge rather than after.** `internal/config/cluster_test.go` named a real deployment host in a comment. This repo is public and history is permanent, so it's rewritten — "real deployments have such a file already on disk" carries the same evidentiary weight. Not your fault: the pre-commit guard that catches this is maintainer-only and contributors don't have it. That's our gap, not yours.

**We found a silent-divergence bug and fixed it here rather than sending it back.** Pruning past a still-connected Lobe diverged it with no error anywhere: `ReadSince` has no contiguity check, `Applier.Apply` only skips `seq <= lastApplied`, so the Lobe applied a stream with a hole and then ACKed the head — leaving the Cortex looking at a fully caught-up replica permanently missing every write in the gap. Measured: lobe at seq 10, prune to 60, first entry served 61, `err = nil`, `lastApplied = 100`.

Your own doc comment already promised these rejoin via snapshot; the code just didn't make it happen. `evictLobesBehind` now drops those peers before the delete, reusing the same `peer.Close()` retry the snapshot-failure path uses.

**And your newest commit had a subtle one.** `Prune` held the log mutex across the flush and compaction — and that mutex is the `Append` mutex, and `Append` is on the synchronous write path. On exactly the bloated store this targets, the first prune would have stalled every write in the process for as long as compacting tens of GB takes. Our test measured zero appends completing during a prune. The flush-and-compact instinct was right, by the way — tombstones sitting on disk for hours is a real thing and worth the extra commit. It just couldn't sit on that lock.

**The larger change: your encoding-aware prune is no longer needed.** #841 landed tonight and relocated the replication log off `0x19` entirely — entries now live at `0x2F 0x01|seq`, metadata at `0x2F 0x02|name`. The idempotency receipts you were carefully stepping around are no longer in the range, so the decode-per-key scan became a plain bounded `DeleteRange`. That also dissolves the performance problem it carried: a first prune over a 76 GB log would have read and full-msgpack-decoded all of it.

We verified the range is genuinely exclusive before simplifying — if it hadn't been, we'd have kept your scan. An unnecessary-but-correct prune beats a fast destructive one.

**Worth saying plainly:** you could have wired up the existing range delete, shipped a smaller diff, and let the `0x19` collision be someone else's problem at a probability nobody would realistically hit. Instead you noticed that *wiring up the prune is what makes the overlap live*, and separated the two by encoding — which is the same trick `PurgeExpiredIdempotency` already uses in the other direction. You found the in-tree precedent. That's also what convinced us to just go relocate the prefix properly, so the whole class is gone.

Two notes on the PR body, no action needed: it still says the default is 50000 (your later commit lowered it to 5000 — the code is right), and CI genuinely never ran on this branch until now, because fork PRs need approval. So the "50 packages, 0 failures" line was your local run, not a gate. It's green now.

**#826 stays open** — actually, it closed tonight too, but by a different mechanism than the issue described. The applier doesn't mirror the log; your own numbers disproved that (a lobe held 152,866 entries against the Cortex's 116,248, and a mirror can't exceed its source). The real causes were the snapshot sender iterating the whole keyspace and `RepLogAppend` being wired unconditionally so a Lobe logged its own local writes.

This is your first PR here and it sits on top of the cluster audit you've been filing since July. #627 through #632, then #724 through #726 and #826 — that series reads like someone who kept pulling the thread instead of stopping once their own node was healthy. Every one of them was real. The sizing table in this description is the kind of evidence that makes a fix easy to trust.

Welcome properly, and please keep going.

likesjx:
Thanks for the writeup on the merge. `evictLobesBehind` is the better answer — dropping the peer before the delete fixes the cause, where the streamer-side check I'd have written from your notes only catches the symptom.

Taking you at your word that the "didn't hold up" answer is the more useful one.

## How the layers split

Complementary, and split by what happens when each is gone.

**SQLite Context Graph** — operational truth for one node: materialized guests, apartments, role incarnations, leases, config, the self-heal queue. Transactional, and it has to work when nothing else is up, because it's what the daemon reads to decide what to start. Boring on purpose.

**Muninn** — cognitive memory. The layer agents reason *with*, not the layer the runtime is built *on*.

**LifeGraph** (Memgraph) — entity-relationship structure over the operator's life. Traversal questions the other two aren't shaped for.

Not a fallback chain; three different questions. The rule I landed on is that nothing in the runtime's startup path may depend on Muninn, and everything an agent thinks with does.

## Where it didn't hold up

**1. Observers accept local writes and never forward them.** The big one. A write to a fleet-shared vault on a Lobe lands locally, replicates nowhere, and is invisible to the fleet — and nothing errors, so it looks like it worked. I built single-writer routing on my side: hotel config carries a `muninn_write_route` pointing at the Cortex node, and `default` / `user_*` writes forward over my mesh to that node, applied against its local Muninn, idempotent per `{vault}:{concept}`. Not airtight — if the forward can't be enqueued I fall back to a local write, which strands the memory on the replica rather than losing it.

This is the one place I built something because Muninn's model didn't do what I needed, rather than because I needed a different question answered.

**2. No read fallback.** The client holds one `base_url`. Local daemon down, recall fails — no second endpoint, no "ask the Cortex instead." Writes survive a local outage via the forwarding above; reads don't. I've considered routing reads over my mesh too, but that puts a mesh round-trip on the hot path of a turn and I haven't measured whether it's viable.

**3. `self_*` and `session_*` don't replicate, and neither does the vault registry.** I understand why — per-host by design. But agent-private recall is then strictly local: a node losing its daemon loses that agent's autobiographical memory for the duration, with no fleet-level story for it. The constraint I design around most.

**4. A cluster rebuild wiped the key store while my hotel DB stayed valid.** Every engram write started 401ing. The tokens in my config were well-formed and simply no longer existed on the other side — nothing said "unknown token," it just said no.

**5. Trust tier and soft-delete never reach my agents — at least half mine.** The engram type I parse into carries id, vault, concept, content, tags, confidence and timestamps, nothing else; recall then flattens that to a string for the model. So an agent can't tell a high-trust memory from a low-trust one, or see that something was retracted. I don't know how much of that is me not asking versus them not being in the activate response — if they're there, I'm dropping them, and I'd like to know which. A structured recall path is in flight on my side, not yet merged.

1 and 3 shape what I can build the most. None of them sent me looking for a different memory layer.

## What Philotic asks of the memory layer

- **Three nodes** — one Cortex under systemd, two macOS observers, on a Tailscale tailnet. Muninn is the only thing in my stack that crosses hosts directly; everything else goes through my own gossip mesh.
- **Many clients per node, not one.** Currently ~6 philote processes on one node plus gateway and MCP guests, each with its own identity, each reading and writing its own `self_*` vault alongside the shared ones.
- **Vaults scoped per agent, session, and user**, resolved at call time. Seven on my busiest node.
- **Supervised, not assumed.** The hotel probes every 60s, broadcasts a `MuninnStatus` to every guest on flip, and files unreachability into a self-heal queue. Recall sits on the hot path of a turn, so latency matters more than throughput — sampled activate calls ran 66–172ms, which is comfortable.
- **Low absolute volume.** I haven't instrumented the read/write split, so only what I measured: ~2 API requests/hour on the primary when I took the sizing numbers for the PR. The log growth was stark precisely because the store is otherwise near-idle.

The highest-value change for a consumer like mine would be **a read path that can point at the Cortex when the local daemon is unavailable.** Writes already survive a local outage because I forward them over my mesh; reads just fail. A second endpoint in the client config — or any sanctioned way to ask another node — closes the last gap where a healthy fleet still can't answer a recall.

## Demo

No video, and I'd rather say so than promise one I won't cut. What I can do is a terminal walkthrough — bootstrap, an agent turn that recalls and writes, the write landing on the Cortex rather than locally — which is the part relevant to you. Say the word and I'll record it.

Fair warning: this is an experiment, not a product, and the operational tooling is rougher than the core — my deploy path has silently no-op'd, and I've filled a disk with my own backups. What I'd stand behind is the transport discipline, the supervision model, and the standard of evidence: I try not to claim something works until I've watched it work live, which is why a fair amount in my own notes is still marked unverified.

Happy to go deeper on any of the five — especially 1 and 3.

likesjx:
Correction on point 1, before you read it and wonder which version I was on.

I described observers as accepting local writes silently and never forwarding them. That was 0.9.0 behaviour, and I wrote it in the present tense — my fleet moved to 0.11.0 the same day I posted, and the Cortex-only write gate refuses instead. I hit it myself an hour later:

```
this node is not the cluster's Cortex; writes are accepted only on the Cortex
(this node's role is ); retry against the Cortex "nid-..." at 100.64.212.8:8490
```

Refusing with the Cortex named is the right answer and it closes the failure mode I built around. My `muninn_write_route` forwarding predates it and is now belt-and-braces rather than load-bearing — I'd probably still keep it, since it turns a refusal into a completed write without the caller needing to know where the Cortex is, but that is my convenience, not your bug.

Two small things while I'm here.

The refusal message reports the role as empty — `(this node's role is )`. My `cluster.yaml` has `role: observer` set, and the node plainly knows what it is, since it refused. So the gate is right and only the message is missing the field. Minor, but an error that correctly refuses while misreporting your own role is the kind of thing that sends someone hunting a config problem they do not have. Happy to file it if you'd like it tracked.

And on the read-fallback ask from my last comment: it stands, but the framing is now sharper. With the gate in place, a Lobe is unambiguous about writes — it tells you where to go. Reads have no equivalent: a local daemon that is down just fails, with no notion of asking the Cortex instead. The asymmetry is more visible now that the write side answers properly.

Issue #600, opened 2026-07-12
`muninn upgrade` installs without integrity verification, keeps no rollback binary, and bypasses service managers


johanneshauer:
Observed at the v0.8.0 tag, `cmd/muninn/upgrade.go` (code reading; the happy path works — these are integrity/operations concerns).

## What it does today

1. **No checksum verification.** Every release publishes `checksums.txt`, and `install.sh` verifies against it — but the `upgrade` path never fetches or checks it. The tarball is downloaded and extracted straight to a temp file next to the current executable.
2. **The "verify" step executes the unverified download.** Verification is `chmod 0755` + running `<tmp> version` and substring-matching the version string — i.e. the first integrity signal is *executing* the binary.
3. **No rollback copy.** The install is `os.Rename(tmp, exe)` (upgrade.go:546 at the tag) — atomic, but the old inode is unlinked. If the new version misbehaves, there is nothing to roll back to.
4. **Service-manager unaware.** Stop/start is pid-file + direct fork (`stopProcess` on `<dataDir>/muninn.pid`, then `runStart`). On a systemd-managed host (as encouraged by `contrib/muninndb.service`), a unit variant with `Type=simple` and no pid file is common — then `upgrade` concludes no daemon is running, skips the stop, and swaps the binary **under the live service**, which keeps running the deleted inode until the next unit restart. The Pebble-flock guard in `runStart` does make the post-install start fail loudly instead of double-running — but only after the swap has already happened, and `upgrade` then reports failure with no old binary left. Related earlier report about CLI lifecycle vs. the unit fighting each other: #391.

## Suggestions

1. Fetch `checksums.txt` from the same release and verify the tarball's sha256 **before** extracting/executing anything.
2. Rename the old binary to `muninn.<oldversion>.bak` instead of unlinking it (and say so in the output).
3. Detect a service manager before touching the daemon (e.g. a systemd unit that is `active` for this data dir / `INVOCATION_ID` heritage) and either delegate stop/start to it or refuse with clear advice.

Happy to send a PR for (1) and (2) if you agree with the direction.

johanneshauer:
[...]

PR #748, opened 2026-07-31
fix(cli): keep the outgoing binary and surface the release's upgrade note (#600)

johanneshauer:
Two of the gaps #600 listed, both about what `muninn upgrade` leaves the
operator when an upgrade turns out badly. The checksum gap from the same
issue was closed in #666; these are the remainder.

## Keep a rollback binary

`selfUpdate` renamed the downloaded binary straight over the running one, so
the previous release existed nowhere on the machine afterwards. Backing out
meant re-downloading it — from a host whose muninn is already misbehaving, at
the moment its operator least wants a network dependency.

The outgoing binary is now kept next to the new one as
`muninn.<oldversion>.bak`, and the path is printed after the install.

It links rather than moves, so the original never stops existing: the install
step replaces the directory entry while the link keeps the old inode
reachable. A move would open a window in which the machine has no muninn at
all, and would leave it with none if the install then failed. Filesystems
without hard links fall back to a copy, so the artifact exists either way.
The version is sanitized before it becomes a path component — it is stamped
in by -ldflags, so it is only as trustworthy as whoever built the binary.

## Print the release's upgrade note

The flow printed a *link* to the release notes and then asked for
confirmation. A release that requires a manual step says so in those notes,
and v0.9.0 is the case in point: a one-time, one-way on-disk migration whose
note reads "back up your data directory before upgrading". An operator
running `muninn upgrade -y` never saw it.

The release body is now scanned for the blockquote the CHANGELOG and release
notes already use for this — a bolded "Upgrade note" marker — and it is
rendered before the confirmation prompt, and under `--check` too.

Deliberately quiet on anything it does not recognise: this reads release
notes to inform the operator, so a parsing miss may cost a warning but must
never cost an upgrade.

The latest-release fetch already decoded the release JSON and discarded the
body, so this needs no extra request; the `latestVersionFn` test hook becomes
`latestReleaseFn` and `latestVersion()` stays for status.go's version hint.

## Verification

- `gofmt -l` clean, `go vet ./cmd/muninn/` clean, full `cmd/muninn` suite
  green under `-race`.
- Both structural tests were shown to fail without the change, not just pass
  with it:
  - replacing the link with a rename fails
    `TestPreserveBinary_LeavesOriginalInPlace`
  - dropping the blockquote anchor fails
    `TestUpgradeNote_FindsTheBlockquoteWhenProseMentionsItFirst`
  The first version of that second test passed either way — it asserted that
  prose is ignored, which the collection loop already guaranteed. It is now a
  body where prose mentions the phrase *before* the real blockquote, which is
  the case that actually distinguishes the two.
- Not exercised locally: the Windows path (unreachable — `runUpgrade` returns
  before `selfUpdate` on Windows) and the hard-link fallback, which needs a
  filesystem without link support.

---

Refs #600. Two of its four items; the checksum one landed in #666, and the
service-manager one is a separate PR that overlaps this file — whichever of the
two lands first, I will rebase the other.

### DIFF
--- a/cmd/muninn/coverage_hardening_test.go
+++ b/cmd/muninn/coverage_hardening_test.go
@@ -2086,13 +2086,13 @@ func TestConfigureTools_Windsurf_Path(t *testing.T) {
 }
 
 // ---------------------------------------------------------------------------
-// upgrade.go: runUpgrade with injected latestVersionFn
+// upgrade.go: runUpgrade with injected latestReleaseFn
 // ---------------------------------------------------------------------------
 
 func TestRunUpgrade_SkipsDevBuild(t *testing.T) {
-	old := latestVersionFn
-	latestVersionFn = func() (string, error) { return "", nil }
-	defer func() { latestVersionFn = old }()
+	old := latestReleaseFn
+	latestReleaseFn = func() (releaseInfo, error) { return releaseInfo{}, nil }
+	defer func() { latestReleaseFn = old }()
 
 	out := captureStdout(func() {
 		runUpgrade([]string{})
@@ -2103,9 +2103,9 @@ func TestRunUpgrade_SkipsDevBuild(t *testing.T) {
 }
 
 func TestRunUpgrade_UpToDate_Injected(t *testing.T) {
-	old := latestVersionFn
-	latestVersionFn = func() (string, error) { return "v0.0.1", nil }
-	defer func() { latestVersionFn = old }()
+	old := latestReleaseFn
+	latestReleaseFn = func() (releaseInfo, error) { return releaseInfo{TagName: "v0.0.1"}, nil }
+	defer func() { latestReleaseFn = old }()
 
 	out := captureStdout(func() {
 		runUpgrade([]string{})
@@ -2116,9 +2116,9 @@ func TestRunUpgrade_UpToDate_Injected(t *testing.T) {
 }
 
 func TestRunUpgrade_NetworkError(t *testing.T) {
-	old := latestVersionFn
-	latestVersionFn = func() (string, error) { return "", fmt.Errorf("network down") }
-	defer func() { latestVersionFn = old }()
+	old := latestReleaseFn
+	latestReleaseFn = func() (releaseInfo, error) { return releaseInfo{}, fmt.Errorf("network down") }
+	defer func() { latestReleaseFn = old }()
 
 	oldExit := osExit
 	osExit = func(c int) {}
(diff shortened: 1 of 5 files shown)


scrypster (changes requested):
Okay, this is good work. We fetched it fresh and both halves are real: reverting preserveBinary to the old rename semantics reddens your leaves-original test, and your honest note about the first version of the blockquote test passing both ways checks out, the shipped version does not. The #666 checksum-before-execute ordering is untouched, we verified that against the live code rather than the description.

One thing to sort before this lands.

1. Strip control bytes from the rendered note. strings.Fields removes whitespace but not C0 controls, and we proved an escape sequence survives all the way to the terminal (it cleared the screen and retitled the window in our probe). The note is printed at the exact moment the operator is deciding whether to trust a binary swap, which is the worst possible moment for the output to be attacker-influenced. After the line handling, drop runes where !unicode.IsPrint(r) and r != ' '. Watch the ordering trap: sanitize after the line splitting, the parser needs the newlines alive until then. Happy to hand over our probe as a test shape if it saves you time.

Two nits, take or leave. The 1200-byte truncation can split a UTF-8 rune, and the real v0.9.0 note has multibyte dashes in it, so that is reachable. And the ReplaceAll in the CRLF test is a no-op, the literal already contains the CRLFs it is meant to introduce.

Also, same housekeeping note as your other one: we will strip the Co-Authored-By trailer at squash, repo policy keeps attribution lines out of the history.

None of this is a no, the shape is right. Worth knowing that #749 landed, so you will pick up a small conflict in upgrade.go on your next rebase. We merged both directions locally before landing it and the auto-merge was clean, so it should be mechanical.

Send the sanitizer and this goes in.

scrypster (approved):
Merging. Ran the full gate against current develop — trial-merge clean, build/vet/gofmt green, `-race` on `cmd/muninn` ok — and RED-checked both structural tests myself rather than trusting them.

Reverting `preserveBinary` to a bare `os.Rename` fails `TestPreserveBinary_LeavesOriginalInPlace`; removing the blockquote anchor in `upgradeNote` makes it match the prose line instead of the real note. Neither passes both ways.

This is the last of #600's three gaps — checksum landed in #666, the service-manager refusal in #749 yesterday, and this is the rollback binary.

Three things I checked specifically:

**The ordering obligation holds.** `verifyChecksum` still runs before `verifyBinary`; that's obligation #6 in the drift doc and this doesn't reorder it.

**It plays correctly with #792**, which landed today and changed `selfUpdate` to refuse reporting success while a daemon survives. The exe-scan case is handled, not just the PID-file one — `daemonWasRunning` is still set so the restart fires.

**A failed preserve refuses the upgrade** rather than silently skipping the rollback artifact. Disk full, permissions, whatever: it removes the tmp download and returns before touching the live binary. That's the one that mattered most to me — an upgrade that quietly proceeds without a rollback path is precisely the silent-degradation class this project refuses.

Nice touch sanitizing the version string before it becomes a path component, collapsing `..` and trimming leading dots. That value comes from build-time `-ldflags` and is about as trustworthy as anything gets here, and you treated it as if it weren't anyway.

One small hardening I'm adding myself after merge rather than sending back: the release-note text reaches the terminal without stripping control bytes. The threat model is thin — only release-publish access can reach it — but it's cheap to close and it's the kind of thing worth doing before the access model ever widens.

You've now touched the upgrade path, the release cache keys, the TLS/init surface and the calibration protocol. That range doesn't go unnoticed.

PR #749, opened 2026-07-31
fix(cli): refuse to self-update when a service manager owns the daemon (#600)

johanneshauer:
The third item in #600. `muninn upgrade` does not bypass the service manager
passively — it actively competes with it.

## What happens today

`selfUpdate` stops the daemon by PID, replaces the binary, and restarts it by
calling `runStart`. Under systemd the unit owns that lifecycle: a
`Restart=always` unit sees its main process die and starts its own replacement
while the binary is being swapped, and whichever process wins, the unit's
notion of the daemon no longer matches the one that is running. Nothing in the
flow notices.

The same is true in the other direction: after a successful self-update the
daemon running is one systemd did not start and does not track.

## The change

Before any of that, `runUpgrade` now asks whether a service manager owns the
running daemon, and refuses with the commands to do it properly:

    sudo systemctl stop muninndb
    sudo muninn upgrade
    sudo systemctl start muninndb

That sequence works because the guard only fires while a daemon is actually
running — stopped, there is no lifecycle to contend for, and the unit picks up
the new binary on its next start.

Placed after `--check`, so reporting that an update exists stays available on a
service-managed host; only performing one is refused.

## Detection

Ownership is read from the daemon's cgroup, which needs no external command and
no dbus. Two details decide whether it is right:

**The leaf of the cgroup path decides, not any ancestor.** A daemon started by
hand inside a systemd user session sits under
`/user.slice/.../user@1000.service/app.slice/session-3.scope` — an ancestor is
a `.service`, but the process belongs to a `.scope` and no unit manages it.
Matching any `.service` in the path would refuse to upgrade on every
interactively-started daemon.

**It cannot rely on the PID file.** `writePID` is reached only through
`runStart`'s fork path, so a `Type=simple` unit that execs the server directly
(`ExecStart=... --daemon --data ...`) never writes one — while systemd is
unmistakably managing the process. That is an ordinary way to run a
non-forking daemon, not a misconfiguration, so the guard must not depend on an
artifact it legitimately lacks. The PID file is used when present, since it
names the daemon for *this* data directory; otherwise any live process running
this same binary under a unit is enough to make a self-update unsafe.

## Scope

Only systemd is detected. launchd, OpenRC, s6 and Windows services report no
ownership, so behaviour there is exactly as before — this refuses only where it
can prove a manager exists, never on suspicion.

No `--force`. An escape hatch here would be a flag whose only use is to do the
unsafe thing, and the refusal already names the supported path. Happy to add
one if you would rather have it.

## Verification

- `gofmt -l` clean, `go vet ./cmd/muninn/` clean, full `cmd/muninn` suite green
  under `-race`.
- Both load-bearing tests were shown to fail without the code that makes them
  pass: matching an ancestor instead of the leaf fails
  `TestServiceUnitFromCgroup_InteractiveSessionIsNotManaged`, and dropping the
  process-table fallback fails
  `TestServiceManagerOwnsDaemon_FallsBackWhenNoPIDFile`.
- The cgroup and `/proc/<pid>/exe` shapes were read from a live systemd-managed
  v0.9.0 instance — effective `Type=simple`, no PIDFile, `ExecStart` running
  the server directly. Its cgroup is `0::/system.slice/muninndb.service` and
  `/proc/<pid>/exe` resolves to the installed binary, matching the fixtures
  used here; that host has no `muninn.pid`, which is what prompted the fallback.
- Not exercised: the guard end-to-end on that host, which would mean installing
  a built binary over a production one. Non-Linux behaviour is unchanged by
  construction (no /proc, no detection).

---

Refs #600, the third of its four items. Kept separate from #748 (the rollback
binary and upgrade note) because this one makes a policy call — refusing an
operation that works today — and should be rejectable on its own without
holding up the mechanical half.

Both touch `cmd/muninn/upgrade.go`, so they will conflict textually: this adds
a block after the `--check` early return, #748 changes the release fetch and the
install step. Keep-both, not a redesign. Happy to rebase this onto whichever
lands first.

*(Edited after opening: an earlier revision described the reference host as a
`Type=forking` unit with a PIDFile it never received. That was my misreading of
the base unit file — a drop-in overrides it, and `systemctl show` reports the
effective config as `Type=simple` with no PIDFile. The corrected reasoning is
above and is the stronger case: the PID file is absent by design there, not by
misconfiguration.)*

### DIFF
--- a/cmd/muninn/service_manager.go
+++ b/cmd/muninn/service_manager.go
@@ -1,0 +1,158 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+)
+
+// Detection of a service manager owning the running daemon.
+//
+// `muninn upgrade` stops the daemon by PID and restarts it itself. When a service
+// manager owns the daemon that is not a safe sequence: the service manager owns the
+// restart policy, so a systemd unit with Restart=always races the upgrade — it can
+// bring the daemon back up while the binary is mid-replacement, and afterwards the
+// unit's view of the process no longer matches reality.
+//
+// Only systemd is detected. Other managers (launchd, OpenRC, s6, Windows services)
+// report no ownership, so behaviour there is unchanged from today.
+
+// readProcCgroup returns the contents of /proc/<pid>/cgroup. A variable so tests can
+// supply cgroup layouts that cannot be produced on the machine running them.
+var readProcCgroup = func(pid int) (string, error) {
+	b, err := os.ReadFile(fmt.Sprintf("/proc/%d/cgroup", pid))
+	return string(b), err
+}
+
+// serviceUnitFromCgroup returns the systemd unit named by a /proc/<pid>/cgroup body.
+//
+// The leaf of the cgroup path is what decides ownership, not any ancestor. A daemon
+// started from an interactive shell inside a systemd user session sits under
+// ".../user@1000.service/app.slice/session-3.scope" — an ancestor is a .service, but
+// the process belongs to a .scope and no unit manages its lifecycle. Matching on any
+// ".service" in the path would report every interactively-started daemon as managed.
+func serviceUnitFromCgroup(content string) (string, bool) {
+	for _, line := range strings.Split(content, "\n") {
+		line = strings.TrimSpace(line)
+		if line == "" {
+			continue
+		}
+		// cgroup v2: "0::/system.slice/muninndb.service"
+		// cgroup v1: "1:name=systemd:/system.slice/muninndb.service"
+		parts := strings.SplitN(line, ":", 3)
+		if len(parts) != 3 {
+			continue
+		}
+		cgPath := parts[2]
+		if cgPath == "" || cgPath == "/" {
+			continue
+		}
+		// cgroup paths are always slash-separated, independent of the host's separator.
+		leaf := cgPath[strings.LastIndex(cgPath, "/")+1:]
+		if strings.HasSuffix(leaf, ".service") {
+			return leaf, true
+		}
+	}
+	return "", false
+}
+
+// serviceUnitForPID reports the systemd unit that owns pid, if any.
+func serviceUnitForPID(pid int) (string, bool) {
+	content, err := readProcCgroup(pid)
+	if err != nil {
+		// No /proc, no cgroup file, or no permission: we cannot show that a service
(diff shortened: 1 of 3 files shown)


scrypster (approved):
Nice. The leaf-versus-ancestor cgroup decision is the load-bearing call here and we proved it earns its keep: mutating the matcher to accept any .service ancestor reddens your interactive-session test, and deleting the process-table fallback reddens the no-pidfile test.

We also went and proved the harm rather than assuming it, in a Linux container running the real upgrade path (a self-signed fake GitHub serving the releases API, checksums and a genuine tarball, so the actual fetch, checksum verify, verifyBinary and install ran). On a Type=simple unit, develop prints "Stopping daemon ✓" having stopped nothing, exits 0, and leaves the old daemon serving traffic off a deleted inode while muninn version reports the new one. With Restart=always it is worse: the daemon gets SIGTERMed out from under systemd, restarted off the old binary, muninn.pid destroyed, and the upgrade's own restart dies on the Pebble lock. So the thing you were worried about is real, and a bit uglier than the PR describes.

Your guard fires before anything destructive, which we checked the direct way: the fake GitHub's access log for a refused run contains exactly one line, the releases API call. No checksums fetch, no tarball, no temp file, no stop, no rename. And we measured the false-positive side across five scenarios and found none, including the subtle one, an interactive daemon whose cgroup ancestor is a .service. Your leaf-only rule is what makes that come out right.

Two coverage notes, filed as #792 rather than held against this PR. Our own shipped contrib/muninndb.service does not trip the guard, because the unit keeps MUNINNDB_DATA to itself so the pid-file lookup misses, and the exe scan compares /usr/local/bin/muninn against /opt/muninndb-src/muninndb-server. And launchd is a no-op for it entirely. The root cause underneath both is broader than systemd: selfUpdate infers "a daemon is running" from a pid file only muninn start ever writes, then prints its checkmark unconditionally. That is the follow-up, and your guard is the right hard-refusal case within it.

Two smaller things, neither blocking. The pid-file branch early-returns even when that daemon is unmanaged, so a hand-started daemon on the default data dir can mask a managed one elsewhere (the comment claims otherwise, worth aligning one or the other). And systemd user units get sudo advice where systemctl --user is correct, cosmetic since refusing is still the right outcome.

One housekeeping note: we are stripping the Co-Authored-By trailer at squash, repo policy keeps attribution lines out of the history. Merging.

PR #750, opened 2026-07-31
fix(enrich): parse errors carry a category, never the model's output (#641)

### LINKED ISSUE #641: Enrichment parse-error logs can expose raw model output

johanneshauer:
Enrichment parse errors embedded up to 160 bytes of the model's response via
`truncateForError`, and those errors are logged. Model output is a function of
memory content, so a fragment of it in a log is a fragment of a user's memory
in a log.

## One correction to the issue

The issue says the payload "may also appear in verbose LLM logs". It is not
limited to verbose. `Run` logs each failed stage with `slog.Warn(..., "err",
err)` — warn level, no flag — so on a default-configured server every parse
failure writes response bytes to the log. `recordParseError` and
`recordComplete` are the verbose-only paths on top of that. The aggregate
error returned by `Run` also carried the payload onward to its callers.

The regression test covers both levels, and it is the default-level subtest
that fails without the fix.

## The change

Parse failures now return a typed `*ParseError` carrying stage, a stable
category (`unmarshal_failure`, `type_mismatch`, `empty_result`), and the
response's size in bytes — enough to tell an empty reply from a wall of prose
without reproducing either. `truncateForError` is gone; nothing referenced it
elsewhere and it had no test coverage.

`recordParseError` additionally lifts `category` and `response_bytes` out as
log attributes, so the classification can be filtered on without parsing prose.

Two details worth flagging for review:

- A type mismatch reports `json.UnmarshalTypeError.Field` but deliberately not
  its `Value`. The standard library renders `Value` as descriptions like
  `"number -5"` — that would put a value from the response straight back into
  the message. `Field` comes from our own struct tags.
- The byte count is of the response as received, not of the extracted JSON, so
  "the model returned 4 KB and none of it parsed" stays visible.

This is the same line the transport layer already holds: `providerHTTPError`
drains the provider's body specifically so it is "never retain[ed] or log[ged]".
The parse layer was the inconsistent one.

I did not add an opt-in payload-dump mechanism. It would be new surface for a
diagnostic nobody has asked for yet, and the category plus size covers the
cases I can name. Happy to add one behind an explicit off-by-default flag if
you want it.

## Verification

- `gofmt -l` clean, `go vet ./internal/plugin/...` clean, `go build ./...`
  clean, `internal/plugin/...` green under `-race`.
- No existing test asserted on the old payload-bearing message, so nothing
  needed rewriting to accommodate this.
- RED-proven: restoring the pre-fix error for the entities path fails
  `TestParseError_MessageCarriesNoPayload/entities` and both subtests of
  `TestPipelineRun_DoesNotLogModelOutput`, on the canary string.
- `TestParseError_RemainsPermanentNotRetryable` pins the #640/#643 boundary: a
  parse error is still not a `*plugin.ProviderError` and still does not abort
  the pipeline, so malformed output stays permanent and transient provider
  failures stay retryable.
- `TestParseError_SuccessfulParsingIsUnchanged` pins that valid responses parse
  exactly as before.

---

Closes #641. @dpearson2699 said on the issue this was free to pick up — shout if
that crossed anyone's plan and I will step back.

Independent of #748 / #749; touches no shared files.

### DIFF
--- a/internal/plugin/enrich/parse.go
+++ b/internal/plugin/enrich/parse.go
@@ -3,7 +3,6 @@ package enrich
 import (
 	"bytes"
 	"encoding/json"
-	"fmt"
 	"strings"
 
 	"github.com/scrypster/muninndb/internal/plugin"
@@ -103,6 +102,7 @@ func extractJSON(s string) string {
 // ParseEntityResponse parses the JSON response from the entity extraction call.
 func ParseEntityResponse(raw string) ([]plugin.ExtractedEntity, error) {
 	raw = strings.TrimSpace(raw)
+	n := len(raw)
 	jsonStr := extractJSON(raw)
 
 	if rawEntities, ok, err := extractTopLevelField(jsonStr, "entities"); err == nil && ok {
@@ -111,23 +111,23 @@ func ParseEntityResponse(raw string) ([]plugin.ExtractedEntity, error) {
 		}
 		var entities []plugin.ExtractedEntity
 		if err := json.Unmarshal(rawEntities, &entities); err != nil {
-			return nil, fmt.Errorf("invalid entity response JSON: %s", truncateForError(jsonStr))
+			return nil, newParseError("entities", n, err)
 		}
 		return validateAndDedupeEntities(entities), nil
 	}
 
 	// Try to parse as direct array
 	var entities []plugin.ExtractedEntity
-	if err := json.Unmarshal([]byte(jsonStr), &entities); err == nil {
-		return validateAndDedupeEntities(entities), nil
+	if err := json.Unmarshal([]byte(jsonStr), &entities); err != nil {
+		return nil, newParseError("entities", n, err)
 	}
-
-	return nil, fmt.Errorf("invalid entity response JSON: %s", truncateForError(jsonStr))
+	return validateAndDedupeEntities(entities), nil
 }
 
 // ParseRelationshipResponse parses the JSON response from the relationship extraction call.
 func ParseRelationshipResponse(raw string) ([]plugin.ExtractedRelation, error) {
 	raw = strings.TrimSpace(raw)
+	n := len(raw)
 	jsonStr := extractJSON(raw)
 
 	if rawRelationships, ok, err := extractTopLevelField(jsonStr, "relationships"); err == nil && ok {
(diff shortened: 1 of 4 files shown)


scrypster (approved):
This is great, and the part we valued most is that you closed the whole class rather than the one reported path.

We verified it independently rather than taking the tests on faith. On current develop, with slog wired exactly the way the server wires it at the default info level, all four ordinary model behaviours leak: truncated JSON, a fenced block, a plain-English refusal, and a type mismatch. That last detail is the one that matters, because it means this was not a misbehaving-provider bug. It was routine small-model output putting memory content into WARN logs. A sweep of 18 hostile responses across the four parsers found 45 leaking cases on develop and 0 on your branch.

Two things we found that go beyond what the PR claims, both in your favour. The same error string also reaches the ring buffer that backs the web console's Logs page, and it reaches the caller over MCP and REST through retry_enrich. Because you fixed at the error-construction source rather than at the log sites, all of those are cleaned automatically. That was the right layer.

We also reverted your fix with your tests kept in place and watched all four parser paths and both log modes go red, and we checked the subtle one: the json UnmarshalTypeError handling deliberately reads the field name and never the value, so a model-chosen map key cannot escape either.

One tiny non-blocking note. Bytes counts after TrimSpace, so it slightly undercounts the wire response. Fine for its purpose, just noting it so nobody later treats it as exact.

Filed #790 for the remaining surface in the same class: the embed providers still interpolate raw response bodies into their errors, and the embed request body carries the memory text. Separate scope, not yours to carry here.

Merging. You have been all over the enrich and embed correctness work for a while now, and it does not go unnoticed. Thank you for the discipline on this one. A privacy-class fix is exactly where closing the class instead of the instance earns its keep.

Issue #719, opened 2026-07-29
followup(recall): measure per-embed-model anisotropy baseline on a canonical diverse corpus (reproducible, not vault-derived)


scrypster:
COG-26 (semantic floor) ships bge-small b=0.520 (=μ+1.3σ) derived from measurements on a single production vault. Per principle #11, this should be a reproducible MODEL characterization, not a value derived from one vault: ship a small canonical diverse-text corpus (unrelated sentences across many domains), embed it with each supported model, compute mean+kσ random-pair cosine → the model's registry baseline, with provenance. Add a `muninn calibrate-model` / documented procedure so the constant is reproducible and new models (ollama/openai/voyage) get characterized the same way. Also: the k in μ+kσ (currently 1.3) is a precision/recall knob — document it. This makes the per-model default honest+reproducible while the per-vault `SemanticFloor` override handles outliers.

johanneshauer:
Happy to take this one — here is a second model characterized against a canonical corpus, plus three findings about the procedure itself that I think have to be settled before the numbers mean anything.

## Measurement

**Model:** `bge-m3`, 1024-dim, served by Ollama, `/api/embed`, no instruction prefix, L2-normalized before cosine.
**Corpus:** 160 sentences across 20 unrelated domains, 8 per domain (volcanology, beekeeping, railway signalling, jazz history, textile manufacturing, maritime law, ornithology, glassblowing, monetary policy, typography, forestry, epidemiology, chess, cartography, ceramics, civil engineering, immunology, astronomy, cooking, paleontology). Neutral general-knowledge statements, comparable length, no proper nouns tied to any organization.

| pair population | n | mean | sd | μ+1.3σ |
|---|---|---|---|---|
| **cross-domain only** | 12,160 | **0.4111** | **0.0578** | **0.4863** |
| same-domain | 560 | 0.4897 | 0.0658 | 0.5752 |
| all pairs | 12,720 | 0.4145 | 0.0604 | 0.4931 |

The row to register for this model is the cross-domain one: **b = 0.4863** at k = 1.3.

## Finding 1 — "diverse corpus" is not a sufficient spec; same-domain pairs must be excluded

[...]

Issue #716, opened 2026-07-29
finding(entities): real vault is largely entity-sparse (tag-organized) — entity-anchored features under-fire


scrypster:
While measuring #712 on a real 3,296-memory production vault clone: many real memories have **zero extracted entities** and are organized by tags instead. E.g. both canonical facts in the versioned chain return `entities: (none)` with rich tags (5 tags each — topic, version, and status markers such as `final`).

Implication: features that anchor on the entity graph (entity-boost, entity-scoped recall, #712 version-clustering, cross-domain discovery) will UNDER-fire on real vaults where entity extraction didn't populate. The tag layer is the actual organizing signal here.

Open questions:
- Is entity extraction failing/disabled on write for this vault, or is it expected sparsity? (The scout couldn't test entity/graph quality — needs an admin session.)
- Should tag-topic overlap be a first-class anchor signal alongside entities for these features?
- Is there a backfill path to extract entities for existing entity-sparse memories?

This reframes several reliable-colleague features to lean on tags (which real data has) rather than assume entity-rich graphs.

johanneshauer:
A second vault's denominator — and it points the other way on the headline finding.

Measured on an independently-used production vault: 2,892 engrams, ~5 months, multi-user, external Ollama `bge-m3` (1024-dim) rather than the bundled bge-small. Aggregate counts only; no content, concepts, entity names, tags, or vault names left the machine — same discipline as the #743 write-up.

## This vault is not entity-sparse

| window | engrams | with ≥1 entity link | coverage |
|---|---|---|---|
| whole vault | 2,892 | 2,751 | **95.1%** |
| ≤ 2026-06-07 (pre-enum) | 1,229 | 1,200 | 97.6% |
| ≥ 2026-06-15 (enum active) | 1,173 | 1,086 | **92.6%** |

The post-enum figure is the one worth putting next to #743's: **92.6% here against 12.8% there**, over the same calendar window and the same server version.

So "real vaults are entity-sparse" does not reproduce. It looks like a property of **how a vault is written**, not of real vaults.

## The mechanism, via your own control

Re-running your batch-vs-single split on this vault. Batch detection here is not a judgement call — the inter-write gap distribution is sharply bimodal (p25 = 0 ms, p50 = 15.7 s), and the result is flat across thresholds from 200 ms to 2000 ms:

| write path | pre-enum | post-enum | change |
|---|---|---|---|
[...]

PR #752, opened 2026-07-31
feat(embed): canonical corpus + reproducible anisotropy baseline (#719)

johanneshauer:
The model-only half of #719: a fixed corpus that ships with the repo, and the
statistic computed over it, so a `noiseBaselineRegistry` value can be
re-derived by anyone for any embedder and reviewed as a diff rather than
trusted.

Deliberately does NOT touch the registry, and does not add a CLI surface —
see "what this leaves open" below.

## What lands

- `internal/plugin/embed/anisotropy_corpus.jsonl` — 160 short general-knowledge
  statements over 20 unrelated domains, 8 per domain, no proper nouns tied to
  any organization, embedded via `go:embed`. Written for this purpose.
- `internal/plugin/embed/anisotropy.go` — `AnisotropyCorpus()` and
  `AnisotropyBaseline(entries, vectors, k) → AnisotropyResult`, carrying μ, σ,
  b=μ+kσ, pair count, and p95/p99/max of the unrelated-pair distribution.
- `docs/internals/anisotropy-calibration.md` — the protocol.

## Three things the measurement forced

**Same-domain pairs must be excluded, and the corpus format has to support
that.** A corpus of N sentences over D domains does not supply N(N−1)/2
unrelated pairs — within-domain pairs are related by construction and sit
measurably higher, so counting them inflates b and makes recall abstain more
than the operator asked for. The contamination scales with the square of
per-domain depth: 4.4% of pairs at 8 sentences per domain, 9.5% at 20. This is
why the corpus carries a `domain` label per line; a corpus shipped as bare text
can only be calibrated approximately.

**k is the only part of μ+kσ that encodes a product decision.** μ and σ are
model properties; k trades false-admits against false-abstains. The doc records
it as one global default with the per-vault `SemanticFloor` override, and asks
that a chosen k be defended by reporting what share of unrelated pairs it
admits — because "μ+1.3σ" reads like "almost nothing gets through" and on a
1024-dim model I measured it is about one unrelated pair in ten.

**b belongs to the model *and its implementation*.** Pooling, instruction-prefix
convention (bge v1.5 uses one for queries, bge-m3 does not), truncation and
quantization all move the distribution, so two callers can both honestly say
"bge-small-en-v1.5" and be entitled to different constants. `AnisotropyResult`
carries the numeric provenance; the doc lists the rest.

Relatedly, `AnisotropyBaseline` normalizes internally rather than trusting the
caller, because Ollama's `/api/embed` and `/api/embeddings` return normalized
and unnormalized vectors for the same model and input. There is a test pinning
scale-invariance.

## A protocol divergence worth settling before more entries land

The shipped `bge-small-en-v1.5` value (b=0.520, μ=0.450 σ=0.054) came from 432
nonsense-query × real-passage pairs — a query→passage distribution. This
procedure measures random cross-domain sentence pairs. Both estimate "cosine
between unrelated text", but they are different populations and the numbers are
not interchangeable. Whichever you prefer should be used for every entry, or
the registry stops being comparable to itself.

I have a bge-m3 (1024-dim, Ollama) characterization under this procedure —
cross-domain μ=0.4111 σ=0.0578 over 12,160 pairs, b=0.4863 at k=1.3 — but
deliberately did not add it to the registry: it would change runtime behaviour
for bge-m3 users on the strength of a protocol that differs from the one behind
the existing entry. It belongs there once the protocol question is settled.

## What this leaves open

`muninn calibrate-model` is not here. Wiring it needs a decision I would rather
take from you than guess: an operator-invoked command that writes a registry
entry, or a build-time step that regenerates a checked-in table. I lean to the
latter for shipped models — the constants stay reviewable in a diff and
reproducible in CI — with the command available for characterizing a model you
do not ship, which is the case I actually have. Say which and I will add it.

## Verification

- `gofmt -l` clean, `go vet ./internal/plugin/embed/` clean, `go build ./...`
  clean, package green under `-race`.
- `TestAnisotropyBaseline_ExcludesSameDomainPairs` is the load-bearing one and
  is constructed so it fails if the exclusion is removed: it recomputes with
  every entry given a unique domain (so no pair is ever same-domain, i.e. all
  N(N−1)/2 pairs count) and asserts the mean rises. My first version of that
  test relabelled into two alternating domains, which still excludes
  same-parity pairs and so proved nothing — it reported fewer pairs, not more.
- Corpus invariants are tested too (size, domain spread, no duplicate text),
  since a duplicated line would be a same-pair masquerading as a cross-domain
  one.

---

Refs #719. Complements the measurement I posted on the issue.

Scoped to the library + corpus + protocol on purpose: the CLI shape is a design
call that is yours, and I did not want a speculative command to be the reason
the reproducible-corpus part is hard to review. Happy to add
`muninn calibrate-model` in whichever shape you pick, on this branch or a
follow-up.

### DIFF
--- a/docs/internals/anisotropy-calibration.md
+++ b/docs/internals/anisotropy-calibration.md
@@ -1,0 +1,92 @@
+# Measuring an embedding model's noise baseline
+
+The semantic-abstention floor (COG-26) rescales cosine by a per-model constant
+`b`, held in `noiseBaselineRegistry` (`internal/plugin/embed/baseline.go`). `b`
+is the point above which a cosine is evidence of relevance rather than an
+artefact of the model's anisotropy.
+
+This document is the procedure for deriving `b` for a model, so the constant is
+a reproducible property of the model rather than a measurement of whichever
+corpus happened to be at hand. Re-derive; do not hand-edit.
+
+## Why a canonical corpus
+
+A baseline derived from a deployment's own vault characterizes that vault's
+subject matter as much as the model. Two consequences:
+
+- It cannot be reproduced by anyone else, so it cannot be reviewed — only trusted.
+- It travels: a constant measured on one corpus ships to every deployment,
+  including those whose text looks nothing like it.
+
+`internal/plugin/embed/anisotropy_corpus.jsonl` is the fixed alternative: 160
+short general-knowledge statements across 20 unrelated domains, 8 per domain,
+carrying no proper nouns tied to any organization. It ships in the repo, so a
+change to a constant arrives as a diff alongside the corpus that produced it.
+
+## The statistic
+
+For corpus entries `e_i` with embeddings `v_i`:
+
+1. L2-normalize every `v_i`. (Cosine is scale-invariant, so this also makes the
+   result independent of whether the backend returns unit vectors — Ollama's
+   `/api/embed` and `/api/embeddings` differ on exactly this for the same model
+   and input.)
+2. Take the cosine of every pair `(i, j)`, `i < j`, **where `domain(i) != domain(j)`**.
+3. `b = μ + kσ` over that population.
+
+`AnisotropyBaseline` in `internal/plugin/embed/anisotropy.go` implements this.
+
+### Same-domain pairs are excluded, and that is not a detail
+
+A corpus of *N* sentences over *D* domains does not supply *N(N−1)/2* unrelated
+pairs. Within-domain pairs are related by construction and sit measurably
+higher, so including them inflates `b` and makes recall abstain more than the
+operator asked for. The contamination scales with the square of per-domain
+depth: at 8 sentences per domain it is 4.4% of pairs; at 20 it is 9.5%.
+
+This is why the corpus format carries a `domain` label per line. A corpus
+shipped as bare text cannot be calibrated correctly, only approximately.
+
+## k is a policy choice, not a measurement
+
+`μ` and `σ` are properties of the model. `k` is not — it is the only part of
+`μ+kσ` that encodes a product decision, and it trades false-admits against
+false-abstains. It therefore belongs in one documented place with a per-vault
+override (`SemanticFloor`), never as a per-model value, which would hide a
+policy choice inside a measurement.
+
+To choose or defend a `k`, report what it excludes on the corpus, e.g.:
+
+| k | share of unrelated pairs at or above the floor |
+|---|---|
+| 1.0 | ~16% |
+| 1.3 | ~10% |
+| 2.0 | ~3% |
+| 3.0 | ~0.3% |
+
+(Indicative figures from one 1024-dim model; recompute per model.)
+
+## What a registry entry must record
+
+`b` alone is not reproducible. Two callers can both honestly say
+`bge-small-en-v1.5` and be entitled to different constants: pooling (CLS vs
+mean), whether an instruction prefix is prepended (bge v1.5 uses one for
+queries; bge-m3 does not), truncation, and quantization all move the
+distribution.
+
+So a characterization should carry: model identifier, backend/runtime, pooling,
+prefix convention, dimension, corpus size and pair count, `μ`, `σ`, `k`, the
+resulting `b`, and a hash of the corpus used. `AnisotropyResult` holds the
+numeric part of that.
+
+## Protocol divergence to be aware of
+
+The shipped `bge-small-en-v1.5` value (`b = 0.520`, μ=0.450 σ=0.054) was derived
+from **432 nonsense-query × real-passage pairs** — a query→passage distribution.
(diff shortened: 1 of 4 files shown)


scrypster (changes requested):
Thanks for this one, and for building the instrument rather than hand-waving the reproducibility problem. #719 asked for a canonical corpus precisely so the next model can be registered honestly, and this is that.

We ran your procedure end to end against the bundled bge-small-int8 rather than reading it. Your k-table holds up almost exactly on a 384-dimension model: 15.6, 9.7, 2.7 and 0.2 percent above the floor for k of 1, 1.3, 2 and 3, against your predicted 16, 10, 3 and 0.3. That cross-model agreement is the strongest thing in this PR and it deserves to be stated as a result rather than left implicit.

The same run is why it needs changes first. Your procedure produces b = 0.586 for the exact model the registry carries at 0.520. And 0.586 is above 0.558, which an adversarial pass already proved too aggressive: at that value a genuine paraphrase measured at cosine 0.59 false-abstained and recall returned empty. We also measured max cross-domain cosine at 0.78 on your corpus, where COG-26 states the model's noise ceiling as roughly 0.596. So:

1. Publish the divergence in the protocol-divergence section, with the numbers rather than adjectives. Please do not resolve it by touching 0.520, the relevance-band work that just shipped calibrates against it.
2. Populate AnisotropyResult.Model and add a corpus hash. Your own doc says a characterization must carry the model identifier and a hash of the corpus, and right now the field marshals empty on every run.
3. The contamination arithmetic: 4.4 percent is right and we reproduced it exactly, but 9.5 percent does not follow from anything we can derive, and the share is linear in per-domain depth at fixed N rather than quadratic. The exclusion itself is right, we measured it moving b by +0.009 on the real model, so state that measured number instead.
4. Specify batch discipline in the protocol. One 160-text batch versus 160 single calls moves b by 7.8e-4, and two people following the procedure need to get the same digits.
5. One sentence in the doc recording that all 160 lines are original prose written for this corpus under the repo licence. Cheap to write now, unfalsifiable later. We read all 160 and they are clean, but the artifact should say so itself.
6. Cross-link the new doc from baseline.go and COG-26, and note which protocol the shipped 0.520 came from. Anyone registering the next model will otherwise follow the old pointer.

Worth naming honestly: our read is that your sentence-to-sentence population is a comparability instrument across models, and the shipped query-to-passage 0.520 matches the runtime use better, since the floor is applied to query-to-passage cosines at recall time. Whether a registry entry should ever be derived from the canonical corpus is a decision that deserves its own adversarial pass rather than arriving as a side effect of this PR. That is not a criticism of the corpus, it is the reason the divergence has to be written down.

You have carried the embed-correctness thread for a long stretch now and it shows. Please send the revisions.

scrypster (approved):
Merging — and I read this one hardest for the trap it would have been easy to fall into.

**You didn't ship a number that becomes law.** The registry isn't touched; `grep` for `noiseBaselineRegistry` in the diff shows zero writes. And you specifically **held back the bge-m3 characterization you already have** (μ=0.4111, σ=0.0578, 12,160 pairs) because it was measured under a different protocol than the one entry currently in the registry.

That's the right call, and worth saying plainly: I would not have caught you doing the wrong thing here, because you caught it yourself first. Principle #11 exists because a number tuned on one corpus quietly becomes every other vault's answer — and the failure mode is that it looks like diligence right up until it's wrong for someone else's data.

The framing in `anisotropy-calibration.md` is exactly right too: μ and σ are model properties, k is a policy choice that belongs in one documented place with a per-vault `SemanticFloor` override, never as a per-model value. That matches the existing shape at `baseline.go` — a per-model cold-start hint with a per-vault escape hatch — rather than inventing a parallel one.

Verified: trial-merge clean, build/vet/gofmt green, `-race` on `internal/plugin/embed` ok. RED-checked the one test you call load-bearing by removing the same-domain exclusion from `AnisotropyBaseline` — filtered and unfiltered collapse to identical values and pair counts, so it proves nothing without the fix. It holds up.

The corpus is 160 lines across 20 domains, 20K on disk — trivial permanent cost. I scanned every line for proper nouns: only generic technical terms. Clean for a public repo.

**The one decision that's mine, not yours:** which protocol wins for the registry going forward — your cross-domain sentence pairs, or the existing nonsense-query × passage method behind the current `bge-small-en-v1.5` entry at 0.520. You correctly declined to make that call in a PR, and I'd rather settle it deliberately than let it default. Once I do, send the bge-m3 entry and/or `calibrate-model` in whatever shape fits.

Landing the corpus, library and doc now regardless — they're useful independent of that call.

Separately: your cross-vault entity-coverage measurement on #716 falsified our own generalization, and that's in the decision record. Measuring a second vault and reporting that it contradicts the first is the contribution this project values most.

PR #754, opened 2026-07-31
fix: solid ground — consolidate data loss, inverted abstention, contradictions surface, and the write-effort cliff

scrypster:
Four defects found by putting the product in front of AI models as its **actual user** — two full rounds of hands-on evaluation, ~90 live tool calls each, plus a third round on this branch that caught a regression *in this branch*.

## What's fixed

| defect | evidence |
|---|---|
| **`consolidate` destroyed the fact** | archived its own survivor; recall → 0 |
| **abstention was inverted** | refused answerable queries, answered unanswerable ones at 0.996 |
| **`muninn_contradictions` unusable** | `[]` for 30s, empty concepts, zero `detected_at` |
| **the write-effort cliff** | 7.7× content in JSON → **2.3×** |

### `consolidate` — using the merge tool destroyed the memory

`Consolidate` writes the merged content then archives every input. But `Write` is exact-content deduplicated, so when the merged text matched an input **byte-for-byte** — the *natural* case, since the obvious merge of near-duplicates is usually one of them verbatim — it returned that input's existing id, which the archive loop then soft-deleted.

```
consolidate(ids=[A,B], merged=<A's content>)
  → {"id": A, "archived": [A, B]}        ← archived its own survivor
  read A → soft_deleted   read B → soft_deleted   recall → 0 results
```

A success response naming an id that no longer resolved. Fixed by skipping the input that *is* the merged result; a second test pins that genuinely-new content still archives both inputs, so the fix can't silently become "stop archiving."

### Abstention — one line, not two bugs

The gate compared against `final`, which is divided by the query's own maximum (`scale = 1/maxRaw`). That pins the argmax to exactly its Confidence, so **the best candidate of *any* query was unconditionally exempt from abstention** — including a query the vault cannot answer. The same division pushed genuine matches under the bar.

It got worse with use: `raw` exceeds 1.0 whenever the ACT-R prior saturates (3.24× on full Hebbian boost), so **the more a vault is used, the more reliably garbage queries got confident answers**. Separately, the prior multiplied `ContentMatch` *before* the gate, so a recently-touched memory faced a bar up to **32× lower** than an untouched one.

Two halves, neither shippable alone: gate on `absolute = min(Raw, ContentMatch, 1) × Confidence`, and move the surface default 0.5 → 0.1. `ContentMatch` is structurally capped at 0.6 (semantic-only) / 0.4 (lexical-only), so **no honest absolute score reaches 0.5** without near-verbatim wording. Recall worked at 0.5 *only because the rescale lied*.

Ranking is untouched — `final` still orders results, so Hebbian and PAS keep full power to **promote** a relevant memory; they simply cannot promote an irrelevant one past the bar.

**Deliberately excluded:** the RRF and legacy `DisableACTR` paths still gate on `final`. Neither computes a comparable `ContentMatch`. Both carry a comment saying so. I over-applied this twice and caught it via tests.

Measured, deterministic across four runs: **NDCG@5 0.5508 → 0.6410, FPR 43.8% → 6.2%.**

Verified live, both directions:
```
"who is the scheduler tech lead"  (raw cosine 0.818)   0 results → returns it
"what is the annual budget"                          0.996 hit → 0 results
literal control                                          1.000 → 1.000
```

### Contradictions — pending is not none

An explicit `contradicts` association is committed before `muninn_link` returns, so the read path unions the 0x0A markers with a scan for edges that have no marker yet, labelled `pending_detection`. The caller gets the pair *immediately*, plus `pending_count` / `detected_count` / `scan_complete`. Detection remains entirely the worker's job — the read path flags nothing and submits no `ConfidenceUpdate` (that would re-entangle the penalty #747 just untangled).

Concepts resolve on read (a rename would otherwise leave a stale concept — the exact plausible-wrong-value class this surface exists to stop). `detected_at` now comes from the marker value; **no new prefix**, legacy 16-byte values still decode to an *unknown* time that is omitted rather than emitted as `0001-01-01`. Benchmarked: 5.1 ms per 100k associations.

### The middle gear

`entities: ["PostgreSQL", "Auth Service"]` — bare strings, types resolved from the vault's own entity table. Plus inline `[[markup]]`, a tokenizer and no model. Measured with tiktoken on identical facts:

| | tokens | × content |
|---|---|---|
| bare content | 130 | 2.0 |
| **bare strings** | **171** | **2.6** |
| **`[[markup]]`** | **152** | **2.3** |
| fully structured | 503 | **7.7** |

Marginal price of an entity-bearing write: **+373 → +22**. Neither mechanism invents a name; types are resolved only for names the *caller* supplied.

## Honest notes

- **A regression shipped in this branch and was caught by re-evaluating it.** `[[markup]]` was wired into `remember` but not `evolve`, so identical input stored differently per verb. Fixed here; found by a model mid-evaluation, not by a test.
- **The measurement harness was flaky and fixing it changed the numbers.** The hot set was chosen by ranging a *map*; with it pinned, the old gate's FPR is 43.8%, not the 19–31% the flaky runs suggested. My first determinism fix pinned the timestamp and produced a perfectly reproducible measurement of *nothing* — age went to 30 days, the prior never saturated, and both arms matched. The age is what must be pinned.
- **Known gaps, not fixed here:** `absolute_score` / `abstained` are not plumbed through the MCP response, so at that surface an empty result still isn't self-describing; self-links (`A contradicts A`) are accepted; and a write is not recallable until the embedder catches up, silently — an evaluator called that last one disqualifying.

## Verification

RED-first throughout, with RED-sanity shown both ways. Two tripwire tests that *pinned* the old abstention defect now assert the corrected behaviour. The historical-0.5 test is updated with reasoning rather than silently flipped. `go build` / `go vet -tags localassets ./...` clean, `gofmt -l` empty, full `./internal/... ./cmd/...` green. Live end-to-end on sandbox daemons on non-default ports; `~/.muninn` never touched. Synthetic fixtures only; every agent grepped its own diff.


scrypster:
Review cycle complete. Summary for the record:

1. **Adversarial review** (refute-mode) returned NEEDS-WORK with 8 findings — including two silent-wrong regressions this branch itself introduced: plain-text `evolve` destroying content (the markup extractor's `("", nil)` contract assigned unconditionally), and the "REST recall default" edit actually landing on `handleSubscribe`, which would have made every write fire every push subscription.
2. **All 8 fixed** in `6b39085` — threshold ownership centralized in the engine's fusion-aware coerce (ACT-R 0.1 / weighted_sum 0.5 / rrf 0.001, MCP now forwards 0 like every other transport), explain given a real negative-threshold diagnostic bypass and gated on the same quantity recall gates, the second 0x0A writer unified with timestamp carry-forward, the consolidate guard comparing parsed ULIDs, and the docs/tripwires reconciled.
3. **Independent re-verification** reproduced the RED state for every fix that claimed one ("every empirical claim in its message survived independent reproduction") and returned APPROVE with 3 residuals — a test that passed against pre-fix code, three stale doc sites, and the negative-threshold wire posture.
4. **All 3 addressed** in `f0c41b0`: the type-resolution half of the markup fix is now actually pinned; the last 0.5-era prose is re-derived as clearly-labelled history; and negatives are clamped at REST/gRPC/MBP — the in-process bypass is deliberately not a wire contract, so out-of-range wire input gets exactly the pre-change semantics.

Full suite green; `-race` green on storage, activation, and all three wire transports.

Known residuals, deliberately deferred and on record: self-link guard is MCP-only (REST/gRPC/MBP/engine.Link can still write one); `[[markup]]` covers remember/remember_batch/evolve but not the tree verbs or apply_enrichment; REST `/api/contradictions` still returns flat pairs without pending state; gRPC does not map `abstained`; nested `[[a[[b]]c]]` partially mangles.

Issue #763, opened 2026-08-01
Version-cluster queries must resolve to the current head — fresh-evolve misses make evolved truth unretrievable (round-6 GPT knockout)


scrypster:
Round-6 hands-on evaluation (sandbox at 33f1230), GPT evaluator, its stated single highest-value change:

> "Make current-version retrieval an invariant. If a query semantically reaches any member of a version cluster, recall should resolve it to the authoritative current version before thresholding and ranking. Call 18 should be impossible."

## The reproduced failure

Immediately after `muninn_evolve` (last-writer-wins → field-level merging), the natural query "What is the current rule for merging simultaneous edits?" returned only an adjacent memory and **omitted the freshly evolved decision entirely**. A rephrase in deep mode found it. The evaluator's framing: "a memory system that stores truth but fails to retrieve it under a natural phrasing is not dependable enough to drive autonomous decisions" — and this is now the **only** blocker between this evaluator and primary-memory use ("I would genuinely use this build as secondary long-term memory").

Mechanism (to be verified): the query's wording matched the predecessor's text, but the predecessor is expired/excluded from default recall, and the successor's fresh embedding/wording didn't clear the gate — so the truth was stored, current, and invisible. The relevance the stale wording earned was discarded instead of redirected to the declared head.

## Proposed invariant (declared chains only — mechanically sound, no guessing)

When a candidate matches at any pipeline stage but is excluded because it is superseded/expired, and it carries a DECLARED `RelSupersedes` chain, substitute the chain head into the candidate pool with the match's admission evidence (before thresholding, annotated `substituted_for: <predecessor id>`). This is confined to explicit declarations — the same channel the currency work (#758) treats as ground truth — so it cannot launder similarity into supersession.

Also check the embed-lag side: a successor whose embedding hasn't landed yet must count as "not indexed yet," not "not relevant" (loud-degradation doctrine).

Evaluator's abandonment criterion, for the record: "if misses like call 18 remained common enough that every recall required several paraphrases and modes."

Related: #758 (advisory precision — the annotation layer), round-5/6 eval findings, the fresh-evolve retrieval window blocker carried since round 4.

Issue #762, opened 2026-08-01
Association decay grinds every edge to the 0.05 floor within ~a day on active vaults — learned weight structure cannot survive


scrypster:
Found during the #756/#759 post-deploy accounting (aggregate scan of backup copies; details in https://github.com/scrypster/muninndb/issues/756#issuecomment-5151988047).

## Observation

Strongest surviving association weight store-wide, across three backups of the origin deployment:

| Backup | max assoc weight |
|---|---|
| 2026-07-28 | ~0.62 |
| 2026-07-31 09:08 | ~0.76 |
| 2026-07-31 17:34 | **~0.05** (the floor) |

A decay pass between 09:08 and 17:34 ground every edge in the store to the pruning floor. This is not one vault — it's the store-wide maximum.

## Why this matters more than it looks

- **"Strengthens with use" cannot be observed under this regime.** #757 just restored full-confidence Hebbian learning after the encoder bug, but if decay flattens all weights to 0.05 within roughly a day, learned salience has a half-life too short to influence anything — weight-ordered traversal (0x03 is sorted by weight) degenerates to arbitrary order, and phase4/PAS boosts read from a flat landscape.
- **It silently biases the planned cognition ablation trial** (full-cognition vs content-only arms on real queries): if the trial runs on a vault whose edges are all at the floor, the cognition arm is structurally handicapped and the result would measure the decay configuration, not the mechanism.
- Possible interaction with the very bug #757 fixed: for weeks, every edge that reached weight 1.0 read back as 0 — LTP could never persist a strong edge, so decay always operated on a population with no protected peak. Whether the current decay rate is inherently too hot, or was tuned/observed against that corrupted population, needs to be established before touching constants (principle #11: per-vault self-derived thresholds, never hardcoded from one sample).

## Suggested first steps (measure before changing anything)

1. Instrument or offline-compute the decay half-life implied by the current parameters vs access rates on real vault copies (aggregates only).
2. Check whether reinforcement (access-driven LTP) can outrun decay at realistic recall frequencies, now that 1.0 edges persist.
3. Only then decide: per-vault decay calibration, a slower floor approach, or a protected band for high-confidence/declared edges (declared edges arguably should not decay to indistinguishability at all — a declared `contradicts` at 0.05 ranks below noise).

Related: #760 (role-blind decay in clusters), #756 (the accounting that surfaced this).

PR #832, opened 2026-08-02
docs(invariants): COG-22 amendment label — #763 is COG-28, not COG-27

isaac-ranger:
One-word docs fix: the COG-22 amendment added with #763 cites its invariant as COG-27, but #763's invariant is **COG-28** (version-head substitution). COG-27 is association decay (#762) — a reader following the pointer lands on the wrong invariant. No code change.


scrypster (approved):
Verified and correct — merging.

Checked the claim independently rather than taking it: on current develop, **COG-27 is association decay** (`DecayAssocWeights`, the peak-anchored elapsed-time ceiling, from #762) and **COG-28 is version-head resolution** ("if a query's evidence reaches ANY member of a DECLARED version chain…", from #763). The COG-22 amendment cites #763 by number and COG-27 by label, so the two disagree and a reader following the pointer lands on decay instead of substitution.

Worth more than a one-word fix suggests. COG-22 is one of the denser invariants in the file and its amendment is precisely the part someone reads when they need to know *why* a predecessor may be named as lineage. Sending them to the wrong invariant costs the reader the reasoning.

Thanks for reading the invariants closely enough to catch a cross-reference — that is not a document people usually proofread.

PR #874, opened 2026-08-12
fix(grpc): regenerate proto stubs via protoc-gen-go (#873)

madeinoz67:
Closes #873.

## Summary

The committed proto stubs (`proto/gen/go/muninn/v1/{service.pb.go, service_grpc.pb.go}`) were hand-written Go structs carrying protobuf struct tags but **no `proto.Message` implementation**. The server uses grpc-go's default proto codec (`grpc.NewServer` with no custom codec), whose decode is `proto.Unmarshal(bytes, msg)` — which requires `msg` to satisfy `proto.Message`. So every RPC failed before reading any wire bytes:

```
rpc error: code = Internal desc = grpc: error while marshaling:
proto: failed to marshal, message is *muninn_v1.HelloRequest, want proto.Message
```

All gRPC RPCs were broken for any standard `protoc-gen-go` client. MBP / REST / MCP were unaffected (they don't traverse the proto codec). Discovered building the go-rag ↔ MuninnDB bridge.

## The fix

Regenerate the stubs from `proto/muninn/v1/service.proto` with `protoc-gen-go` v1.36.11 + `protoc-gen-go-grpc` v1.6.2 (pin-matched to `go.mod`).

The **wire format is unchanged** (the proto diff is one line: the `;muninn_v1` Go-package suffix; verified by marshal comparison). The **Go API moves** from the hand-written initialism style (`ID`/`OK`/`SessionID`/…) to the `protoc-gen-go` style (`Id`/`Ok`/`SessionId`/…) — protoc-gen-go does not apply the Go initialism rule. So the in-package callers in `internal/transport/grpc/` are updated. The `mbp` (engine) side of `engine_adapter.go` is left untouched — only the `pb` field names change. Also `[]ActivationItem` (value slice) → `[]*ActivationItem` (pointer) for repeated messages.

Supporting changes:
- **`batch_extra.go`** — relocate `BatchWriteRequest.GetVault()` (the auth interceptor's `vaultNamer` path depends on it; `BatchWriteRequest` has no `vault` field, so `protoc-gen-go` doesn't generate the getter).
- **`service.proto`** — `option go_package = "...muninn/v1;muninn_v1"` pins the Go package name to `muninn_v1` (what the hand-written stubs used) so regenerated output is a drop-in. Wire `package muninn.v1;` unchanged.
- **`go.mod`** — `google.golang.org/protobuf` promoted from `// indirect` to direct.

## Tooling (prevents recurrence)

[...]

PR #876, opened 2026-08-12
test(engine): fix -race flake in COG-29 contradiction-debt fixture (#875)

### LINKED ISSUE #875: TestContradictionDebt_ZeroDeclaredAtIsUnknownNotEpoch flakes under -race (#871 / COG-29)
## Summary

`TestContradictionDebt_ZeroDeclaredAtIsUnknownNotEpoch` (added by #871 / COG-29, commit `e39045f`) flakes under `-race`: it fails 2/5 runs with `want two pairs, got Count:1`, and intermittently fails at the fixture's `WriteAssociation` with `association write would replace a different rel_type at the same weight: existing rel_type 16 ... new rel_type 2`. It passes reliably without `-race`. This makes `develop`'s CI red under the normal `-race` gate, and any PR merged against current `develop` inherits the red (spotted on #874).

## Reproduction

Clean `develop` @ `3a18561` (no other branch's changes):

```
$ go test -race -run 'TestContradictionDebt_ZeroDeclaredAtIsUnknownNotEpoch' ./internal/engine/ -count=5
--- FAIL: TestContradictionDebt_ZeroDeclaredAtIsUnknownNotEpoch (0.14s)
    engine_contradiction_debt_test.go:332: want two pairs, got &{Count:1 ...}
--- FAIL: TestContradictionDebt_ZeroDeclaredAtIsUnknownNotEpoch (0.13s)
    ...
FAIL
```

Passes 3/3 without `-race`. No `DATA RACE` report is printed — it's a timing-dependent logic outcome under slower `-race` instrumentation, not a detected data race.

## Two failure modes

1. **`engine_contradiction_debt_test.go:320`** — `debtPairFixture`'s explicit `WriteAssociation(RelContradicts, weight 0.8)` returns `ErrAssocRelTypeCollision` ("existing rel_type 16 … new rel_type 2"). I.e. the `autoAssoc` worker has already emitted a competing `rel_type 2` edge between the near-identical pair at weight 0.8 before the fixture writes the declared edge.
2. **`engine_contradiction_debt_test.go:332`** — both fixtures' declared-edge writes succeed, but `ContradictionDebt` returns `Count:1` (one declared pair not seen by the readout).

## What I've ruled out

- **Silent clobber.** `internal/storage/assoc_reltype_guard.go` returns `ErrAssocRelTypeCollision` on a rel_type mismatch at the same `(src, weight, target)` — it is a hard reject, "no explicit replace affordance." So the declared `RelContradicts` edge is not silently overwritten; mode 1 is the guard doing its job, and in mode 2 both declared edges survive on disk.
- **Naive stale cache.** `declaredContradictionsCached` is keyed on `store.ContradictsWriteGen(ws)`, which bumps on declared-edge writes and invalidates the cached scan. The readout runs once, cold, at the end of the test.

## Likely race surface (not yet pinned)

The `autoAssoc` worker (drained by `WaitWriteTimeIdle`) emits a `rel_type 2` association between the two near-identical engrams each fixture writes ("number 6 frog" vs "number 8 frog"; "26 mm" vs "24 mm"), at weight 0.8 — the same weight the fixture uses for the declared `RelContradicts` edge. The fixture calls `waitWriteTimeIdle()` *after* writing the declared edge, so under `-race` the worker's emission can interleave with the declared-edge write (mode 1) or with the readout's scan/gen bookkeeping (mode 2). This is the `docs/internals` testing-hermeticity (#722) async-drain class: the dependent write/readout isn't deterministically serialized against the worker.

The part I have not yet pinned is exactly how mode 2 loses a pair when both declared edges are on disk — whether it's the gen counter, the scan, or pair-direction handling under the slower interleaving. I'm instrumenting that path now.

## Suggested fix direction

- **Test hermeticity (#722):** serialize the fixture against `autoAssoc` — drain the worker after the two `Write` calls and *before* the declared-edge `WriteAssociation`, so the worker's auto-emission settles first; and confirm the readout runs against a fully quiescent engine. This is likely the minimal correct fix and is contained to the test/helper.
- **If mode 2 is a readout bug:** the gen counter / scan / pair-direction logic in `declaredContradictionsCached` + `contradictionReportFrom` may have a window under `-race` where a just-written declared edge isn't reflected. Worth a close look regardless.

## Context

Discovered because PR #874's CI merges against current `develop` and inherits this red; #874 itself (proto-stub regen + gRPC transport) is unrelated and its own jobs are green. I'm continuing to investigate and will follow with a fix PR linked here.

`make`/`go` repro is above; happy to pair on the instrumentation.

madeinoz67:
Fixes #875.

## Root cause

`TestContradictionDebt_ZeroDeclaredAtIsUnknownNotEpoch` (and other callers of `debtPairFixture`) flake under `-race` on a slow/contended runner (CI `ubuntu-latest`; reproduced in a `--cpus 0.5` linux container). `debtPairFixture` writes an explicit **declared** `RelContradicts` (rel_type 2) edge between two near-identical facts at weight 0.8. The contradiction worker independently **detects** the same near-identical pair and emits its own (different-RelType, 0x10) detected edge at the same weight. When the worker's edge lands first, `assoc_reltype_guard` rejects the fixture's declared write:

```
engine_contradiction_debt_test.go:320: association write would replace a different
rel_type at the same weight: existing rel_type 16 at weight 0.8 … new rel_type 2
```

and the fixture's `WriteAssociation` fatals. (A second failure mode — `Count:1` — is the same race observed downstream.)

The guard is intentionally one-RelType-per-`(src, weight, target)` (`assoc_reltype_guard.go`: "does not give multi-RelType coexistence … deliberately defers"), so a *declared* edge and an *auto-detected* marker cannot coexist at the same key.

## Fix

Give the fixture's declared edge a weight distinct from the worker's auto-detected marker (0.8 → 0.5). Both edges then land at different keys; the race window disappears. The readout counts Declared edges regardless of weight, so `Count` / `DeclaredAt` / `Oldest` assertions are unaffected. Minimal, test-only, with a comment explaining the collision.

## Verification (RED → GREEN on the reproduction environment)

Reproduced in a cpu-throttled linux container (`podman run --cpus 0.5 golang:1.26 …`), which matches CI's slow/contended timing:

```
# before this fix
$ go test -race -run TestContradictionDebt ./internal/engine/ -count=50   # (--cpus 0.5)
--- FAIL: TestContradictionDebt_ZeroDeclaredAtIsUnknownNotEpoch (0.06s)
    engine_contradiction_debt_test.go:320: …existing rel_type 16… new rel_type 2
FAIL

# after this fix
$ go test -race -run TestContradictionDebt ./internal/engine/ -count=100  # (--cpus 0.5)
ok  	github.com/scrypster/muninndb/internal/engine	172.061s
```

Functional assertions still pass (`--- PASS`, `Count==2`, zero-`DeclaredAt` sorts first). The flake does not reproduce off-CI on an unloaded macOS/linux host — it needs the contended-runner timing, which is why CI caught it and local `make` didn't.

## Scope

Test-only. Does not touch engine logic, the guard, or the worker. A separate, deeper question the fix does **not** address (out of scope here): `WaitWriteTimeIdle` drains `autoAssoc`/`neighborWorker`/`goalLinkWorker`/`fts`/`activation` but not the contradiction worker, and declared-vs-detected edges can't coexist at one key by design — both are latent and worth their own thread, but neither is needed to stop this flake.

### DIFF
--- a/internal/engine/engine_contradiction_debt_test.go
+++ b/internal/engine/engine_contradiction_debt_test.go
@@ -50,8 +50,13 @@ func debtPairFixture(t *testing.T, eng *Engine, vault, subject, factA, factB str
 		t.Fatal(err)
 	}
 	ws := eng.Store().ResolveVaultPrefix(vault)
+	// Weight deliberately distinct from the auto-detected contradiction marker's
+	// weight. The contradiction worker emits its own (different-RelType) detected
+	// edge between these near-identical facts; if both edges land at the same
+	// (src, weight, target) the assoc_reltype_guard rejects the second write and
+	// the fixture flakes. The readout counts Declared edges regardless of weight.
 	if err := eng.Store().WriteAssociation(ctx, ws, idB, idA, &storage.Association{
-		TargetID: idA, RelType: storage.RelContradicts, Weight: 0.8, Confidence: 1,
+		TargetID: idA, RelType: storage.RelContradicts, Weight: 0.5, Confidence: 1,
 		CreatedAt: declaredAt,
 	}); err != nil {
 		t.Fatal(err)


scrypster (approved):
Stephen, thank you for this. I am sorry it sat for six days. You opened it on 12 Aug and today is 18 Aug. That delay is on us.

The collision is real and it is already hurting other work. PR #874's CI, merged against develop at 3a18561, died on this exact line:

```
--- FAIL: TestContradictionDebt_ZeroDeclaredAtIsUnknownNotEpoch (0.01s)
    engine_contradiction_debt_test.go:320: association write would replace a different
    rel_type at the same weight: existing rel_type 16 at weight 0.8 … new rel_type 2
```

rel_type 16 is RelRefines (0x0010). The novelty worker writes it at the frog pair's Jaccard, which is exactly 0.8. WaitWriteTimeIdle never drains that worker, so the declared RelContradicts write at 0.8 can lose the key. It is not a contradiction-worker detected marker. You already corrected the 0x0A FlagContradiction half on #875 on 14 Aug. The comment still committed here names the wrong worker. Please follow up with a five-line comment fix that says novelty / RelRefines so the next reader does not inherit the misdiagnosis. Johannes already did, from this comment.

The 0.8 to 0.5 move itself is the right test-only fix. I ran Jaccard on every debtPairFixture pair against the live novelty code. Only the frog pair is 0.8. None is 0.5. The readout counts Declared edges regardless of weight, so the COG-29 assertions stay honest. Locally, `go test -race -run TestContradictionDebt ./internal/engine/` is green on this head, and this PR's CI is green.

Merging now so #874 can rebase. Product collision stays on #880.

scrypster:
Taking the RelRefines comment wording on our side. Nothing else needed from you.

Issue #895, opened 2026-09-04
Replication: pre-upgrade leader's 42-byte 0x21 writes strand permanently on an upgraded (v7-stamped) node — rolling-upgrade direction undocumented


madeinoz67:
## Summary

Found by adversarial review of the #894 predicate-hash fix. In a replica-first rolling upgrade, relationship writes replicated by a not-yet-upgraded leader are applied byte-verbatim on an upgraded follower whose migration is already stamped — leaving legacy-layout keys that no migration ever re-keys.

## Mechanism (probe-proven on the fix branch)

- `Applier.Apply` applies OpBatch reprs byte-verbatim with no length/version gate (`internal/replication/applier.go:83-88`).
- The join handshake gates only the MBP wire protocol (`internal/replication/join.go:130`); a storage-key migration does not bump it.
- The upgraded node stamped migration v7 at its open. Refuse-newer is silent at 7==7, so replicated 42-byte 0x21 writes from the old leader land as strays that `RekeyRelationshipPredicateHash` never touches again.

Executable consequences on the branch:

1. Re-upserting the same logical assertion produces **two rows** (one 42-byte, one 49-byte key) for one logical record.
2. `RelinkRelationshipEntity` skips the 42-byte row (length guard) — stale entity names survive an entity merge.
3. `deleteEntityLinks` deletes the 42-byte 0x21 row but **leaves both orphaned 0x26 index entries**.

The pre-fix binary cannot be a sustained leader after the upgrade round-trip (refuse-newer blocks an old binary opening a migrated DB), so the exposure is the rolling-upgrade window itself — but the strays it creates are permanent, not transient.

## Recovery (works today)

`--force-migration-rerun` on the upgraded node re-runs v7, which re-keys any 42-byte key it finds regardless of stamp. The #894 PR pins this path with a test and documents both mixed-version directions (upgrade the leader first; force-rerun if a replica-first upgrade already happened).

## Design directions (maintainer call — this issue exists to get that decision made)

1. **Open-time stray sweep** for length-discriminated migrations: scan the prefix at open and re-key strays. Cost: an O(prefix) scan at every open even when clean.
2. **Version-gate replicated applies**: the applier rejects or quarantines key layouts that predate the local migration stamp. Bigger surface (replication semantics, log divergence on reject), but makes the bad state unrepresentable rather than recovered-from.
3. **Status quo + documentation**: the out-of-band posture already in `migrate.go` extended to name this direction (this is what the #894 PR ships).

[...]

Issue #893, opened 2026-09-04
RFC: Observation-grade entity relationships (OSINT / threat-intel enrichment)


madeinoz67:
## Summary

The entity knowledge graph is the natural landing surface for indicator-grade enrichment — OSINT resolvers (people, IPs, phone numbers, domains) and threat-intel feeds (STIX-style predicates like `communicates_with`, `attributed_to`, `indicates`) both want to write typed, time-bounded, multi-source edges between entities, and to read them back with that structure intact. The external-enrichment path (`muninn_get_enrichment_candidates` → resolver → `muninn_apply_enrichment`, `source:"osint"` / `"threat_intel"`) already exists and requires no plugin ABI.

What the storage layer gives them today has one defect that loses data and three gaps that lose meaning. This RFC proposes collision-safe predicate keys, `first_seen`/`last_seen` temporal windows, per-source observations, and a home for indicator/CTI entity types — as three separable increments.

### 1. Unknown predicates collide in the key space (data loss today)

`RelationshipRecord` values are keyed at `0x21 | ws | engramID | fromHash | relTypeByte | toHash` (`PebbleStore.UpsertRelationshipRecord`, `internal/storage/entity.go:760`). The predicate component of that key is a single byte produced by `relTypeByteFromString` (`internal/storage/entity.go:1150`):

```go
if b, ok := relTypeBytes[relType]; ok { return b }
return 0xFF
```

Every predicate string outside the built-in map folds to `0xFF`. The record's *value* keeps the full string, but the *key* does not — so two records asserted from the same engram about the same entity pair with different unmapped predicates (`communicates_with` and `attributed_to`) land on the identical key and `batch.Set` silently overwrites: last write wins, the first predicate is gone.

This is not hypothetical for the target use case. A threat-intel report (one engram) routinely asserts several predicates over overlapping entities ("198.51.100.7 `communicates_with` c2.example.net", "198.51.100.7 `hosts` payload-x", "c2.example.net `attributed_to` Gorgon Group"). An OSINT pass over one memory can assert `phone_of` and `registered_to` for the same person. Writing those today loses all but one edge per (engram, pair) — and loses them quietly, which is the worst failure class this project has (`docs/internals/invariants.md`; #582/#585/#589 lineage).

The predicate vocabulary is free-form by design (`validateRelationships`, `internal/plugin/enrich/parse.go:303`, applies no vocabulary), so nothing at the write surface prevents the collision.

### 2. No temporal validity

[...]

PR #896, opened 2026-09-04
fix(storage): 0x21 predicate-hash key component ends unmapped-predicate collision

### LINKED ISSUE #894: Entity relationship records: unmapped predicates collide on the 0xFF key byte (silent last-write-wins data loss)
## Summary

Two entity-relationship records asserted from the same engram about the same entity pair, with different predicate strings, land on the identical Pebble key — the second write silently overwrites the first. Last write wins, the lost predicate never surfaces anywhere.

## Mechanism (verified by code inspection)

Relationship records are keyed with a single predicate byte:

```
0x21 | ws | engramID | fromHash | relTypeByte | toHash
```

- `PebbleStore.UpsertRelationshipRecord` (`internal/storage/entity.go:760`) and the batch path `pebbleStoreBatch.WriteRelationshipRecord` (`internal/storage/batch.go:389`) both compute `relTypeByte := relTypeByteFromString(record.RelType)`.
- `relTypeByteFromString` (`internal/storage/entity.go:1150`) maps strings in the built-in `relTypeBytes` map to distinct bytes and folds **every other string to `0xFF`**:

```go
func relTypeByteFromString(relType string) uint8 {
    if b, ok := relTypeBytes[relType]; ok {
        return b
    }
    return 0xFF
}
```

- The full predicate string survives in the record *value*, but the *key* cannot distinguish two unmapped predicates. `batch.Set(relKey, val, nil)` then overwrites: one engram, one entity pair, at most one unmapped predicate survives.

The predicate vocabulary is free-form by design — `validateRelationships` (`internal/plugin/enrich/parse.go:303`) applies no vocabulary — so nothing at the write surface prevents this.

## Impact

- Any consumer using custom predicates (OSINT/threat-intel enrichment writing `communicates_with`, `attributed_to`, `phone_of`, `registered_to`, ... over the `muninn_apply_enrichment` / `muninn_remember` paths) loses all but one edge per (engram, entity pair), silently. This is the silent-plausible-wrong failure class `docs/internals/invariants.md` treats as the worst case.
- A single unmapped predicate per pair works fine and reads back with its full string, which makes the loss easy to miss: it only appears once a *second* predicate is asserted.

## RED test sketch

```go
// Two different unmapped predicates, same engram, same entity pair:
// both must be retrievable. Fails today — both map to key byte 0xFF.
func TestUpsertRelationshipRecord_DistinctPredicatesDoNotCollide(t *testing.T) {
    // store two records: ("Gorgon Group", "c2.example.net", "attributed_to")
    //                    ("Gorgon Group", "c2.example.net", "communicates_with")
    // same engramID, same vault.
    // Assert both round-trip via the relationship read path.
}
```

## Fix direction

Hash the predicate string into the key — the same pattern `keys.EntityNameHash` already uses for entity names in this very key — instead of a vocabulary byte. Identical predicates still coalesce; distinct predicates get distinct keys.

This changes the `0x21` key layout, so it needs a numbered storage migration (the #611 machinery: fail-loud, refuse-newer, `--force-migration-rerun` recovery). The `0x26` rel-entity index keys carry no predicate component (`RelEntityIndexKey(ws, entityHash, engramID)`) and are unaffected.

Full design (temporal windows, per-source observations, indicator entity types) is proposed in the RFC: #893 — this issue tracks just the data-loss defect, which stands alone and is worth fixing regardless of the RFC's fate.

madeinoz67:
Fixes #894.

## What this fixes

Two relationship predicates outside the old 11-entry `relTypeBytes` vocabulary, asserted from one engram about the same entity pair, both folded to the `0xFF` key byte and collided on one `0x21` key — the second write silently destroyed the first. As #894 documents, this was the common case, not an exotic one: `belongs_to`, `integrates_with`, `deployed_on` and `alternative_to` (four of the ten predicates the enrich plugin's default prompt suggests) were all unmapped.

## The change

- **Key layout**: the 1-byte predicate discriminant becomes `keys.PredicateHash(relType)` — an 8-byte Siphash of the RAW predicate string, the same construction the same key already uses twice for entity names. Key grows 42 → 49 bytes; named offset constants exported beside the constructor.
- **The hash is deliberately un-normalized**: the old map was exact-match, so `"Uses"`/`"uses"` already occupied different keys; a normalized hash would merge them — a silent loss, the exact failure class this fix kills. Predicate normalization is a separate semantic decision, explicitly deferred.
- **`relTypeBytes` is deleted, not frozen** — post-fix, a vault's own predicates are the vocabulary. (A frozen map + escape-byte hybrid was rejected: retaining the constant keeps the drift hazard alive and needs a promotion-as-migration rule forever.)
- The predicate byte was write-only — no reader ever decoded it (every consumer reads the msgpack value's `RelType` string), which is what makes the layout change safe. Both writers updated (`UpsertRelationshipRecord`, the batch twin used by evolve carry and startup repair) plus the two key-offset consumers (`RelinkRelationshipEntity`, `deleteEntityLinks`).
- **Migration v7** re-keys legacy 42-byte records: key length is the sole discriminator (one record family owns the prefix), the value's `RelType` — never the legacy byte — is authoritative for the destination hash, undecodable values fail loud with the version unstamped, Set+Delete ride one batch so a torn pass converges on re-run, Sync every 500 with batch re-allocation. Historical loss is not repaired: only the survivor's bytes were ever on disk.

## Testing

- **RED pair, shown failing on `34b505f`**: storage-level two-unmapped-predicates coexist (`communicates_with record was destroyed by the attributed_to write`) and engine-level `Engine.Write` → `ExportGraph` (1 edge, want 2). Both green on the branch.
- Both-ways controls: mapped-predicate pair coexists before AND after; same-predicate re-assert stays one upserted record.
- Sabotage-verified pins: constant-hash in either writer or in v7 → tests fail; frozen-copy tag drift → pin fails; legacy layout pinned by literal bytes.
- Migration: rekey, idempotent re-run, torn-state convergence, fail-loud on undecodable, layout-pins vs the live constructors, registration pins (`MaxRegisteredVersion()==7` exact in v7's test, v6 softened to a floor, refuse-downgrade mirror), a 600-record pass forcing the >500 batch re-allocation path, and `TestMigrationV7_ForceRerunRekeysReplicatedStrays` pinning the documented recovery below.

## Real-vault measurement

Migration + equality checks on a copy of a dormant production vault (enrichment enabled; original untouched), measured with an independent raw-Pebble scanner plus engine `ExportGraph` on both binaries:

| | before | after (v7) | 2nd open |
|---|---|---|---|
| 0x21 records | 130 (all 42-byte) | 130 (all 49-byte, 0 strays) | unchanged |
| `0xFF` fold exposure | 7 (5.4%) | 0 | 0 |
| ExportGraph edges | 129 | 129 | 129 |
| value bytes | baseline | byte-identical | byte-identical |

Vocabulary drift on that vault: 3 of 9 distinct predicates (`used_by`, `owns`, `alternative_to`) were outside the old map — exactly the folded 7 records. One slot carried `{co_occurs_with, used_by}`: it survived pre-fix only because `co_occurs_with` happens to be mapped. Honest scope note per the pre-committed acceptance rule: 130 records is below the 500 threshold for claiming torn/batch proof from a real vault, so those claims rest on the synthetic mechanism tests above; this pass contributes the exposure measurement and the no-loss/no-drift evidence. The realistic upgrade path (pending v4/v5/v6 applied on first open) also ran clean across 0x21.

## Mixed-version clusters (from the adversarial review pass)

Both directions are hazardous, in opposite ways; the operator rule is **upgrade the leader first, then followers promptly**:

1. Upgraded leader → pre-upgrade follower: applies 49-byte entries fine, but its length-guarded cascades skip relationship cleanup on hard-delete/entity-merge until upgraded.
2. Pre-upgrade leader → upgraded, v7-stamped follower: the replication applier commits replicated batch reprs byte-verbatim, so legacy 42-byte writes land AFTER the stamp and a plain reopen never re-runs v7 — permanent strays (double rows on re-upsert, stale-name row surviving merge, orphaned `0x26` pairs). **Recovery: `muninn start --force-migration-rerun` on the upgraded nodes** (pinned by test). The structural cure — version-gating replicated applies, or an open-time stray sweep — is a replication-subsystem design decision tracked in #895.

## Deferred

RFC #893 Increment 1 (observation-grade records: `first_seen`/`last_seen`, per-source observations); predicate normalization; STO-15's 0x03/0x04 key-format half; the 0x0A single-partner overwrite; entity-graph export/clone; a census of other fold-to-constant key components; historical-collision repair (impossible by construction).


madeinoz67:
The red Vulnerability scan is a stdlib/toolchain artifact, not this change: all six findings are Go standard-library advisories fixed in `go1.26.6` while CI builds on `go1.26.5` (GO-2026-5972 `encoding/asn1`, GO-2026-5026 `net/http`, +4 more), and every example trace is in files this PR does not touch (`internal/replication/tls.go`, `internal/plugin/embed/*`, `cmd/muninn/vault.go`, `internal/audit/webhook_sink.go`, `internal/plugin/transport.go`). Evidence it predates the branch: develop's own run on this PR's exact base commit `34b505f` (run 33605384065, Sep 2) passed the same scan — the advisories entered the vuln DB after that, so any run on any branch today fails it until the toolchain bumps. Happy to do the `go 1.26.6` bump as a separate one-line PR if wanted; keeping it out of this one to leave the storage diff reviewable.

